PR #6, opened 2019-03-05
Fix query and message for ignition check

yrobla:


dustymabe:
Hey @yrobla - thanks for the contribution! I notice a few things.

- There is a merge commit. We shouldn't include that in the PR. Can you remove it?
- Can you break out your whitespace changes into a separate commit?

PR #7, opened 2019-03-05
README: add instructions for using installer on FCOS

dustymabe:
This will help people who are trying to use Fedora CoreOS on
bare metal understand how to use the installer to perform
the installation on those nodes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,112 @@ via a dracut module.
 
 
 ## Kernel command line options for coreos-installer running in the initramfs
+
 * coreos.inst=yes - Instruct the installer to run
 * coreos.inst.install_dev - The block device on the system to install to
 * coreos.inst.image_url - The url of the coreos image to install to this device
 * coreos.inst.ignition_url - The url of the coreos ignition config (optional, enter
   coreos.inst.ignition_url=skip to not load an ignition config)
 
-## Testing out the installer script
+## Using the installer on Fedora or RHEL CoreOS
+
+This installer is incorporated into Fedora CoreOS and RHEL CoreOS.
+There are ISO images that can be downloaded that will allow for an
+install to be performed on bare metal hardware either via ISO install
+or via a PXE install. While ISO install is supported we certainly
+recommend PXE if you environment supports it since it is more friendly
+to automation.
+
+#### Grab an ISO image and bare metal image
+
+For Fedora CoreOS you can currently grab an ISO image from the output
+of the current development pipeline located
+[here](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/).
+
+The ISO image can install in either legacy boot (BIOS) mode or in UEFI
+mode. You'll have to make sure you download the appropriate related
+image for the mode you'd like to use. It is a good idea to download
+them before doing the install as the artifacts server that is being
+used currently has very slow download speeds.
+
+For example download:
+
+- [fedora-coreos-29.731.iso](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/fedora-coreos-29.731.iso)
+
+and one of the following two:
+
+- [fedora-coreos-29.731-metal-bios.raw.gz](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/fedora-coreos-29.731-metal-bios.raw.gz)
+- [fedora-coreos-29.731-metal-uefi.raw.gz](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/fedora-coreos-29.731-metal-uefi.raw.gz)
+
+**NOTE** The artifacts output of the pipeline are development
+         artifacts. The links above will quickly become
+         broken because we prune builds. As we get closer to
+         an official release we'll have stable links but for
+         now you'll have to find your own links from the
+         [latest directory](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/).
+
+#### Test a PXE based install
+
+Using the ISO images you can also do a PXE based install. You can
+mount up the ISO images and use the `initramfs.img` and `vmlinuz`
+for PXE boot. Here is an example `pxelinux.cfg` that I used to perform
+a PXE boot:
+
+```
+DEFAULT pxeboot
+TIMEOUT 20
+PROMPT 0
+LABEL pxeboot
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,13 +6,100 @@ via a dracut module.
...
+Now press `<ENTER>` (isolinux) or `<CTRL-x>` (grub) to kick off the
+install. The install will occur on tty2 and there are very few good
+log statements or debug opportunities. Thee user experience here
+needs work and is tracked in https://github.com/coreos/coreos-installer/issues/5 .
+
+The install should progress and eventually reboot the machine. After
+reboot the machine will boot into the installed system and the
+embedded ignition config should run on first boot.
+
+#### Test a PXE based install
```
ajeddeloh: Lets move this above ISO, since its an easier path lets encourage users to use it by putting it first.

README.md
```diff
@@ -6,13 +6,100 @@ via a dracut module.
...
+         [latest directory](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/).
+
+#### Test an ISO based install
+
+You can test an install on a bare metal machine by burning the ISO to
+disk and booting it or using ISO redirection via a LOM interface.
+Alternatively you can use a VM like so:
+
+```
+virt-install --name cdrom --ram 4500 --vcpus 2 --disk size=20 --accelerate --cdrom /path/to/fedora-coreos-29.731.iso --network default
```
ajeddeloh: My gut says to have a qemu example as well. the libvirt tooling tends to be more "automagical" which is great once you understand it, but hides the details that users reading this documentation might want. Having both is probably best to cover both the "I just want to run it in a vm" case and the "I'm looking at how you run it in a vm to gain better understanding" case.

README.md
```diff
@@ -6,13 +6,100 @@ via a dracut module.
...
 * coreos.inst.image_url - The url of the coreos image to install to this device
 * coreos.inst.ignition_url - The url of the coreos ignition config (optional, enter
   coreos.inst.ignition_url=skip to not load an ignition config)
 
-## Testing out the installer script
+## Using the installer on Fedora or RHEL CoreOS
+
+This installer is incorporated into Fedora CoreOS and RHEL CoreOS.
+There are ISO images that can be downloaded that will allow for an
+install to be performed on bare metal hardware either via ISO install
```
ajeddeloh: I'd phrase this in a way that suggests we like PXE more.

README.md
```diff
@@ -6,13 +6,100 @@ via a dracut module.
...
+For example download:
+
+- [fedora-coreos-29.731.iso](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/fedora-coreos-29.731.iso)
+
+and one of the following two:
+
+- [fedora-coreos-29.731-metal-bios.raw.gz](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/fedora-coreos-29.731-metal-bios.raw.gz)
+- [fedora-coreos-29.731-metal-uefi.raw.gz](https://ci.centos.org/artifacts/fedora-coreos/prod/builds/latest/fedora-coreos-29.731-metal-uefi.raw.gz)
+
+**NOTE** Those links will quickly become broken because we prune
```
ajeddeloh: Since we're going to merge bios/uefi, we should make a note that these are just unstable in general until we have an actual release.

README.md
```diff
@@ -6,13 +6,100 @@ via a dracut module.
...
+for PXE boot. Here is an example `pxelinux.cfg` that I used to perform
+a PXE boot:
+
+```
+DEFAULT pxeboot
+TIMEOUT 20
+PROMPT 0
+LABEL pxeboot
+    KERNEL fedora-coreos-29.731.iso/images/vmlinuz
+    APPEND initrd=fedora-coreos-29.731.iso/images/initramfs.img console=tty0 console=ttyS0 coreos.inst=yes coreos.inst.install_dev=sda coreos.inst.image_url=http://192.168.1.101:8000/fedora-coreos-29.731-metal-bios.raw.gz coreos.inst.ignition_url=http://192.168.1.101:8000/config.ign
```
ajeddeloh: remove the first `console=tty0` since `console=ttyS0` will overwrite it?

README.md
dustymabe: I can do that

README.md
dustymabe: ehh - since it's just an example I'd prefer to not since it might be more confusing than helpful. when we make official docs it might be nice to have an *expandable* clicky option for either virt-install or qemu directly, WDYT?

README.md
dustymabe: yeah I think we'll just update this when we move to official docs. I can add a note here about these being development images, though

README.md
ajeddeloh: I'm coming from the other direction where using `virt-install` is more confusing that helpful. I'd error on the side of using tools that do as little as possible in docs since it's more clear what's happening under the hood.

README.md
dustymabe: I will remove the 2nd option but do note that kernel mesages will go to every `console=` device on the command line. The last `console=` device will get set as the primary console.

ajeddeloh: Today I learned, I'd keep it then.

README.md
dustymabe: Not necessarily trying to be too prescriptive here, but I will add a note about PXE probably being better for automation..

README.md
dustymabe: can you give me the equivalent qemu command and I'll paste it in?

dustymabe:
pushed up a fixup commit that addresses most comments

dustymabe:
> pushed up a fixup commit that addresses most comments

i'll need to squash before we merge

README.md
ajeddeloh:
```
qemu-system-x86_64 -cpu host -accel kvm -cdrom fedora-coreos-29.731.iso -drive file=drivefile,if=virtio,format=raw -nographic -m 5G
```
should work, but I'm having trouble testing it. Will dig in tomorrow. (install dev would be `vda` not `sda` in this case)

dustymabe:
squashed commits and add #TODO for the qemu command. Thanks Andrew.

README.md
dustymabe: Added a #TODO to add a qemu command - I want to push this out so I can link people to it and close out some other tickets so let's add the qemu command in a follow up.

PR #8, opened 2019-03-06
Correct check in skipping ignition file, and message in download

yrobla:
I also included removing extra spaces, i'm unable to remove those from the commit, as editor automatically remove those.


dustymabe:
can you use an editor that doesn't automatically remove them? or just make the whitespace changes in the first commit, and the fix in the 2nd.

yrobla:
Created a new one, generated with differnet editor

PR #15, opened 2019-03-20
README: fix .treeinfo contents

arithx:
Adds the line `version = 29` to the `.treeinfo` file declaration.
Without this the `virt-install` will fail with `No option 'version' in
section: 'general'`.


dustymabe:
so you did confirm it works when you add that?

arithx:
@dustymabe yeah, adding that line removed the error message shown in the commit message.

PR #18, opened 2019-03-22
*: persist ip= kcmdline opts to firstboot initramfs

### LINKED ISSUE #17: detect networking kcmdline args and write them to a file
If someone provides static networking configuration via the kernel command line for the run of the installer then it would be nice if we persist those kernel command line options into the first boot of the real machine (as networking configuration that is done via ignition will not apply to the initramfs where the embedded ignition config may have references to external resources).

One way we could implement something like this is by detecting these options (we'll want to be selective about which ones we propagate) and then add them to the kernel command line on first boot just like we do for `ignition.firstboot` using: 

https://github.com/coreos/ignition-dracut/blob/b11553f9e2b530fbbb5a2504bc72b364829c879c/grub/02_ignition_firstboot#L1-L13

arithx:
Adds support for persisting `ip=` kernel command-line flags from the
installer initramfs to the firstboot initramfs via creating an
`ignition.firstboot` file that can be sourced on the boot partition of
the installed image.

Closes #17

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -318,6 +318,25 @@ write_ignition_file() {
 }
 
 ############################################################
+# Helper to persist networking options to initramfs first boot
+############################################################
+write_networking_opts() {
+    if [ ! -f /tmp/networking_opts ]; then
+        return
+    fi
+
+    dialog --title 'CoreOS Installer' --infobox "Embedding provided networking options" 5 70
+    # check for the boot partition
+    mkdir -p /mnt/boot_partition
+    # TODO check to make sure the disk with label 'boot'
+    #      is part of ${DEST_DEV}
+    mount /dev/disk/by-label/boot /mnt/boot_partition
+    trap 'umount /mnt/boot_partition' RETURN
+
+    echo "set ignition_network_kcmdline=\"$(cat /tmp/networking_opts)\"" >> /mnt/boot_partition/ignition.firstboot
+}
+
+############################################################
 # START HERE
 ############################################################
 import_gpg_key() {
@@ -595,6 +614,9 @@ main() {
 
     # If one was provided, install the ignition config
     write_ignition_file
+
+    # If networking options were present, persist to firstboot initramfs
+    write_networking_opts
 
     if [ ! -f /tmp/skip_reboot ]
     then
--- a/dracut/30coreos-installer/parse-coreos.sh
+++ b/dracut/30coreos-installer/parse-coreos.sh
@@ -24,6 +24,28 @@ then
     echo $IGNITION_URL >> /tmp/ignition_url
 fi
 
+# Dracut networking args
+# Parse all args (other than rd.neednet) and persist those into /tmp/networking_opts
+# List from https://www.mankier.com/7/dracut.cmdline#Description-Network
+local NETWORKING_ARGS="rd.neednet=1"
+declare -a NET_ARGS=("ip=" "ifname=" "rd.route=" "bootdev=" "BOOTIF=" "rd.bootif=" "nameserver=" "rd.peerdns=" "biosdevname=" "vlan=" "bond=" "team=" "bridge=")
+for NET_ARG in "${NET_ARGS[@]}"
+do
+    NET_OPT=$(getarg $NET_ARG)
+    if [ ! -z "$NET_OPT" ]
+    then
+        echo "persist $NET_ARG to $NET_OPT" >> /tmp/debug
+        NETWORKING_ARGS+=" ${NET_ARG}${NET_OPT}"
+    fi
+done
+# only write /tmp/networking_opts if additional networking options have been specified
+# as the default in ignition-dracut specifies `rd.neednet=1 ip=dhcp` when no options are persisted
+if [ "${NETWORKING_ARGS}" != "rd.neednet=1" ]
+then
+    echo "persisting network options: ${NETWORKING_ARGS}" >> /tmp/debug
+    echo "${NETWORKING_ARGS}" >> /tmp/networking_opts
+fi
+
 if getargbool 0 coreos.inst.skip_media_check
 then
     echo "Asserting skip of media check" >> /tmp/debug


arithx:
NOTE: currently untested & needs to land in conjunction with an RPM containing https://github.com/coreos/ignition-dracut/pull/65

Additionally we will need some additional syncing of what exactly we're persisting, this initial implementation only persists `ip=` but will be updated based on decisions from #17

coreos-installer
```diff
@@ -317,6 +317,25 @@ write_ignition_file() {
...
+
+    dialog --title 'CoreOS Installer' --infobox "Embedding provided networking options" 5 70
+    # check for the boot partition
+    mkdir -p /mnt/boot_partition
+    # TODO check to make sure the disk with label 'boot'
+    #      is part of ${DEST_DEV}
+    mount /dev/disk/by-label/boot /mnt/boot_partition
+    trap 'umount /mnt/boot_partition' RETURN
+
+    cp /tmp/ip_opt /mnt/boot_partition/ignition.firstboot
```
dustymabe: if we stick with the implementation in https://github.com/coreos/ignition-dracut/pull/65 then i think we'd need to make the format of the file be `set ignition_network_kcmdline="$(cat /tmp/ip_opt)"`. I chose to make the file able to be sourced so that if we find other uses for it we can use other variable names in the future.

Also maybe we should append to the file rather than overwriting it completely.

coreos-installer
```diff
@@ -317,6 +317,25 @@ write_ignition_file() {
...
+        return
+    fi
+
+    dialog --title 'CoreOS Installer' --infobox "Embedding provided networking options" 5 70
+    # check for the boot partition
+    mkdir -p /mnt/boot_partition
+    # TODO check to make sure the disk with label 'boot'
+    #      is part of ${DEST_DEV}
+    mount /dev/disk/by-label/boot /mnt/boot_partition
+    trap 'umount /mnt/boot_partition' RETURN
```
dustymabe: Now that we're mounting/unmounting in two places in this file I'll ask a dumb question: does this unmount on return of this function?

arithx: I believe so, https://ss64.com/bash/trap.html directly calls out shell functions for `RETURN` and testing inside of a shell script locally seems to work. I'll make another note in this comment chain about it when I test this properly.

coreos-installer
arithx: Is this file currently used for anything? I explicitly just overwrote it because my understanding was that it was currently only checked to exist rather than using it's contents.

dustymabe: > Is this file currently used for anything? 

nope

> I explicitly just overwrote it because my understanding was that it was currently only checked to exist rather than using it's contents.

That is a correct assumption and this code as is works fine. My point is that we may find more uses for this file in the future and we can enable that now by using variables within the file and appending rather than entire file contents.

arithx: Fair enough, I'll update it to append.

dustymabe:
just tested this out. I used this short howto on setting network args using dracut:

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-configuring_ip_networking_from_the_kernel_command_line

One thing that we definitely need to do probably is propagate forward at least the `nameserver` arg also. Can you add that to this PR?

dustymabe:
> just tested this out.

Forgot to mention.. It worked!

arithx:
Updated to propogate every dracut networking arg listed in https://www.mankier.com/7/dracut.cmdline#Description-Network

If you have a quicker local setup for testing would you mind giving it another spin @dustymabe

dustymabe:
> If you have a quicker local setup for testing would you mind giving it another spin @dustymabe

+1 will test with at least `ip=` and `nameserver=` args.

dracut/30coreos-installer/parse-coreos.sh
```diff
@@ -24,6 +24,110 @@ then
...
+local RDBOOTIF_OPT=$(getarg rd.bootif=)
+local NAMESERVER_OPT=$(getarg nameserver=)
+local PEERDNS_OPT=$(getarg rd.peerdns=)
+local BIOSDEVNAME_OPT=$(getarg biosdevname=)
+local VLAN_OPT=$(getarg vlan=)
+local BOND_OPT=$(getarg bond=)
+local TEAM_OPT=$(getarg team=)
+local BRIDGE_OPT=$(getarg bridge=)
+
+if [ ! -z "$IP_OPT" ]
```
dustymabe: any opposition to running this as a for loop instead?

dustymabe:
One thing I think of as i was going through to test this. I wonder if we should enable disabling this functionality in case a bug comes up or causes unexpected behavior.. Then again people can just delete `$ignition_firstboot` from their kernel command line if they need to at grub on first boot I guess.

dracut/30coreos-installer/parse-coreos.sh
arithx: Re-working it into a loop works for me

dracut/30coreos-installer/parse-coreos.sh
```diff
@@ -24,6 +24,28 @@ then
...
+# List from https://www.mankier.com/7/dracut.cmdline#Description-Network
+local NETWORKING_ARGS="rd.neednet=1"
+declare -a NET_ARGS=("ip=" "ifname=" "rd.route=" "bootdev=" "BOOTIF=" "rd.bootif=" "nameserver=" "rd.peerdns=" "biosdevname=" "vlan=" "bond=" "team=" "bridge=")
+for NET_ARG in "${NET_ARGS[@]}"
+do
+    NET_OPT=$(getarg $NET_ARG)
+    if [ ! -z "$NET_OPT" ]
+    then
+        echo "persist $NET_ARG to $NET_OPT" >> /tmp/debug
+        NETWORKING_ARGS="${NETWORKING_ARGS} ${NET_ARG}${NET_OPT}"
```
dustymabe: minor - I like using `+=` in situations like these. i.e. `NETWORKING_ARGS+=" ${NET_ARG}${NET_OPT}"`

WDYT?

arithx: Either way WFM; I'm not of a particularly strong opinion either way on bash-isms. Updated.

dustymabe:
This LGTM - merging.

PR #20, opened 2019-03-29
Use `dd conv=sparse`

cgwalters:
This will skip the empty space in the filesystems, and will
rather dramatically speed up the install.

Not really necessary after
https://github.com/coreos/coreos-assembler/pull/332
but doesn't conflict either.

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -588,7 +588,7 @@ write_image_to_disk() {
     log "Writing disk image"
     
     zcat /mnt/dl/imagefile.gz |\
-    dd bs=1M iflag=fullblock oflag=direct of="${DEST_DEV}" status=none
+    dd bs=1M iflag=fullblock oflag=direct conv=sparse of="${DEST_DEV}" status=none
 
     for try in 0 1 2 4; do
         sleep "$try"  # Give the device a bit more time on each attempt.


cgwalters:
Rebased :surfing_man:  (though I didn't test this)

arithx:
LGTM (tested on my local libvirt setup); can we get someone who has a baremetal setup to test this before merge?

dustymabe:
> can we get someone who has a baremetal setup to test this before merge

is that required? i'm thinking a virt setup should suffice.

dustymabe:
+1 - i'm doing tests in this area too. still virt, but at least some other prodding of it

bgilbert:
This isn't safe.  The image might contain zeroed blocks for reasons other than unallocated space, and if the target disk already has data on it, we'll write a corrupt image to disk.

cgwalters:
> This isn't safe. The image might contain zeroed blocks for reasons other than unallocated space, and if the target disk already has data on it, we'll write a corrupt image to disk.

Yeah, you're right.  A problem masked by testing in virt where I think you'll normally see zeroed blocks.

cgwalters:
https://github.com/coreos/coreos-installer/pull/70

Issue #4, opened 2019-02-25
tty messages obscure and/or block keyboard input in the ncurses UI


mrguitar:
I hit the same things as @e-minguez did here: https://github.com/coreos/coreos-installer/issues/2#issuecomment-465659981

I was unable to enter the image URL when this message appeared: 
![installer-ui](https://user-images.githubusercontent.com/1071620/53369007-cc83e780-390f-11e9-8f64-1b8793504fc0.png)

Later I couldn't see the progress bar (identical to the other issue)
![screenshot_fcos_2019-02-25_14 26 10](https://user-images.githubusercontent.com/1071620/53369070-f50be180-390f-11e9-9838-0581f8fd927c.png)

But the install completed perfectly the second time with the options appended to the kernel! This is a paper cut we still need to fix.

dustymabe:
yep. known issue - needed to be documented. Thanks!

PR #23, opened 2019-04-09
Add coreos-installer.target and make default

jlebon:
What we really want when in "installer mode" is to override the default
flow of the bootup. The cleanest way to do this is to define a new
default target for systemd rather than `initrd.target`. That way we
don't have to awkwardly wedge ourselves between existing services. One
practical consequence of this for example is that systemd no longer
prints "Waiting" messages, since we're freely running without
constraints.

Closes: #4

### DIFF
--- a/dracut/30coreos-installer/coreos-installer-generator
+++ b/dracut/30coreos-installer/coreos-installer-generator
@@ -1,0 +1,23 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+set -e
+
+UNIT_DIR="${1:-/tmp}"
+
+cmdline=( $(</proc/cmdline) )
+cmdline_arg() {
+    local name="$1" value="$2"
+    for arg in "${cmdline[@]}"; do
+        if [[ "${arg%%=*}" == "${name}" ]]; then
+            value="${arg#*=}"
+        fi
+    done
+    echo "${value}"
+}
+
+if [[ $(cmdline_arg coreos.inst) == "yes" ]]; then
+    ln -sf "/usr/lib/systemd/system/coreos-installer.target" \
+        "${UNIT_DIR}/default.target"
+fi
--- a/dracut/30coreos-installer/coreos-installer.service
+++ b/dracut/30coreos-installer/coreos-installer.service
@@ -1,15 +1,9 @@
 [Unit]
 Description=CoreOS Installer
-Requires=network-online.target
-After=network-online.target
-After=dracut-initqueue.service
-Before=dracut-pre-mount.service
-ConditionKernelCommandLine=coreos.inst=yes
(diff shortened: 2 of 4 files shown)


jlebon:
This is tested on FCOS. Testing on RHCOS now.

jlebon:
And now tested on RHCOS as well. :heavy_check_mark:

yuqi-zhang:
Tested an RHCOS build via `cosa`, confirm that the generated iso no longer blocks the installer prompts. LGTM!

dracut/30coreos-installer/coreos-installer-generator
```diff
@@ -0,0 +1,23 @@
...
+    local name="$1" value="$2"
+    for arg in "${cmdline[@]}"; do
+        if [[ "${arg%%=*}" == "${name}" ]]; then
+            value="${arg#*=}"
+        fi
+    done
+    echo "${value}"
+}
+
+if [[ $(cmdline_arg coreos.inst) == "yes" ]]; then
```
ajeddeloh: quotes around `$()`?

dracut/30coreos-installer/coreos-installer.service
```diff
@@ -1,15 +1,9 @@
...
-After=dracut-initqueue.service
-Before=dracut-pre-mount.service
-ConditionKernelCommandLine=coreos.inst=yes
-OnFailure=emergency.target
-OnFailureIsolate=yes
+After=coreos-installer.target
 
 [Service]
-Type=oneshot
+Type=simple
```
ajeddeloh: Any reason for this change?

dracut/30coreos-installer/coreos-installer.service
```diff
@@ -1,15 +1,9 @@
 [Unit]
 Description=CoreOS Installer
-Requires=network-online.target
-After=network-online.target
```
ajeddeloh: I think we dropped this? Was that intentional?

dracut/30coreos-installer/coreos-installer-generator
jlebon: The `[[` is a bash intrinsic which knows not to get tripped up on things like this. :)

dracut/30coreos-installer/coreos-installer.service
jlebon: Essentially, this service is the main process in this installer "runlevel". Think of it like `gdm.service` or something. One this service is started, systemd's job is done as far as booting things up.

dracut/30coreos-installer/coreos-installer.service
jlebon: AFAIK it doesn't serve any purpose since we rely on dracut to set up networking, which doesn't integrate with the special `network*` targets.

ajeddeloh: :cry: Can we add some comments explaining that this needs networking? Otherwise we might get bitten when we switch to NM in the initramfs.

Issue #21, opened 2019-04-02
Support curl'ing to a URL during install to a provisioning system


dcritch:
I'm installing via PXE (with Foreman/Red Hat Satellite) and its working great!

The typical flow with Foreman is:
1. click 'build'
2. the pxelinux.cfg for the host gets modified to point to the install initrd/kernel/etc
3. system pxe boots and installs the OS
4. system executes a curl against a specific URL to tell Foreman that the system is built
5. Foreman modifies the pxelinux.cfg so that the system will boot from local disk
6. system reboots in to installed system

The issue I have is that there is no way to do step 4 with CoreOS. As a result, the host permanently stays in build mode and re-installs on every reboot. Manually cancelling the build in Foreman works, but must be done at the right time and ruins the automation of it all.

It would be great if there was a way to curl an arbitrary URL to inform Foreman or any PXE based build system to reset the config to properly complete the installation.

In the case of Foreman, the URL is unique and generated for each build. The pxelinux.cfg files are generated from templates, so I'm including that URL in the hopes of one day being able to pick it up (e.g. from /proc/cmdline) during install and curl'ing it.

Here's my PXE config for reference:
```
TIMEOUT 10
DEFAULT Fedora CoreOS default PXELinux

LABEL Fedora CoreOS default PXELinux
  KERNEL boot/fcos-installer-kernel
  APPEND ip=dhcp rd.neednet=1 initrd=boot/fcos-installer-initramfs.img console=tty0 console=ttyS0 coreos.autologin=tty1 coreos.autologin=ttyS0 coreos.inst=yes coreos.inst.install_dev=vda coreos.first_boot=1 coreos.inst.image_url=http://foreman.example.com/pub/fcos/fcos-metal-bios.raw.gz coreos.inst.ignition_url=http://foreman.example.com/pub/ignition.ign coreos.autologin=1 forman_url=http://foreman.example.com/unattended/built?token=fc697858-ecb2-4e49-bb22-a641d3fc22d3
  IPAPPEND 2
```
I'd like to curl that `forman_url=http://foreman.example.com/unattended/built?token=fc697858-ecb2-4e49-bb22-a641d3fc22d3` before it is too late.

lzap:
Implemented this in https://github.com/coreos/coreos-installer/pull/25 have fun.

dustymabe:
Thanks @lzap - This seems like a reasonable request. We'll consider it along with other similar requests and try to have a solution for this problem in the next iteration of the installer.

[...]

PR #25, opened 2019-05-03
installer: added coreos.inst.callhome_url option

lzap:
Implements a simple callback mechanism for provisioning management software (like Foreman or Red Hat Satellite) to inform the system to exit build mode and reconfigure TFTP/PXE services.

Issue: https://github.com/coreos/coreos-installer/issues/21

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -609,7 +609,7 @@ log() {
 }
 
 #########################################################
-#And Write the image to disk
+#Write the image to disk
 #########################################################
 write_image_to_disk() {
     log "Writing disk image"
@@ -631,6 +631,18 @@ write_image_to_disk() {
 
     log "Waiting for udev"
     udevadm settle
+}
+
+############################################################
+#Perform a HTTP POST to URL to signal that the install is complete
+############################################################
+call_home() {
+    CALLHOME_URL=$(cat /tmp/callhome_url)
+    if [ ! -z "$CALLHOME_URL" ]
+      echo "Calling home: $CALLHOME_URL" >> /tmp/debug
+      curl -k -o /dev/null --data-binary @/tmp/debug "$CALLHOME_URL" >>/tmp/debug 2>&1
+      rm -f /tmp/callhome_url
+    fi
 }
 
 parse_args() {
@@ -709,6 +721,9 @@ main() {
 
     # If networking options were present, persist to firstboot initramfs
     write_networking_opts
+
+    # And call home as the very last step before rebooting
+    call_home
 
     log "Install complete"
 }
--- a/dracut/30coreos-installer/parse-coreos.sh
+++ b/dracut/30coreos-installer/parse-coreos.sh
@@ -24,6 +24,13 @@ then
     echo $IGNITION_URL >> /tmp/ignition_url
 fi
 
+local CALLHOME_URL=$(getarg coreos.inst.callhome_url=)
+if [ ! -z "$CALLHOME_URL" ]
+then
+    echo "preset call home url to $CALLHOME_URL" >> /tmp/debug
+    echo "$CALLHOME_URL" >> /tmp/callhome_url
+fi
+
 # Dracut networking args
 # Parse all args (other than rd.neednet) and persist those into /tmp/networking_opts
 # List from https://www.mankier.com/7/dracut.cmdline#Description-Network


coreos-installer
```diff
@@ -591,6 +591,18 @@ write_image_to_disk() {
...
 }
 
+############################################################
+#Make call home HTTP(s) call as the final step
+############################################################
+call_home() {
+    CALLHOME_URL=$(cat /tmp/callhome_url)
+    if [ ! -z "$CALLHOME_URL" ]
+      echo "Calling home: $CALLHOME_URL" >> /tmp/debug
+      curl -skf -o /dev/null "$CALLHOME_URL" >/tmp/debug 2>&1
```
cgwalters: Would it be helpful to add any other data here?  And related make this a `POST` rather than `GET`?  Something like:
```
(for x in /sys/class/net/*; do
  if ! -f $x/address; then continue; fi
  echo $x $(cat $x/address)
) > /tmp/net-devices.txt
curl -F coreos.inst=1,net-devices=@/tmp/net-devices -skf $CALLHOME_URL
```

or so?

lzap: Cheers Colin! The `POST` is a great comment, we actually had `GET` for years and I've only fixed this few months ago so our endpoint actually accepts more correct `POST`. Let me fix that.

For the data, it's actually too late to provide any info to a provisioning management system because at this point everything was already feeded and initial configuration was done. At least in all our workflows, we heavily rely on configuration management systems like Ansible or Puppet which takes over on the initial boot and talks to our system in a different way (gathering facts and stuff).

However, it's a good point. Our endpoint actually accepts two calls - "finished" and "failed" when the latter accepts arbitrary text (log file) which is stored for later use. However this would require providing two URLs - one for success and one for failed case. This would probably be too complex. We were doing fine with just a simple "call home we are done" call, errors can be always investigated later on.

I am going to change the call to `POST` then.

lzap: Rebased, how about this? Sending the `/tmp/debug/` file in the POST body. I can modify our "success" endpoint to store it in the database or at least log it.

dracut/30coreos-installer/parse-coreos.sh
```diff
@@ -24,6 +24,13 @@ then
     echo $IGNITION_URL >> /tmp/ignition_url
 fi
 
+local CALLHOME_URL=$(getarg coreos.inst.callhome_url=)
+if [ ! -z "$CALLHOME_URL" ]
+then
+    echo "preset call home url to $CALLHOME_URL" >> /tmp/debug
+    echo $CALLHOME_URL >> /tmp/callhome_url
```
ekohl: I'd quote this just to be sure. It can contain `&` which can break in interesting ways.

```suggestion
    echo "$CALLHOME_URL" >> /tmp/callhome_url
```

coreos-installer
```diff
@@ -591,6 +591,18 @@ write_image_to_disk() {
     udevadm settle
 }
 
+############################################################
+#Make call home HTTP(s) call as the final step
+############################################################
+call_home() {
+    CALLHOME_URL=$(cat /tmp/callhome_url)
+    if [ ! -z "$CALLHOME_URL" ]
```
ekohl: For my understanding: isn't this equivalent to `if [ -n "$CALLHOME_URL" ]`?

coreos-installer
```diff
@@ -591,6 +591,18 @@ write_image_to_disk() {
...
+      echo "Calling home: $CALLHOME_URL" >> /tmp/debug
+      curl -skf -o /dev/null --data-binary @/tmp/debug "$CALLHOME_URL" >/tmp/debug 2>&1
```
arithx: Is there a particular reason for this to be an insecure request?

As well, why the silent failure? If a user has specified a call home URL it seems like not being able to dial home is a potentially critical failure.

coreos-installer
lzap: I copy-pasted this, this is my very first dracut patch and I am really not sure if this is being executed by GNU bash or something else. So keeping.

dracut/30coreos-installer/parse-coreos.sh
lzap: Very nice catch.

coreos-installer
lzap: > Is there a particular reason for this to be an insecure request?

Yes, there is. Where would I get a server certificate from? Or is there a mechanism to inject/embed certificates in the initramdisk? Then I will happily remove that, however this will force users to rebuild initramdisk everytime. Pulling it from network does not make any sense to me.

> As well, why the silent failure?

This was a copy-paste error. No particular reason.

lzap:
I rebased, however I am keeping `-k` for now until we confirm.

coreos-installer
cgwalters: Not easily.  However, on a related topic for OpenShift 4 the installer sets things up so that the Ignition served to the device is just a (URL, cert) pair.  See e.g. https://github.com/openshift/machine-config-operator/pull/759
However, we can't quite fit a cert into a kernel command line. 

Maybe what we should do is have e.g. `coreos.inst.callhome_cert_fingerprint` and trust the cert matching that fingerprint.  But eh, don't need to do it in this patch.

lzap: I'll do what you say, however I also lean towards just waiving this.

coreos-installer
```diff
@@ -591,6 +591,18 @@ write_image_to_disk() {
     udevadm settle
 }
 
+############################################################
+#Make call home HTTP(s) call as the final step
```
cgwalters: "Perform a HTTP POST to URL to signal that the install is complete"

coreos-installer
```diff
@@ -591,6 +591,18 @@ write_image_to_disk() {
...
+############################################################
+call_home() {
+    CALLHOME_URL=$(cat /tmp/callhome_url)
+    if [ ! -z "$CALLHOME_URL" ]
+      echo "Calling home: $CALLHOME_URL" >> /tmp/debug
+      curl -k -o /dev/null --data-binary @/tmp/debug "$CALLHOME_URL" >/tmp/debug 2>&1
```
cgwalters: `>>/tmp/debug` right?

coreos-installer
arithx: I was more thinking along the lines of forcing the user to specify HTTP endpoints to convey that the call is insecure rather than accepting HTTPS endpoints and ignoring cert failures.

lzap:
Resolved the two comments, thanks.

cgwalters:
Looks like this needs a rebase :surfing_man:

lzap:
Sure, done.

coreos-installer
```diff
@@ -633,6 +633,18 @@ write_image_to_disk() {
...
 }
 
+############################################################
+#Perform a HTTP POST to URL to signal that the install is complete
+############################################################
+call_home() {
+    CALLHOME_URL=$(cat /tmp/callhome_url)
+    if [ ! -z "$CALLHOME_URL" ]
+      echo "Calling home: $CALLHOME_URL" >> /tmp/debug
+      curl -k -o /dev/null --data-binary @/tmp/debug "$CALLHOME_URL" >>/tmp/debug 2>&1
```
ekohl: I always like using long versions in scripts to see what's going on. So `--insecure` instead of `-k`

lzap:
For the record we agreed that it is not the best approach to start adding more code to the initram, but we failed to figure out something better. Let's have another meeting a look on how to take a different solution probably by using `coreos-installer` directly from a discovered node. I am keeping this opened for now.

dustymabe:
Thanks @lzap for updating us here. 

>  I am keeping this opened for now.

+1

bgilbert:
https://github.com/coreos/fedora-coreos-tracker/issues/203 proposes a solution that could be implemented in the medium term.

fsperling:
We'd really like to see this feature as well.

StykMartin:
Same for Beaker project

ashcrow:
Bump

bgilbert:
Fedora CoreOS now runs the installer from the real root, in the live ISO/PXE image.  To call home, the live image is _itself_ booted with an Ignition config, which configures additional systemd services that run before/after the installer as desired.  Those services can use all the facilities of the real root, and can perform arbitrary custom logic.

RHEL CoreOS doesn't support that functionality yet, but we should get it ported over.  Meanwhile, if you're feeling adventurous, it should be possible to use Fedora CoreOS to install RHEL CoreOS. :slightly_smiling_face: 

I'll close this.  It has too many caveats to support in the long term.

PR #28, opened 2019-05-13
coreos-installer: Honor redirects via curl

ashcrow:
Adds `-L` to all `curl` calls to honor redirects.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1709247

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -361,7 +361,7 @@ get_img_url() {
         fi
 
         IMAGE_URL=$(cat /tmp/image_url)
-        curl -sIf $IMAGE_URL >/tmp/image_info 2>&1
+        curl -LsIf $IMAGE_URL >/tmp/image_info 2>&1
         RETCODE=$?
         if [ $RETCODE -ne 0 ]
         then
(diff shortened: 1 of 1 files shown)


miabbott:
Looks sane to me  :+1:

ashcrow:
Setting WIP until after full testing.

dustymabe:
> Looks sane to me +1

agree. tagging in Jerry for a review since he's been handling the RHCOS side of the installer.

jlebon:
Looks sane to me as well. Might be worth factoring out some helper functions to not regress on this in the future, e.g. `http_head()` (for the `-I` case) and `http_get()` or something.

arithx:
lgtm

yuqi-zhang:
Hum, I created a redirect from localhost:8080->localhost:8081. If I do something like
`wget http://192.168.122.1:8080/r8-bios.raw.gz` I see the redirect and the GET request fine.

But even with this patch, with the installer, it doesn't actually do a GET for `8081`. It sees the 301 and says: oh that's fine, and proceeds to continue and then errors out after the reboot since it never got an image. It also can't get the ignition url either.

ashcrow:
Following up in a higher bandwidth environment to help debug.

yuqi-zhang:
Update: actually it will error for both (complains it can't find the image). I suspect its something else thats the problem and not just the `curl` acting up

ashcrow:
Removing WIP as @yuqi-zhang's testing has succeeded.

PR #31, opened 2019-05-17
coreos-installer: add check for image download exit status

### LINKED ISSUE #29: Not checking `curl` exit code
Just noticed this while reading over the codebase. Should `wait $curl_pid` or something to make sure we error out if `curl` dies halfway.

arithx:
Adds an additional check for the exit status of the curl call that does
the image download.

Closes #29

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -506,10 +506,9 @@ mount_tmpfs() {
 #########################################################
 download_image() {
     log "Downloading install image"
-    curl -L -s -o /mnt/dl/imagefile.gz $IMAGE_URL &
-    curlpid=$!
+    (curl -L -s -o /mnt/dl/imagefile.gz $IMAGE_URL; echo $? > /tmp/curl-rc) &
 
-    while ps --pid $curlpid > /dev/null; do
+    while [ ! -f /tmp/curl-rc ]; do
         # If the image hasn't show up yet then wait a sec and loop
         if [ ! -f /mnt/dl/imagefile.gz ]; then
             sleep 1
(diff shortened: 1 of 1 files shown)


coreos-installer
```diff
@@ -520,6 +520,14 @@ download_image() {
         echo "${PCT}%" >&2
         sleep 1
     done
+
+    wait $curlpid
```
jlebon: Aside, though the `ps --pid` trick isn't foolproof since a new process might take up that same pid.

Offhand another approach could be: `(curl ...; echo $? > /tmp/curl-rc) &`. Then the loop can be `while [ ! -f /tmp/curl-rc ]` and we can drop the `curlpid` var and the `wait` too.

dustymabe: hmm - never used  `wait` before. Does it work for processes that have already completed?

jlebon: Yeah, it'll just return immediately. Note it only works with child processes of the shell though (see `help wait`).

dustymabe: welp - learned something new today

dustymabe:
LGTM

coreos-installer
arithx: I liked the echo RC to file approach and have pushed a change switching to it.

dustymabe:
will run it through a test

arithx:
> will run it through a test

👍 it passed a simplified test set on my local libvirt test setup but I didn't stress all of the potential failure cases as I wasn't hosting a local webserver where I could hit some of the different failure states this should catch.

dustymabe:
seems to work in basic tests



> I didn't stress all of the potential failure cases

same

PR #36, opened 2019-05-21
coreos-installer: enable `pipefail` around `zcat | dd`

### LINKED ISSUE #24: warn/error if disk image is not gzipped
The installer appears to happily consume `.raw` disk images with no indication that anything is going wrong...until you try to boot your host.

It would be grand if the installer could warn the user or error out with some kind of message about requiring gzipped images.

Granted, I don't know if this would be useful in the automated case, but I figure it is worth starting the conversation here.

arithx:
enable `pipefail` around `zcat | dd`

Closes #24

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -611,8 +611,15 @@ log() {
 write_image_to_disk() {
     log "Writing disk image"
     
+    set -o pipefail
     zcat /mnt/dl/imagefile.gz |\
     dd bs=1M iflag=fullblock oflag=direct conv=sparse of="${DEST_DEV}" status=none
+    RETCODE=$?
+    if [[ $RETCODE -ne 0 ]]; then
+        log "failed to write image to disk"
+        exit 1
+    fi
+    set +o pipefail
 
     for try in 0 1 2 4; do
         sleep "$try"  # Give the device a bit more time on each attempt.


arithx:
NOTE: I've tested that `zcat -f` against raw files works correctly but haven't run it through an image build process yet.

jlebon:
Hmm, it looks like `-f` is a catch-all for a lot of different things. A downside of this too is that we now accept any file rather than only one of: raw data or gzip (e.g. the classic curl'ing a website's 404 HTML page when mistyping the address, though I guess that one specifically would've tripped up our `curl -f`).

Here's an idea: we check the output of `file --mime-type` for one of `application/x-gzip` or `application/octet-stream`. If the former, pass through `zcat`, if the latter, feed directly to `dd`, if anything else, error out.

jlebon:
I think part of the issue in #24 is likely because we're not enabling `-o pipefail` by default, so `zcat` failing didn't actually cause an error. We should probably turn that on.

ajeddeloh:
`file` is a pretty awful thing to ship. It has a history of CVEs (trying to find examples, but its name makes that hard).

arithx:
Hmm, I'll punt on supporting raw images from this PR then in favor of enabling `pipefail` around the `zcat`. Do you think it's worth manually checking the first two bytes of the downloaded file for `0x1f 0x8b` (https://tools.ietf.org/html/rfc1952#section-2.3.1) or is the `pipefail` sufficient?

bgilbert:
`coreos-install` does:

```
    if [[ "${IMAGE_FILE}" =~ \.bz2$ ]]; then
        bzip2 -cd "${IMAGE_FILE}" | write_to_disk
    else
        write_to_disk < "${IMAGE_FILE}"
    fi
```

Why not check the image URL for the appropriate file extension?

bgilbert:
@arithx, I'd just let zcat decide whether it's happy with its input.

arithx:
> `coreos-install` does:
> 
> ```
>     if [[ "${IMAGE_FILE}" =~ \.bz2$ ]]; then
>         bzip2 -cd "${IMAGE_FILE}" | write_to_disk
>     else
>         write_to_disk < "${IMAGE_FILE}"
>     fi
> ```
> 
> Why not check the image URL for the appropriate file extension?

We'd have to refactor the image download to respect the file name on the webserver. Right now we're always writing it to [`/mnt/dl/imagefile.gz`](https://github.com/coreos/coreos-installer/blob/master/coreos-installer#L527).

arithx:
Update pushed to back this PR down to just enabling `pipefail` around `zcat | dd`

dustymabe:
> This looks good to me! 

same

> (Any reason we can't just enable it once at the top of the script?)

I think that is probably the right move, though @arithx may be trying to be risk averse. If we do move it to the top there are other places in the file we can remove it I think.

arithx:
> (Any reason we can't just enable it once at the top of the script?)

Probably not, I just kept it in-line with https://github.com/coreos/coreos-installer/commit/666303d41cb6e06c6e31a239824c9819ca1e26a0

If you prefer for it to be applying to the entire file I can re-work this to do that and test it again.

jlebon:
Will leave it up to you! Didn't mean to keep piling onto this patch, hence why I LGTM'ed it. :) Feel free to leave as a follow-up.

arithx:
> Will leave it up to you! Didn't mean to keep piling onto this patch, hence why I LGTM'ed it. :) Feel free to leave as a follow-up.

👍 going to merge then, opened up https://github.com/coreos/coreos-installer/issues/38 for enabling pipefail in the rest of the script.

PR #40, opened 2019-05-28
module-setup.sh: Explicitly add gzip

jlebon:
`zcat` is just a thin shell wrapper around `gzip`. So we need to
explicitly add that to the initrd as well.

### DIFF
--- a/dracut/30coreos-installer/module-setup.sh
+++ b/dracut/30coreos-installer/module-setup.sh
@@ -22,6 +22,7 @@ install() {
     inst_multiple /usr/bin/dd
     inst_multiple /usr/bin/gpg2
     inst_multiple /usr/bin/grep
+    inst_multiple /usr/bin/gzip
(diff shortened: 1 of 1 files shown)


dracut/30coreos-installer/module-setup.sh
```diff
@@ -28,6 +28,7 @@ install() {
     inst_multiple /usr/bin/tail
     inst_multiple /usr/bin/tr
     inst_multiple /usr/bin/zcat
+    inst_multiple /usr/bin/gzip
```
arithx: Can you add this alphabetically?

jlebon: Done!

dustymabe:
> `zcat` is just a thin shell wrapper around `gzip`. So we need to
> explicitly add that to the initrd as well.

I had assumed it was some sort of symlink where the script itself was triggering off of `$0`, but it is indeed a shell script.

Issue #13, opened 2019-03-14
Support setting kernel args for the installed OS image


abhinavdahiya:
When PXE booting FCOS/RHCOS the coreos-installer provides options to install OS image and the Ignition config that should be used to configure on the reboot.

There are cases where users might want to provide kernel args for the OS image that is installed by coreos-installer.

An example is:

the FCOS OS image by default has `console=ttyS0,115200n8` as the kernel parameter. But when you try to PXE boot FCOS in Packet.net `console=ttyS0,115200n8` is required to get output on the [SOS console](https://support.packet.com/kb/articles/sos-serial-over-ssh)

also an Ignition config that needs to fetch extra configuration from network, `rd.neednet=1` is required so that Ignition has DHCP/network available in initramfs.

bgilbert:
Neither of those examples are completely on point: the kernel options for PXE-booted systems are controlled by the PXE server; the Packet platform support will include the correct console settings; and the `rd.neednet` requirement is a bug.

As much as I don't like the idea of users overriding kernel parameters, though, I think this is a good point.  @ajeddeloh, WDYT?

ajeddeloh:
I can see wanting to set the default kernel params for an installed to disk system on BM at install time. For cloud/PXE it doesn't make much sense. That being said the implementation for diskfull BM is hairy:

Currently, with anaconda created images there's not a good way to do this without booting the system.
With BLS images (working on this) there's also no easy way to do this (could edit the BLS fragments themselves but then it'll get lost on upgrade)
Eventually we could have a grub var like `$kernel_extra_args` or something read from a snippet in `/boot` but that's a ways off.

@cgwalters was talking at some point about ostree owning the kcmdline and having it be part of the commit, but I'd argue the user should be able to augment it without needing to create new ostree commits. (I agree the _defaults_ ought to be part of the commit).

dustymabe:
> I can see wanting to set the default kernel params for an installed to disk system on BM at install time.

yep. I asked @abhinavdahiya to open this issue after chatting with him. I thought it was a reasonable feature request.

bgilbert:
One use case for kargs: configuring SMT for mitigating L1TF and MDS.  That does apply more to bare metal than VMs, but bare-metal clouds that provide preinstalled images (e.g. Packet) will need it.  Thus, doing it from the installer isn't really enough; it should be possible to set kargs from Ignition.

cgwalters:
[...]

PR #41, opened 2019-05-30
parse-coreos:  persist ipv6.disable

miabbott:
Some users may want to disable IPv6 as part of firstboot.  This was
noticed in https://bugzilla.redhat.com/show_bug.cgi?id=1715205

### DIFF
--- a/dracut/30coreos-installer/parse-coreos.sh
+++ b/dracut/30coreos-installer/parse-coreos.sh
@@ -24,12 +24,18 @@ then
     echo $IGNITION_URL >> /tmp/ignition_url
 fi
 
+
+# Kernel networking args
+# Currently only persisting `ipv6.disable`, but additional options may be added
+# in the future
+# https://github.com/torvalds/linux/blob/master/Documentation/networking/ipv6.txt
+declare -a KERNEL_NET_ARGS=("ipv6.disable=")
 # Dracut networking args
 # Parse all args (other than rd.neednet) and persist those into /tmp/networking_opts
 # List from https://www.mankier.com/7/dracut.cmdline#Description-Network
 local NETWORKING_ARGS="rd.neednet=1"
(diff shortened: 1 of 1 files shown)


miabbott:
I've labeled this as WIP so we can discuss.  There maybe additional options on the RHCOS side we want to persist; they could be easily added to this PR.

dustymabe:
the reason we missed ipv6.disable before is because it's a kernel parameter and not a dracut parameter so it wasn't in the list. I think in order to be successful we're going to need an escape hatch.. which is https://github.com/coreos/coreos-installer/issues/13 I think. basically we need a way for the  user to be able to say specifically (rather than inferring) a few options that he/she wants to persist to the firstboot

dustymabe:
note that my previous statement doesn't mean we shouldn't merge this, but just noting we should be aware we'll probably hit a lot of these.

dracut/30coreos-installer/parse-coreos.sh
```diff
@@ -28,7 +28,7 @@ fi
 # Parse all args (other than rd.neednet) and persist those into /tmp/networking_opts
 # List from https://www.mankier.com/7/dracut.cmdline#Description-Network
```
dustymabe: can we mention here that not all of them come from dracut and some come from kernel: https://github.com/torvalds/linux/blob/master/Documentation/networking/ipv6.txt

maybe we even make a new array below for ones from the kernel and then we have `DRACUT_NET_ARGS` and `KERNEL_NET_ARGS` that we then combine on the for loop line.

miabbott: separate arrays sounds like a good idea!

miabbott:
@dustymabe pushed a change with two arrays.  It works in a test bash script; haven't tested the actual installer yet

miabbott:
Finally got around to testing this successfully with FCOS.

1.  built new ISO with this change
1.  installed FCOS with ISO, passing `ipv6.disable=1` as part of installer args
1.  freshly installed FCOS has `ipv6.disable=1` is on `/proc/cmdline`
1.  no IPv6 address on interface

Also did the test without the `ipv6.disable` option passed along and got expected results (IPv6 address on interface)

dustymabe:
thanks for testing!

Issue #42, opened 2019-05-30
Use bash strict mode


jlebon:
We should use `set -euo pipefail` right at the beginning, aka strict mode: http://redsymbol.net/articles/unofficial-bash-strict-mode/

Otherwise, random errors are just too easy to miss.

dustymabe:
agree - let's just do it and find any issues now

bgilbert:
No longer needed in `master` (#65), probably will not be fixed in `legacy`.

PR #46, opened 2019-06-20
Add support to xz-compressed images

zonggen:
This change checks magic bytes inside the header of compressed image files.
`gzip` compressed files are checked first 2 bytes while 6 bytes for `xz`
compressed files.

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -527,15 +527,15 @@ mount_tmpfs() {
 #########################################################
 download_image() {
     log "Downloading install image"
-    (curl -L -s -o /mnt/dl/imagefile.gz $IMAGE_URL; echo $? > /tmp/curl-rc) &
+    (curl -L -s -o /mnt/dl/imagefile.compressed $IMAGE_URL; echo $? > /tmp/curl-rc) &
 
     while [ ! -f /tmp/curl-rc ]; do
         # If the image hasn't show up yet then wait a sec and loop
-        if [ ! -f /mnt/dl/imagefile.gz ]; then
+        if [ ! -f /mnt/dl/imagefile.compressed ]; then
             sleep 1
             continue
         fi
-        PART_FILE_SIZE=$(ls -l /mnt/dl/imagefile.gz | awk '{print $5}') 2>/dev/null
+        PART_FILE_SIZE=$(ls -l /mnt/dl/imagefile.compressed | awk '{print $5}') 2>/dev/null
         PCT=$(dc -e"2 k $PART_FILE_SIZE $IMAGE_SIZE / 100 * p" | sed -e"s/\..*$//" 2>/dev/null)
         echo "${PCT}%" >&2
         sleep 1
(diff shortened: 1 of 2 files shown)


zonggen:
I see.. will be updating. Thank you for reviewing !

coreos-installer
```diff
@@ -613,9 +613,23 @@ log() {
 #########################################################
 write_image_to_disk() {
     log "Writing disk image"
-    
+
+    set -o pipefail
+    # Check for gzip and xz correspondingly
+    if [[ $(xxd -ps -l 2 /mnt/dl/imagefile.gz) = '1f8b' ]]
```
bgilbert: `xxd` doesn't exist on every system and doesn't exist in the initramfs.  But this should be possible with tools that do, e.g.:

```
[ "$(dd count=2 bs=1 if=/mnt/dl/imagefile.gz)" = "$(printf '\x1f\x8b')" ]
```

zonggen: Yes I was wondering if `xxd` is widely available, will be updating :+1:

coreos-installer
```diff
@@ -613,9 +613,23 @@ log() {
...
+    set -o pipefail
+    # Check for gzip and xz correspondingly
+    if [ "$(dd count=2 bs=1 if=/mnt/dl/imagefile.gz)" = "$(printf '\x1f\x8b')" ]
+    then
+        DECOMPRESSION_TOOL='zcat'
+    elif [ "$(dd count=6 bs=1 if=/mnt/dl/imagefile.gz)" = "$(printf '\xfd\x37\x7a\x58\x5a\x00')" ]
+    then
+        DECOMPRESSION_TOOL='xzcat'
+    else
+        log "Compressed file format is not supported."
```
jlebon: How about also mentioning in that error also the formats that *are* supported?

coreos-installer
```diff
@@ -613,9 +613,23 @@ log() {
 #########################################################
 write_image_to_disk() {
     log "Writing disk image"
-    
+
+    set -o pipefail
```
jlebon: We don't use any pipes here, so shouldn't have to toggle this on and off, right? (But ideally it'd always be on, that's https://github.com/coreos/coreos-installer/issues/42).

coreos-installer
```diff
@@ -613,9 +613,23 @@ log() {
...
+    # Check for gzip and xz correspondingly
+    if [ "$(dd count=2 bs=1 if=/mnt/dl/imagefile.gz)" = "$(printf '\x1f\x8b')" ]
```
jlebon: Minor/optional: maybe we should rename this filename to something more neutral? E.g. maybe just `imagefile`? Or `imagefile.compressed`?

coreos-installer
zonggen: Sure, thanks for pointing that out!

coreos-installer
zonggen: Yep, I was trying to make things consistent with the rest of the code. Thanks for the link!

coreos-installer
zonggen: Will be updating :ok_hand:

jlebon:
@zonggen To confirm, you successfully tested this in combination with https://github.com/coreos/coreos-assembler/pull/558 ?

zonggen:
Update:Tested along with https://github.com/coreos/coreos-assembler/pull/558, working as intended.

zonggen:
Update:
 - Updated on fedora rpm repo and created a pr (https://src.fedoraproject.org/rpms/coreos-installer/pull-request/7#)

PR #48, opened 2019-06-26
Provide options to create a RAID on ROOT disk

rsevilla87:
The option of creating software based RAID on ROOT disk should be available in the installer.
I'm open to suggestions about modification of the code :).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,13 @@ standalone script or during bootup via a dracut module.
 ## Kernel command line options for coreos-installer running in the initramfs
 
 * coreos.inst=yes - Instruct the installer to run
-* coreos.inst.install_dev - The block device on the system to install to
+* coreos.inst.install_dev - The block device on the system to install to. In case of 
+using an RAID device it should be a comma separated list of block devices names
 * coreos.inst.image_url - The url of the coreos image to install to this device
 * coreos.inst.ignition_url - The url of the coreos ignition config (optional, enter
   coreos.inst.ignition_url=skip to not load an ignition config)
+* coreos.inst.raid_dev - The RAID device name (optional)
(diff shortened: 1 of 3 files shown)


bgilbert:
We explicitly don't want to support this sort of thing in the installer, since storage configuration should be possible on every platform and the installer doesn't run in the cloud.  Instead, storage configuration should be performed by Ignition, which runs on every platform.  Ignition already supports creating RAID volumes for data disks, but we don't currently support root on RAID; see https://github.com/coreos/fedora-coreos-tracker/issues/94.

cgwalters:
Restating in a different way bgilbert's reply: This is starting to create a kind of distinct mini-language/config format for partitioning/storage (like Kickstart) - from my ("old school RHT") view, a major benefit of Ignition is unifying Kickstart and cloud-init.  Then for RHCOS, we support "day 2" management of Ignition via MachineConfig.

One could imagine in the future even a model where the MCO supports re-configuring the partitioning for extant nodes (maybe via re-provisioning at an IaaS level, maybe in-place if we have free raw partitions space w/LVM), etc.  We can't do that type of stuff if the state of the system is invisible to it (i.e. it's not encoded in Ignition).

xguitian:
@cgwalters but ignoring that is not yet supperted on Fedora CoreOS, do you really envision a to move root to a RAID _after_ install?

cgwalters:
> @cgwalters but ignoring that is not yet supperted on Fedora CoreOS, do you really envision a to move root to a RAID after install?

Yes, see https://github.com/coreos/ignition-dracut/pull/107

bgilbert:
Thanks for the PR.  The installer isn't the right place for this sort of functionality, and we have no plans to add it here.  I'll go ahead and close.

PR #50, opened 2019-07-09
*: persist `coreos.no_persist_ip=`

arithx:


dustymabe:
I didn't understand this til I found a breadcrumb trail starting here: https://github.com/coreos/coreos-installer/pull/51#issuecomment-513387495

should we add a comment about this not being used for FCOS and we can remove it one day?

PR #51, opened 2019-07-16
Persist ip config if ip specified, skip ip persistance when `coreos.inst.no_persist_ip` is set

vrutkovs:
Ref: https://jira.coreos.com/projects/RHCOS/issues/RHCOS-363

If `ip` kernel arg is specified in installer it would be persisted by generating `ifcfg` file. This can be disabled by setting `coreos.inst.no_persist_ip=1`
Complex configurations - such as `bridge` would be ignored for now. A more complex configuration settings should be injected in ignition file.

This code would generate ifcfg file if `ip=` param is specified (similar to dracut's writeifcfg module). These ifcfg files are being urlencoded and included in `storage.files` section of the supplied Ignition config

TODO:
* [x] Skip complex configurations (e.g. bridges)
* [x] Skip if Ignition config not set
* [x] Rework to use `ip=` instead and implement `coreos.inst.no_persist_ip`
* [x] Sort out why a broken ignition is being generated


vrutkovs:
PTAL @darkmuggle @arithx

vrutkovs:
>Have you tested this? I don't think that these files would be persisted to the real OS filesystem.

Not, not yet. Perhaps it would be best to generate ignition part for these files and amend or create `$bootmnt/ignition/config.ign`?

darkmuggle:
The installer will get pulled in during the RHCOS build, so this would be the right place for it.

arithx:
> The installer will get pulled in during the RHCOS build, so this would be the right place for it.

I'm not sure I follow, none of this code would be ran if doing a live PXE.

vrutkovs:
>I don't think that these files would be persisted to the real OS filesystem.

It seems so, I can mount `/ostree`, but `/etc/` cannot be directly modified.

Added a few more commits to generate `/boot/ignition/network.ign`, which would be applied by ignition on boot. If everyone agrees with the approach I'll create a new PR to ignition-dracut to apply this file

coreos-installer
```diff
@@ -513,6 +513,137 @@ get_ignition_url() {
...
+    log "Persisting provided network configs"
+    mkdir -p /tmp/ifcfg/
+
+    # Prepare nameservers list
+    NAMESERVERS=()
+    if [ -f /tmp/nameserver ]; then
+        NAMESERVERS=($(cat /tmp/nameservers))
+    fi
+
+    echo '{"ignition": {"version": "2.2.0"}, "storage": { "files": [' > /tmp/network.ign
```
vrutkovs: Hardcoding 2.2.0 version is probably a bad idea, although FCOS should be able to parse that, right?

arithx: No it can't.

vrutkovs:
I must have overlooked the comment about diskless PXE (also this solution would hardcode Ignition version), closing the PR

vrutkovs:
It seems it would work on PXE too.

Removed hardcoded Ignition version - instead the generated files are being included in "files" section of the existing ignition config

ashcrow:
@vrutkovs this is great work, thank you for putting it together! I do think there are few things which may be able to be tightened up for UX/parser gotchas:

1. Persisting ip configuration and nameservers, at least in my mind, should be assumed. Instead, I think having something like `coreos.inst.nopersistip` and `coreos.inst.nopersistns` booleans would be a bit cleaner. This would mean that, if someone provided ip or nameserver items they would be persisted by default. However, if they want to avoid persisting the configuration they could disable it via those (or similar) booleans.

2. Right now we're using both [Ignition v2](https://coreos.com/ignition/docs/latest/configuration-v2_2.html) (RHCOS) and v3 (FCOS) which imposes somewhat of a maintenance burden on tooling shared between composes. Using text replacement for JSON data can lead to malformed output. If that occurs, installation will fail and, for many users, it will be quite tough to debug as to why. I feel that taking the ip/ns configs, making them valid Ignition, and then using Ignitions methods for appending/merging would likely be a safer solution as it promises the results will be valid JSON and a valid Ignition configuration.

arithx:
On the topic of doing merged/appended configs, here's an example approach.

This approach assumes that something is added to the `coreos-installer` specfiles that specifies an Ignition version to `/etc/ignition_version` depending on the OS it's being built for (`v0` for RHCOS, `v2` for FCOS in this example).

Do also note that the files schema is slightly different in Ignition v2 (spec 3.0.0) in that it doesn't allow a `filesystem` key (it figures out the relevant filesystem from the path) for example:

Example Ignition spec 2.2.0 file:
```
{
  "filesystem": "root",
  "path": "/foo/bar",
  "mode": 420,
  "contents": { "source": "data:,example%20file%0A" }
}
```

Example Ignition spec 3.0.0 file:
```
{
  "path": "/foo/bar",
  "mode": 420,
  "contents": { "source": "data:,example%20file%0A" }
}
```

And something like this could be used to build the new `/tmp/ignition.ign` (the previous contents will be put into the relevant `append`/`merge` functions)

```
local OUT_CFG
case $(cat /etc/ignition_version) in
"v0")
  log "generating an Ignition v2.2.0 appended config w/ ifcfg files"
  OUT_CFG=`{
    "ignition": {
      "version": "2.2.0",
      "config": {
        "append": [{
          "source": "data:,$(urlencode <(/tmp/ignition.ign)"
        }]
      }
    },
    "storage": {
      "files": $files_section
    }
  ;;
"v2")
  log "generating an Ignition v3.0.0 merged config w/ ifcfg files"
  OUT_CFG=`{
    "ignition": {
      "version": "3.0.0",
      "config": {
        "merge": [{
          "source": "data:,$(urlencode <(/tmp/ignition.ign)"
        }]
      }
    },
    "storage": {
      "files": $files_section
    }
  }`
  ;;
*)
  log "unknown distribution: unable to generate ifcfg files"
  exit 1
esac
echo $OUT_CFG > /tmp/ignition.ign
```

Disclaimer: I haven't tested the above code

vrutkovs:
>This would mean that, if someone provided ip or nameserver items they would be persisted by default

I'd prefer explicit approach, but it can be reverted

>Using text replacement for JSON data can lead to malformed output

Agree, I'll rework this to implement a suggestion from https://github.com/coreos/coreos-installer/pull/51#issuecomment-512943587

vrutkovs:
>This would mean that, if someone provided ip or nameserver items they would be persisted by default

Implemented that, there is no need to duplicate there params (otherwise the machine won't start without DHCP).

This is still WIP, as the test machine won't start - probably the generated ignition is incorrect?

ajeddeloh:
This wont work on FCOS both because we use ignition spec 3 and because we don't support ifcfg configuration. We ought to at least default it to false instead of true and add a note in the readme saying it does not work on fcos.

arithx:
This functionality has landed in #50 & coreos/ignition-dracut#89

It also needs a line in the base-config for the distro which enables the dracut `ifcfg` module (not done for FCOS as it doesn't enable the NetworkManager module to support ifcfg's).

vrutkovs:
No longer required, solved by RHCOS dracut modules instead

PR #52, opened 2019-07-18
Read Ignition platform ID and write to separate file

zonggen:
This change makes coreos-installer reads ignition platform id from command line option `-p` and appends the value to ignition.firstboot. Also reads Ignition platform ID from kcmdline then to /tmp/platform_id, which will then be picked up and written to ignition.firstboot.

Edit: Now writes to the file `ignition_platform_id`.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/191

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ standalone script or during bootup via a dracut module.
 * coreos.inst.image_url - The url of the coreos image to install to this device
 * coreos.inst.ignition_url - The url of the coreos ignition config (optional, enter
   coreos.inst.ignition_url=skip to not load an ignition config)
+* coreos.inst.platform_id - The ignition platform ID the coreos image to install to
 
 ## Using the installer on FCOS or RHCOS
 
@@ -100,6 +101,7 @@ kernel command line telling it what you want it to do. For example:
 - `coreos.inst.install_dev=sda`
 - `coreos.inst.image_url=http://example.com/fedora-coreos-30.107-metal-bios.raw.gz`
 - `coreos.inst.ignition_url=http://example.com/config.ign`
+- `coreos.inst.platform_id=qemu`
 
 **NOTE** make sure to use a `metal-uefi` image if booting via UEFI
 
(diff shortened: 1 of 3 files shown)


zonggen:
Update:
 - Currently there's no default platform ID added to ignition.firstboot, should I grab the one inside `/proc/cmdline` and append to ignition.firstboot or add nothing if user does not specify platform ID.
 - From the discussion of https://github.com/coreos/fedora-coreos-tracker/issues/191#issuecomment-499656531, I'm assuming I don't need to remove the `ignition.platform.id=` from `/proc/cmdline`?

zonggen:
Update: 
 - Default platform id grep'd from `proc/cmdline`

zonggen:
Update:
 - PR on `cosa` reading platform id from the file: https://github.com/coreos/coreos-assembler/pull/644
 - PR on removing default ignition platform id from `cosa` : https://github.com/coreos/coreos-assembler/pull/645

ajeddeloh:
Shouldn't the image building process be writing the platform id, not the installer?

zonggen:
I think by modifying `coreos-installer`, users would have the flexibility to overwrite the default platform id which was set inside `cosa`. And in https://github.com/coreos/fedora-coreos-tracker/issues/191#issue-450545721, I thought we want to separate the platform id from inline kargs and write to a separate file. What @bgilbert suggested would be remove writing platform id completely from `cosa` and let `installer` to handle the platform id part.

arithx:
The comment you linked mentions that `coreos-installer` should be able to overwrite the argument. It will still need to be set in `cosa` as in most use cases `coreos-installer` is never ran (e.x. `qemu` or any cloud image)

zonggen:
That sounds correct, setting it inside `cosa` and overwrite it in installer if necessary.

dustymabe:
FYI furthering discussion: https://github.com/coreos/fedora-coreos-tracker/issues/191#issuecomment-519198802

zonggen:
Closing this and opening a new PR since we now use `sed` to overwrite BLS as a temporary solution and no long write `platform id` to a separate file.

PR #55, opened 2019-08-13
add rd.net.timeout.carrier parameter to list of kernel opts for first boot

fsperling:
default rd.net.timeout.carrier is 5 seconds. in our dell hardware servers the network cards take a bit longer to come up and this causes the installation to hang forever.

the rd.net.timeout.carrier parameter can be provided during the installation of coreos. unfortunately the parameter is not passed on to the first boot after the installation. 

the pull request add the parameter to the list of params that are passed on.

alternatively the default value could be increased. probably it also would be good if all parameters are passed on to the first boot after the installation.

### DIFF
--- a/dracut/30coreos-installer/parse-coreos.sh
+++ b/dracut/30coreos-installer/parse-coreos.sh
@@ -34,7 +34,7 @@ declare -a KERNEL_NET_ARGS=("ipv6.disable=")
 # Parse all args (other than rd.neednet) and persist those into /tmp/networking_opts
 # List from https://www.mankier.com/7/dracut.cmdline#Description-Network
 local NETWORKING_ARGS="rd.neednet=1"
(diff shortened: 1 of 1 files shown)


dustymabe:
Looks good to me.. cc @arithx @yuqi-zhang

dustymabe:
> LGTM. Minor nit: could you edit the git commit message to have a shorter title and add some context? (Why this is needed, or how to use it, etc.)

agree.. the commit doesn't have the context that the PR description has

fsperling:
@yuqi-zhang @dustymabe Thanks for the reviews. I put more context into the commit message.

dustymabe:
cc @yuqi-zhang

yuqi-zhang:
@dustymabe is this in need of an immediate rebuild anywhere?
(Sorry I missed the original update)

dustymabe:
not immediate, just wanted to make sure we maybe pick it up in the next shipment

PR #59, opened 2019-08-30
Check for dependencies before installer execution

fhemberger:
When you run `coreos-installer` from a live CD (e.g. Debian), some required
packages might not be installed by default. In case of a missing `curl`for
example, the resulting error message is misleading: "failed fetching image
headers from url" makes it sound like a network issue instead of a missing
dependency.

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -406,6 +406,15 @@ import_gpg_key() {
 }
 
 ############################################################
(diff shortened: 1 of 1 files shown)


fhemberger:
@bgilbert @dustymabe @zonggen Hi, could you PTAL?

PR #61, opened 2019-09-05
s390x: add support for zFCP SCSI and ECKD DASD devices on zVM and LPAR 

tuan-hoang1:


tuan-hoang1:
Goes along with : https://github.com/coreos/coreos-assembler/pull/732/

alicefr:
@barthy1

dracut/30coreos-installer/parse-coreos.sh
```diff
@@ -52,6 +52,24 @@ then
...
+    S390X_OPT=$(getarg $S390X_ARG)
+    if [ ! -z "$S390X_OPT" ]
+    then
+        echo "persist $S390X_ARG to $S390X_OPT" >> /tmp/debug
+        S390X_ALL_ARGS+=" ${S390X_ARG}${S390X_OPT}"
+    fi
+done
+if [ -n "${S390X_ALL_ARGS}" ]
+then
+    echo "persisting network options: ${S390X_ALL_ARGS}" >> /tmp/debug
```
tuan-hoang1: copy pasta

coreos-installer
```diff
@@ -626,6 +630,65 @@ log() {
...
+        fi
+        echo $line
+    done <<< $(sfdisk -d /mnt/dl/imagefile)
+    ) | sfdisk "${DEST_DEV}"
+
+    while IFS='' read -r line || [ -n "$line" ]
+    do
+        eval ${line%%:*} ${line##*:}
+        # NOTE: we hardcode block size to 512 for current zFCP SCSI devices
+        dd if=/mnt/dl/imagefile bs=512 iflag=fullblock \
```
tuan-hoang1: Is it because there is no `oflag=direct` or a `$ sync` call so that lsblk fails to list UUID ?

tuan-hoang1:
Should be ready by now.

coreos-installer
```diff
@@ -626,6 +630,65 @@ log() {
...
+    log "Extracting disk image"
+    # extract and remove downloaded image
+    $DECOMPRESSION_TOOL /mnt/dl/imagefile.compressed > /mnt/dl/imagefile
+
+    (while IFS='' read -r line || [ -n "$line" ]
+    do
+        if [[ $line =~ "start=" ]]
+        then
+            # we save the start= and size= here, then dd the partitions later
+            printf "%s:%s\n" \
```
tuan-hoang1: no need to do this, just get a space instead of colon.

tuan-hoang1:
Aha ! I just discovered that `sgdisk -R` allows replicating a whole GPT disk ! Much cleaner, maybe also good for x86 too.

tuan-hoang1:
Let's put this on hold until https://github.com/coreos/coreos-assembler/pull/738 get merged. We might need to make many changes with it coming in.

dracut/30coreos-installer/module-setup.sh
```diff
@@ -35,6 +35,15 @@ install() {
     inst_multiple /usr/sbin/blockdev
     inst_multiple /usr/sbin/wipefs
 
+    if [ "$arch" = "s390x" ]; then
```
arithx: Where is `$arch` being set for this, is it a dracut primitive?

coreos-installer
```diff
@@ -664,31 +682,87 @@ log() {
...
+        # this allows a reboot into new installed disk
+        chreipl "${DEST_DEV}"
+    fi
+}
+#########################################################
+#Handle s390x zFCP SCSI devices
+#########################################################
+write_image_to_zfcp_disk() {
+    log "Extracting disk image"
+    mount_tmpfs $(( $($DECOMPRESSION_TOOL ${imagefile_compressed} | wc --bytes) / 1024 / 1024 + ${TMPFS_MBSIZE} ))
```
arithx: Probably want to add some additional padding.

coreos-installer
```diff
@@ -664,31 +682,87 @@ log() {
...
+
+    (while IFS='' read -r line || [ -n "$line" ]
+    do
+        if [[ $line =~ "start=" ]]
+        then
+            # we save the start= and size= here, then dd the partitions later
+            printf "%s:%s\n" \
+                "$(echo $line | cut -d: -f 2 | cut -d, -f 1 | tr -d ' ')" \
+                "$(echo $line | cut -d: -f 2 | cut -d, -f 2 | tr -d ' ')" \
+                >> /tmp/sfdisk_info
```
arithx: Why write these to a temp file that will have to be parsed back in rather than just storing them inside of an array?

coreos-installer
```diff
@@ -664,31 +682,87 @@ log() {
...
-    RETCODE=$?
-    if [[ $RETCODE -ne 0 ]]; then
-        log "failed to write image to disk"
-        exit 1
+        case $(systemd-detect-virt) in
+            none|zvm) # 'none' means LPAR
+                write_image_to_zfcp_disk
+                ;;
+            qemu) log "Use qcow images for KVM s390x installation instead"; exit 1 ;;
+        esac
```
arithx: Are `none`, `zvm`, & `qemu` the only possible virt options for s390x or should a generic case also be added with a default message?

coreos-installer
```diff
@@ -574,8 +575,13 @@ get_ignition_url() {
 #########################################################
 mount_tmpfs() {
     log "Mounting tmpfs"
+    if [ -n "$1" ]; then
+        options="remount,size=${1}m"
```
arithx: Why the `remount`?

dracut/30coreos-installer/module-setup.sh
tuan-hoang1: Yes, `$arch` is inherited from dracut scripts. It is widely used in dracut itself.

coreos-installer
tuan-hoang1: Good point. I'm redoing this.

coreos-installer
tuan-hoang1: Trying to get a proper number for padding ...

coreos-installer
tuan-hoang1: If 1 argument is provided, assuming it's new size for `/mnt/dl`, `remount` is required to resize to new size.

coreos-installer
tuan-hoang1: According to man (https://www.freedesktop.org/software/systemd/man/systemd-detect-virt.html), and assuming we are not running coreos-installer in containers (could rule out with `systemd-detect-virt --vm`), I think there are only 4 possible options for s390x: 3 above pluses `qemu` (hw emulation, aka only qemu no KVM). Since s390x hw emulation is kind of something we are not allowed to talk about at the moment, not sure if I should include.

tuan-hoang1: LOL I did include `qemu`, which was supposed to be `kvm` instead.

coreos-installer
tuan-hoang1: @arithx : Just in case it's not clear: At this line, tmpfs at `/mnt/dl` is already mounted, and compressed image was downloaded and stored there. We need to extract the image to read its partition table, thus requiring to remount `/mnt/dl` with larger size.

coreos-installer
```diff
@@ -572,24 +575,29 @@ get_ignition_url() {
...
+    else
+        options="size=${TMPFS_MBSIZE}m"
+    fi
     mkdir -p /mnt/dl
-    mount -t tmpfs -o size=${TMPFS_MBSIZE}m tmpfs /mnt/dl
+    mount -t tmpfs -o "$options" tmpfs /mnt/dl
```
ashcrow: nit: It might be a good idea to change `/mnt/dl` into a variable and reuse it. Higher up `imagefile_compressed` and other vars are set under `/mnt/dl`.

tuan-hoang1: Agreed.

coreos-installer
```diff
@@ -662,31 +682,135 @@ log() {
     echo "$1" >&2
 }
 
+#########################################################
+#Update bootloader if needed
+#########################################################
+write_bootloader() {
```
ashcrow: nit: `write_bootloader` is, at least currently, `s390x` specific. This may be a little confusing since boot loader functionality may not always come from this function (EG: see `x86_64`). I suggest either renaming the function. Possibly something like `write_special_bootloader` or `write_arch_dependent_bootloader`, etc..  or something much better than those :smile: Just a suggestion.

dracut/30coreos-installer/module-setup.sh
```diff
@@ -35,6 +35,19 @@ install() {
...
+        inst_multiple -o /usr/bin/wc
```
ashcrow: This is totally fine, though I'm curious, why use `inst_multiple` for single installs?

coreos-installer
```diff
@@ -664,31 +682,87 @@ log() {
...
+    do
+        eval ${line%%:*} ${line##*:}
+        dd if=${imagefile} bs=512 iflag=fullblock \
+            of="${DEST_DEV}" status=progress skip=$start seek=$start count=$size
+        if [[ $? -ne 0 ]]; then
+            log "failed to write image to zFCP SCSI disk"
+            exit 1
+        fi
+    done < /tmp/sfdisk_info
+}
```
tuan-hoang1: I tried briefly with `sgdisk -R` which copies the whole disk image including GPT partition table, but was not very sure if it's a sane thing to do. Needs more checking (with s390x team about zFCP SCSI device layout, sgdisk code, and compare data on disk vs image after running `sgdisk -R` using hex dump tool) if this is desired. Advantage is we don't have to extract the image.

ashcrow:
cc @bgilbert @imcleod

coreos-installer
tuan-hoang1: Ah now I remember why I did this: because this while-do is executed inside a block `( )` so any variable inside that chunk will not be available outside that block.

dracut/30coreos-installer/module-setup.sh
tuan-hoang1: I want to use `-o` option of `inst_multiple` which means, if the file does not exist, just ignore it. That was for when I didn't use the "$arch" check, still left it there. Will change and test again.

tuan-hoang1: Actually, `inst_multiple` was used multiple times in this same file. Even though the full path of the binaries were provided, I also don't know why `inst_multiple` is used from the beginning.  Since I don't want to break something I'm not aware of, I would prefer to keep it like this.

coreos-installer
tuan-hoang1: I cannot come up with a good reasoning for a number, so I added `100` for padding ...

coreos-installer
tuan-hoang1: I add check for 4 of those possibilities. Closing this.

coreos-installer
tuan-hoang1: Resolving.

dracut/30coreos-installer/module-setup.sh
tuan-hoang1: Resolving.

coreos-installer
ashcrow: The variable could be created at a higher scope, and then used, but I follow. It may be a good idea to add a comment to that effect. I could see folks looking at this in 3-6 months and getting a little tripped up :slightly_smiling_face:.

coreos-installer
tuan-hoang1: I made it to `write_zipl_bootloader` and update the comment also.

coreos-installer
tuan-hoang1: @ashcrow : but iirc any change to the variable inside the block would not be remembered outside. Adding a comment.

tuan-hoang1:
cc @arithx for 2nd review pass of existing unresolved discussion.

tuan-hoang1:
Last force push to add a comment for : `/tmp/sfdisk_info`

coreos-installer
arithx: Hmm, is there a reason that the `dd` couldn't just be done inside of this loop?

tuan-hoang1: @arithx : The reason is : output of this loop will look like this :

arithx: Oh, whoops, missed the `sfdisk` pipe at the end.

coreos-installer
tuan-hoang1: Feel free to resolve or add more comments, many thanks.

coreos-installer
tuan-hoang1: Feel free to resolve or add more comments, many thanks.

tuan-hoang1:
I think I have answered and fixed most of the comments.

ashcrow:
Again, @tuan-hoang1, thank you for doing this work! What were the results of your s390x and x86 tests?

tuan-hoang1:
I don't see any issue till now on both platforms.

ashcrow:
> I don't see any issue till now on both platforms.

As in you just hit an issue with the latest changes? Or your tests didn't find any issues with the current set of changes?

tuan-hoang1:
Oh I didn't see any issue with current commits, sorry was not so clear. So basically, if you don't find any issue neither on your environment, I think we are good to go.

tuan-hoang1:
But be aware that my x86 "bare metal" is actually VM booted using virt-manager and attach ISO file. I don't have x86 hardware to, for example, burn the ISO to usb device, plug in a PC and test.

tuan-hoang1:
@ashcrow , @arithx : Since this is still not merged, I add 2 more commits that depend on previous 4 commits.
cc @cgwalters  @martinezjavier for reviewing new commits per our discussion.

cgwalters:
Once procedural request - this should target the `legacy` branch, see https://github.com/coreos/coreos-installer/pull/70#issuecomment-547584094

coreos-installer
```diff
@@ -662,31 +685,150 @@ log() {
...
 #########################################################
 #And Write the image to disk
 #########################################################
 write_image_to_disk() {
     log "Writing disk image"
 
     set -o pipefail
-    # Check for gzip and xz correspondingly
-    if [ "$(dd count=2 bs=1 if=/mnt/dl/imagefile.compressed status=none)" = "$(printf '\x1f\x8b')" ]
+    if [[ "$arch" == "aarch64" || "$arch" == "ppc64le" || "$arch" == "x86_64" ]]
```
ashcrow: I'd suggest reversing this OR catching when we have an unknown arch.

Examples:

```bash
if [[ "$arch" == "s390x" ]]; then
...
else
...
fi
```

or 

```bash
if [[ "$arch" == "aarch64" || "$arch" == "ppc64le" || "$arch" == "x86_64" ]]; then
...
elif [[ "$arch" == "s390x" ]]; then
...
else
...
fi
```

tuan-hoang1:
@ashcrow : I added this block and rebased:
```
diff --git a/coreos-installer b/coreos-installer
index b653ac3..3feadcf 100755
--- a/coreos-installer
+++ b/coreos-installer
@@ -768,6 +768,9 @@ write_image_to_disk() {
                 ;;
             qemu) log "Use qcow images for KVM s390x installation instead"; exit 1 ;;
         esac
+    else
+        log "Unsupported architecture"
+        exit 1
     fi
     set +o pipefail
```
@cgwalters : done, thanks.

tuan-hoang1:
@imcleod @dbenoit17: last commit is about the topic we talked today in the meeting.
Some users wanted to add multiple instances of `rd.znet=` to activate multiple NICs (qeth, QDIO Ethernet). They also wanted to add multiple instances of `ip=` accordingly but I suggested it's better to have only 1 instance of `ip=` and they should config other interface via Ignition. Maybe we need to patch Ignition v2 's `networkd` to understand syntax like this:

```

**Comment by tuan-hoang1** [comment]
Hi, is there anything that I can do from my side ?
I don't even mind re-writing this in another way that maybe friendlier to read. Please let me know.

**Comment by ashcrow** [comment]
I'd suggest rebasing one more time and we can do a final review. At the moment there are conflicts.

**Comment by tuan-hoang1** [comment]
> I'd suggest rebasing one more time and we can do a final review. At the moment there are conflicts.

Rebased, thanks !

I also push 1 additional commit compared to the state last time we talked. This still makes sense to this PR so I put it here too.

**Comment by ashcrow** [comment]
Added a few more folks to help get a second set of eyes for review.

**Comment by tuan-hoang1** [comment]
cc @Prashanth684 @dbenoit17 @andymcc who have been using this code in RHCOS on z/VM  for some time now. Last commit was for David's question on what image can be used to install KVM as bare metal disk.

File: coreos-installer [hunk 14]
```diff
@@ -723,8 +724,14 @@ write_image_to_dasd() {
 
     # the first 2 tracks of the ECKD DASD are reserved
     first_track=2
-    boot_partition=($(fdisk -b 4096 -o DEVICE,START,SECTORS -l ${imagefile} | grep "${imagefile}1" | awk '{print $2,$3}'))
-    root_partition=($(fdisk -b 4096 -o DEVICE,START,SECTORS -l ${imagefile} | grep "${imagefile}4" | awk '{print $2,$3}'))
+    # in RHCOS4.2 using anaconda, we have root partition at index 2 while in FCOS it's 4
```
tuan-hoang1: I don't think this is a good thing to do when RHCOS 4.3 goes out - which matches FCOS. But for the time being, I don't have time/other way to do it.

dbenoit17:
I will test this against the latest commit and follow-up.  Otherwise looks good to me so far.

ashcrow:
cc @darkmuggle @miabbott

zonggen:
Tested by building local FCOS and RHCOS with installer ISO with commits and metal raw image.
 Built both FCOS and RHCOS successfully and the installer worked as intended for x86_64.

ashcrow:
Unless @darkmuggle or @miabbott disagrees I think this should merge.

dbenoit17:
I have tested with all but the abf14cf commit and everything works on s390x as expected.

ashcrow:
@tuan-hoang1 any objection to me merging this?

coreos-installer
```diff
@@ -662,30 +685,146 @@ log() {
...
+            count="$2" \
+            seek="$(( $3 * $block_per_tracks ))"
+        if [[ $? -ne 0 ]]; then
+            log "failed to write image to ECKD DASD device"
+            exit 1
+        fi
+        shift 4
+    done
+}
+#########################################################
```
darkmuggle: Whitespace, please.

coreos-installer
```diff
@@ -662,30 +685,146 @@ log() {
...
+        zipl --verbose -p /tmp/zipl_prm -i $(ls ${boot_partition}/ostree/*/*vmlinuz*) -r $(ls ${boot_partition}/ostree/*/*initramfs*) --target "$boot_partition"
+        if [[ $? -ne 0 ]]; then
+            log "failed updating bootloder"
+            exit 1
+        fi
+        # this allows a reboot into new installed disk
+        chreipl "${DEST_DEV}"
+    fi
+}
+#########################################################
```
darkmuggle: whitespace, please

coreos-installer
```diff
@@ -662,30 +685,146 @@ log() {
...
 #########################################################
 #And Write the image to disk
 #########################################################
 write_image_to_disk() {
     log "Writing disk image"
 
     set -o pipefail
-    # Check for gzip and xz correspondingly
-    if [ "$(dd count=2 bs=1 if=/mnt/dl/imagefile.compressed status=none)" = "$(printf '\x1f\x8b')" ]
+    if [[ "$arch" == "aarch64" || "$arch" == "ppc64le" || "$arch" == "x86_64" ]]
```
darkmuggle: `if [[ "${arch}" =~ (aarch64|ppc64le|x86_64) ]]` might be a bit clearer.

coreos-installer
tuan-hoang1: Checked, thanks !

coreos-installer
tuan-hoang1: Checked, thanks !

coreos-installer
tuan-hoang1: Checked, thanks !

tuan-hoang1:
Thanks @dbenoit17 @zonggen for testing !
Let's do it @ashcrow !

PR #62, opened 2019-09-17
Add tool to embed Ignition config in live ISO image

bgilbert:
Part of https://github.com/coreos/fedora-coreos-config/pull/171.

### DIFF
--- a/coreos-iso-embed-ignition
+++ b/coreos-iso-embed-ignition
@@ -1,0 +1,197 @@
+#!/usr/bin/python3
+#
+# coreos-iso-embed-ignition - embed an Ignition config in a CoreOS live ISO
+# Copyright (C) 2019 Red Hat
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import argparse
+import gzip
+import os
+import shutil
+import struct
+import subprocess
+import sys
+import tempfile
+
+FILENAME = 'config.ign'
+
+class EmbedException(Exception):
+    pass
+
+
+class CopiedFileHolder:
+    '''
+    Holder for a file handle which can optionally be copied to another one,
+    then modified, then finalized.  If copied but not finalized, the copy
+    will be deleted when GCed.
+    '''
+    def __init__(self, fh):
+        self.fh = fh
+        self._copied = False
+        self._complete = False
+
+    def __del__(self):
+        if self._copied and not self._complete:
+            os.remove(self.fh.name)
+
+    def copy_to(self, target):
+        assert not self._copied
+        assert not self._complete
+        src = self.fh
+        self.fh = target
+        self._copied = True
+        shutil.copyfileobj(src, target)
+
+    def finish(self):
+        self._complete = True
+
+
+def make_cpio(ignition_fh):
+    '''
+    Make a gzipped CPIO archive containing the specified Ignition config.
+    '''
+    with tempfile.TemporaryDirectory(prefix='coreos-iso-') as tempdir:
+        config_path = os.path.join(tempdir, FILENAME)
+        with open(config_path, 'w') as out_fh:
+            shutil.copyfileobj(ignition_fh, out_fh)
+        proc = subprocess.Popen(['cpio', '-o', '-D', tempdir, '-H', 'newc',
+                '--quiet'], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        archive, _ = proc.communicate(FILENAME.encode())
+        if proc.returncode != 0:
+            raise EmbedException(f'cpio -o exited with status {proc.returncode}.')
+    return gzip.compress(archive)
+
+
+def extract_cpio(buf):
+    '''
+    Extract a gzipped CPIO archive and return the contents of the Ignition
+    config.
+    '''
+    cpio = gzip.decompress(buf)
+    proc = subprocess.Popen(['cpio', '-i', '--quiet', '--to-stdout',
+            FILENAME], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    config, _ = proc.communicate(cpio)
+    if proc.returncode != 0:
+        raise EmbedException(f'cpio -i exited with status {proc.returncode}.')
+    return config
+
+
+def get_location(iso_fh):
+    '''
+    Given an ISO image, return the offset and length of the Ignition
+    embed area.
+    '''
+    fmt = '<8s2Q'
+    size = struct.calcsize(fmt)
+    # Read the last bytes of the ISO 9660 System Area
+    iso_fh.seek(32768 - size)
+    buf = iso_fh.read(size)
+    if len(buf) < size:
+        raise EmbedException('Short ISO image file.')
+    magic, offset, length = struct.unpack(fmt, buf)
+    if magic != b'coreiso+':
+        raise EmbedException('Unrecognized CoreOS ISO image.')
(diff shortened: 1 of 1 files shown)


coreos-iso-embed-ignition
```diff
@@ -0,0 +1,172 @@
...
+
+def extract_cpio(buf):
+    '''
+    Extract a gzipped CPIO archive and return the contents of the Ignition
+    config.
+    '''
+    cpio = gzip.decompress(buf)
+    with tempfile.TemporaryDirectory(prefix='coreos-iso-') as tempdir:
+        proc = subprocess.Popen(['cpio', '-i', '-D', tempdir, '--quiet',
+                FILENAME], stdin=subprocess.PIPE)
```
jlebon: Optional: could probably use `--to-stdout` here and just `TemporaryFile` directly. If the OS supports `O_TMPFILE`, it avoids creating a user visible inode entirely.

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,172 @@
...
+    '''
+    with tempfile.TemporaryDirectory(prefix='coreos-iso-') as tempdir:
+        config_path = os.path.join(tempdir, FILENAME)
+        with open(config_path, 'w') as out_fh:
+            shutil.copyfileobj(ignition_fh, out_fh)
+        proc = subprocess.Popen(['cpio', '-o', '-D', tempdir, '-H', 'newc',
+                '--quiet'], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        archive, _ = proc.communicate(FILENAME.encode())
+        if proc.returncode != 0:
+            raise EmbedException(f'cpio exited with status {proc.returncode}.')
```
jlebon: Might be worth specifying `cpio -o` here and `cpio -i` in the other similar exception to differentiate them.

bgilbert:
@jlebon We could fold the commits together if you think it's important.  They were developed and tested separately, and in those cases I typically retain that structure in the commit history.

bgilbert:
Updated and folded.

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
...
+
+def do_remove(args, fh, offset, length):
+    fh.seek(offset)
+    fh.write(bytes(length))
+
+
+def main():
+    args = parse_args()
+    image = CopiedFileHolder(args.image)
+    if getattr(args, 'output', None) is not None:
```
dustymabe: could we just set the default value for the output arg to be None and just say `if args.output:` ?

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
...
+
+
+def get_location(iso_fh):
+    '''
+    Given an ISO image, return the offset and length of the Ignition
+    embed area.
+    '''
+    fmt = '<8s2Q'
+    size = struct.calcsize(fmt)
+    iso_fh.seek(32768 - size)
```
dustymabe: where does 32768 come from? We should make it a constant at the top of the script and maybe link to where it's defined

dustymabe:
reviewing now

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
...
+    buf = iso_fh.read(size)
+    if len(buf) < 24:
+        raise EmbedException('Short ISO image file.')
+    magic, offset, length = struct.unpack(fmt, buf)
+    if magic != b'coreiso+':
```
dustymabe: did we just make up `coreiso+` ? if so why not `coreosiso+` ?

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
...
+                type=argparse.FileType('r+b'),
+                help='ISO image')
+
+    return parser.parse_args()
+
+
+def do_embed(args, fh, offset, length):
+    cpio = make_cpio(args.config)
+    if len(cpio) > length:
+        raise EmbedException('Compressed Ignition config is too large: {} > {}'.format(len(cpio), length))
```
dustymabe: how large is too large? we should probably document and also tell the user here

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
...
+import sys
+import tempfile
+
+FILENAME = 'config.ign'
+
+class EmbedException(Exception):
+    pass
+
+
+class CopiedFileHolder:
```
dustymabe: what value is this class giving us?

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
...
+    return parser.parse_args()
+
+
+def do_embed(args, fh, offset, length):
+    cpio = make_cpio(args.config)
+    if len(cpio) > length:
+        raise EmbedException('Compressed Ignition config is too large: {} > {}'.format(len(cpio), length))
+    if not args.force:
+        fh.seek(offset)
+        if fh.read(length) != bytes(length):
```
dustymabe: A comment about "make sure all zeroes" would be nice.

coreos-iso-embed-ignition
dustymabe: basically garbage collection if we error out in the middle?

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/python3
+
```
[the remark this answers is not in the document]
bgilbert: It'll be in the ISO platform docs, which don't exist yet.

coreos-iso-embed-ignition
bgilbert: Yeah, magic number is magic.  My thinking was: `coreiso+` is 8 bytes, which is the same size as the other fields.

coreos-iso-embed-ignition
bgilbert: We tell the user what we can; see the last part of the message.  We don't know how well their Ignition config will compress, though.

coreos-iso-embed-ignition
bgilbert: Yup.

coreos-iso-embed-ignition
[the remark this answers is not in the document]
dustymabe: What do you think about putting a blurb here in the file for now?

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
...
+
+def main():
+    args = parse_args()
+    image = CopiedFileHolder(args.image)
+    if getattr(args, 'output', None) is not None:
+        image.copy_to(args.output)
+    offset, length = get_location(image.fh)
+    args.command(args, image.fh, offset, length)
+    image.finish()
```
dustymabe: would be nice to get a log statement when we do an embed or remove to let the user know something happened. THis is what I saw:

```
$ /var/b/shared/code/github.com/coreos/coreos-installer/coreos-iso-embed-ignition embed -c /var/b/shared/code/github.com/dustymabe/virtscripts/user-systemd.ign  ./fedora-coreos-30.20190918.dev.3-live.iso 
$ echo $?
0
```

And was like "wait, did something happen?" I then ran `show` to find out, which is nice, but would also like to get a oneliner about it successfully embedding the config.

bgilbert: Gonna push back on that one.  That's not how Unix tools behave.  :slightly_smiling_face:

dustymabe: I guess it depends on the tool. Knowing everything I know about the script it just seemed like I loaded a big gun then fired it, yet there was no sound. Was kinda eerie. :eyes:

coreos-iso-embed-ignition
bgilbert: I've commented it, but it's used in a single place and not susceptible to being changed, so I'm thinking of leaving it inline.

coreos-iso-embed-ignition
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/python3
```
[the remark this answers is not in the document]
bgilbert: Yeah, more or less.  I'll add the GPLv3 because that's the LICENSE file already in the repo, but I don't have an opinion there.

coreos-iso-embed-ignition
bgilbert: In this case, the issue is that some subcommands have an `output` argument and some don't.

bgilbert:
Updated.

coreos-iso-embed-ignition
dustymabe: +1 - I missed that point. Looks good

PR #65, opened 2019-10-25
Rewrite coreos-installer in Rust

bgilbert:
Features:

- Checks that target disk is not in use
- Obtains image from local file, remote URL, or stream metadata
- Streaming fetch and decompression
- Compression format sniffing
- Progress reporting
- GPG signature verification, mandatory unless `--insecure` specified
- Does not write partition table until signature is verified
- Can copy Ignition config to disk
- Can override Ignition platform ID
- Can write first-boot kernel arguments
- Doesn't depend on uniqueness of `/boot` UUID
- Wipes destination partition table on error
- Only executable dependencies: `gpg`, `lsblk`, `udevadm`

Command-line parameters are not compatible with the original, but the intent is that the `coreos.inst.*` kargs, to be implemented in a separate wrapper script, will maintain compatibility.


lucab:
It would be useful to turn on travis on this repo, before merging this PR. I don't seem to have the powers to do that myself though.

src/main.rs
```diff
@@ -0,0 +1,464 @@
...
+
+        // return a generic error so our exit status is right
+        bail!("install failed");
+    }
+
+    println!("Install complete.");
+    Ok(())
+}
+
+// Parse command-line arguments.
```
lucab: Here and pretty much everywhere else in the codebase: docstrings require three slashes `///`.

src/main.rs
```diff
@@ -0,0 +1,464 @@
...
+            }
+            let _ = std::io::stdout().flush();
+        }
+    });
+
+    // Wrap in a BufReader so we can peek at the first few bytes for
+    // format sniffing.  Don't trust the content-type since the server
+    // may not be configured correctly, or the file might be local.
+    // Then wrap in a reader for decompression.
+    let mut buf_reader = BufReader::with_capacity(16384, &mut progress_reader);
```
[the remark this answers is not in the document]
bgilbert: No particular reason.  I didn't want to implicitly trust that the default buffer size is large enough for the largest magic number.  I've switched to the default buffer with an assert.

src/main.rs
```diff
@@ -0,0 +1,464 @@
...
+        if sniff.len() >= 2 && &sniff[0..2] == b"\x1f\x8b" {
+            Box::new(GzDecoder::new(buf_reader))
+        } else if sniff.len() >= 6 && &sniff[0..6] == b"\xfd7zXZ\x00" {
+            Box::new(XzDecoder::new(buf_reader))
+        } else {
+            Box::new(buf_reader)
+        }
+    };
+
+    // Cache the first MB of input and write zeroes instead.  This ensures
```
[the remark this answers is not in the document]
bgilbert: The first MiB should cover all the disk headers because it's usually where the first partition starts on modern systems (x86 at least).  It's probably a larger buffer than necessary.

Credit to @imcleod for the idea!

src/main.rs
```diff
@@ -0,0 +1,464 @@
...
+            None
+        };
+        Box::new(StreamLocation::new(
+            matches.value_of("stream").unwrap(),
+            matches.value_of("architecture").unwrap(),
+            base_url.as_ref(),
+        )?)
+    };
+    // and report it to the user
+    println!("{}", location);
```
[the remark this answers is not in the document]
bgilbert: The `Display` implementations in `source.rs` produce human-readable output.  It'll be one of:

```
Copying image from /home/bgilbert/tmp/fedora-coreos-30.20191014.0-metal.x86_64.raw
Reading signature from /home/bgilbert/tmp/fedora-coreos-30.20191014.0-metal.x86_64.raw.sig
```
or:
```
Downloading image from http://localhost:8080/fedora-coreos-30.20191014.0-metal.x86_64.raw.xz
Downloading signature from http://localhost:8080/fedora-coreos-30.20191014.0-metal.x86_64.raw.xz.sig
```
or:
```
Downloading testing image and signature
```
or:
```
Downloading image and signature referenced from https://builds.coreos.fedoraproject.org/streams/testing.json
```

src/main.rs
```diff
@@ -0,0 +1,464 @@
...
+    let uname = nix::sys::utsname::uname();
+    // Args are listed in --help in the order declared here.  Please keep
+    // the entire help text to 80 columns.
+    let matches = App::new("coreos-installer")
+        .version(crate_version!())
+        .setting(AppSettings::DeriveDisplayOrder)
+        .setting(AppSettings::UnifiedHelpMessage)
+        // ways to specify the image source
+        .arg(
+            Arg::with_name("stream")
```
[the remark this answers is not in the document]
bgilbert: The current installer requires manually specifying a URL, on the command line or on the kernel command-line, and that's a really awkward UX.  I don't think users generally care, or should have to care, about the URL their OS image is coming from.  So I'd strongly prefer to keep the stream logic in coreos-installer, and also add something similar for RHCOS if there's equivalent metadata we can use.

By way of comparison, CL `coreos-install` understands channels, the `current` pointer, and `version.txt`.

src/main.rs
```diff
@@ -0,0 +1,464 @@
...
+    // Wrap in a BufReader so we can peek at the first few bytes for
+    // format sniffing.  Don't trust the content-type since the server
+    // may not be configured correctly, or the file might be local.
+    // Then wrap in a reader for decompression.
+    let mut buf_reader = BufReader::with_capacity(16384, &mut progress_reader);
+    let mut decompress_reader: Box<dyn Read> = {
+        let sniff = buf_reader.fill_buf().chain_err(|| "sniffing input")?;
+        if sniff.len() >= 2 && &sniff[0..2] == b"\x1f\x8b" {
+            Box::new(GzDecoder::new(buf_reader))
+        } else if sniff.len() >= 6 && &sniff[0..6] == b"\xfd7zXZ\x00" {
```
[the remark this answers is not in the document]
bgilbert: Yep, both of the magic numbers are endian-invariant.

src/source.rs
```diff
@@ -0,0 +1,273 @@
...
+        // load signature file if present
+        let mut signature = Vec::new();
+        let result = OpenOptions::new().read(true).open(&self.sig_path);
+        if result.is_ok() {
+            result
+                .unwrap()
+                .read_to_end(&mut signature)
+                .chain_err(|| "reading signature file")?;
+        } else {
+            eprintln!("Couldn't read signature file: {}", result.unwrap_err());
```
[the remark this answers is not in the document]
bgilbert: I was trying to avoid the extra `Option<_>` and had an explicit check for a zero-length vector, but you're right that I should do this properly.

src/blockdev.rs
```diff
@@ -0,0 +1,168 @@
...
+    if !result.status.success() {
+        // copy out its stderr
+        eprint!("{}", String::from_utf8_lossy(&*result.stderr));
+        bail!("lsblk of {} failed", device);
+    }
+    let output: LsBlk =
+        serde_json::from_slice(&*result.stdout).chain_err(|| "decoding lsblk JSON")?;
+    let mut found: Option<BlkDev> = None;
+    for dev in output.blockdevices {
+        if dev.label.is_none() || dev.label.as_ref().unwrap() != label {
```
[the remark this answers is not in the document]
bgilbert: Doesn't seem to.

src/main.rs
```diff
@@ -0,0 +1,464 @@
...
+        .arg(
+            Arg::with_name("device")
+                .help("Destination device")
+                .required(true)
+                .takes_value(true),
+        )
+        .get_matches();
+
+    // build image location
+    let location: Box<dyn ImageLocation> = if matches.is_present("image-file") {
```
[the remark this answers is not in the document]
bgilbert: Yeah.  I'd prefer to use `ArgGroup`s and `conflicts_with`, but that doesn't seem to play well with Args that have defaults.

bgilbert:
Thanks for the review!  Updated, and Travis is enabled.  Also, as we discussed OOB, I've moved the functionality to an `install` subcommand for future-proofing.

src/blockdev.rs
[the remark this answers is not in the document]
lucab: Uhm, sad.
I think this idiom will be covered by either [`contains()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.contains) or [`as_deref()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref) once those are stabilized.

src/source.rs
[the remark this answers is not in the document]
lucab: I acknowledge that the nested `Option<Vec<_>>` makes the code more annoying to write. On the upside, it makes clearer for any future reader what's going on.

src/main.rs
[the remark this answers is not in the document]
lucab: Oh yeah, I totally agree such logic should be in coreos-installer (I've just realized my previous comment was ambiguously worded, sorry).

My reasoning was focused on "how to expose this logic via CLI". Right now there are three switches that are mutually exclusive (plus additional options), and we likely have to fit RHCOS too in the future.

I was pondering alternative CLI approaches (still sharing the same logic under the hood), like:
 * dedicated verbs like `install-url`, `install-file`, `install-fcos-stream` and `install-rhcos-release`
 * composable verbs like `resolve-fcos-stream` and `resolve-rhcos-release`, producing URLs on stdout for other `install-url` and `fetch-url`

I hope it's a bit clearer now, sorry for the initial ambiguity.

lucab:
An additional wishlist item that came to my mind: it would be great if we could run the installer from a container, just with the bindmounts for the target (rationale: that would allow us to use any linux-host as a provisioner).
Did you maybe try that, or see any clear blocker? Does it sound useful to you (perhaps coupled with a quay autobuilder task) or am I getting off track?

src/signing-keys.asc
```diff
@@ -0,0 +1,91 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
jlebon: Hmm, why not just pick up the keys from `fedora-gpg-keys`/`redhat-release-coreos` (i.e. `/etc/pki/rpm-gpg`) ?

src/main.rs
```diff
@@ -0,0 +1,487 @@
...
+extern crate clap;
+#[macro_use]
+extern crate error_chain;
+extern crate flate2;
+#[macro_use]
+extern crate nix;
+extern crate progress_streams;
+extern crate reqwest;
+extern crate tempdir;
+extern crate xz2;
```
jlebon: Shouldn't need these with Rust 2018.

src/main.rs
```diff
@@ -0,0 +1,487 @@
...
+    println!("Writing Ignition config");
+
+    // make parent directory
+    let mut config_dest = mountpoint.to_path_buf();
+    config_dest.push("ignition");
+    create_dir_all(&config_dest).chain_err(|| "creating Ignition config directory")?;
+
+    // do the copy
+    config_dest.push("config.ign");
+    let mut config_in = OpenOptions::new()
```
jlebon: Why not just `std::fs::copy` here?

src/main.rs
```diff
@@ -0,0 +1,487 @@
...
+                .chain_err(|| format!("opening bootloader config {}", path.display()))?;
+            let mut orig_contents = String::new();
+            config
+                .read_to_string(&mut orig_contents)
+                .chain_err(|| format!("reading {}", path.display()))?;
+
+            // Rewrite the config.  Assume that we will only install
+            // from metal images and that their bootloader configs will
+            // always set ignition.platform.id.  Fail if those
+            // assumptions change.  This is deliberately simplistic.
```
jlebon: This is fine, though should probably no-op return early on if `platform` is `metal`.

src/signing-keys.asc
bgilbert: coreos-installer should run on arbitrary distros, which may not have those files.

jlebon: OK so the idea here is that we'll publish binaries instead of/in addition to containers? What I'm trying to avoid is adding yet another spot that needs updating during each major release rollover. Can we pull it in as part of the build process at least?

lucab: It may make sense to overlay-merge (by filename?) two sets of keys:
 * these hardcoded ones (for fallback)
 * any on-disk ones (possibly fresher)

src/main.rs
bgilbert: We get a bit better control this way:

- `config_out.create_new(true)` ensures that we'll fail if the file already exists for some reason.  (Which it shouldn't, of course.)
- `std::fs::copy` also copies permissions, which may not be desired.

src/main.rs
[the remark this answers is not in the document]
bgilbert: Ah, okay.  I agree that the current approach complicates the formal semantics, but it seems consistent with how other programs handle input/output options.  For that reason, the current approach feels more intuitive to me than having separate subcommands.  Also, to keep the UX simple, I'd rather not require piping.

There will also be a systemd unit that runs coreos-installer with parameters from kargs, and the kargs will necessarily have the same problem: there's not a great way to specify a mode.  We don't have to keep the two sets of options in sync (and in fact we can't, for compatibility reasons), but maybe it's still relevant here.

bgilbert:
> it would be great if we could run the installer from a container, just with the bindmounts for the target (rationale: that would allow us to use any linux-host as a provisioner).
> Did you maybe try that, or see any clear blocker? Does it sound useful to you (perhaps coupled with a quay autobuilder task) or am I getting off track?

I haven't tried it, but I like the idea.  I was originally thinking we'd offer a binary for download, but it does have dynamic library dependencies so that may not work well on other distros:

```
	liblzma.so.5 => /lib64/liblzma.so.5 (0x00007fb47a786000)
	libssl.so.1.1 => /lib64/libssl.so.1.1 (0x00007fb47a6f0000)
	libcrypto.so.1.1 => /lib64/libcrypto.so.1.1 (0x00007fb47a414000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007fb47a40e000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fb47a3ed000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fb47a3d3000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fb47a20b000)
	libz.so.1 => /lib64/libz.so.1 (0x00007fb47a1f1000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fb47ac29000)
```

A container image would sidestep the whole problem.  I probably won't pursue it immediately, but filed issue https://github.com/coreos/coreos-installer/issues/67.

src/signing-keys.asc
bgilbert: > OK so the idea here is that we'll publish binaries instead of/in addition to containers?

It was, but if the container works and the UX for running it isn't too bad, I'd favor shipping that instead of a binary.

> any on-disk ones (possibly fresher)

Where would we get these?  If from `/etc/pki/rpm-gpg`, there are no symbolic references for the current key, so we'd have to either import everything we find there (boo) or have a heuristic for importing keys (boo).

If we do ship a bare binary, this would also mean that behavior would differ depending on the host OS, which doesn't seem great.

> What I'm trying to avoid is adding yet another spot that needs updating during each major release rollover. Can we pull it in as part of the build process at least?

These are the roots of trust for install images, and IMO it's safer to manage them manually.  We can always start that way and add dynamic roots or more automation later on.

In comparison, CL `coreos-install` intentionally hardcodes the GPG key and does not provide a command-line mechanism for overriding it.

bgilbert:
Updated!

src/main.rs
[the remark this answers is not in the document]
lucab: Ack, I'll stick to your call here then.

lucab:
We could (in theory) cross-compile to a static binary with musl, but in practice if it works I'd prefer pursuing the container route.

src/verify.rs
```diff
@@ -0,0 +1,142 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use error_chain::bail;
+use std::fs::{metadata, set_permissions, OpenOptions};
+use std::io;
```
crawford: This can be combined below:

```rust
use std::io{self, Read, Write};
```

src/main.rs
```diff
@@ -0,0 +1,480 @@
...
+    println!("Install complete.");
+    Ok(())
+}
+
+/// Parse command-line arguments.
+fn parse_args() -> Result<Config> {
+    let uname = nix::sys::utsname::uname();
+    // Args are listed in --help in the order declared here.  Please keep
+    // the entire help text to 80 columns.
+    let app_matches = App::new("coreos-installer")
```
crawford: As an alternative to using Clap directly, StructOpt is pretty slick.

src/main.rs
```diff
@@ -0,0 +1,464 @@
...
+struct Config {
+    device: String,
+    location: Box<dyn ImageLocation>,
+    ignition: Option<String>,
+    platform: Option<String>,
+    firstboot_kargs: Option<String>,
+    insecure: bool,
+}
+
+quick_main!(run);
```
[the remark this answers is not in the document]
bgilbert: `quick_main` still does better pretty-printing than the default `main`.

src/main.rs
bgilbert: Nice!  I played with it a bit and don't think it's worth a refactor right now, but I'll keep it in mind for the future.

lucab:
I don't have any further high-level concerns, and the code itself looks good. I could just stamp the PR but I don't know what are the overarching integration plans wrt. the old one.

@bgilbert @jlebon @arithx how do you want to proceed here?

src/signing-keys.asc
jlebon: So as an example right now we don't have any CI covering the `coreos-installer` path. So bumping from f30 to f31 without bumping the keys here would've possibly gone completely unnoticed until it shipped to users. (A gap we should fix of course!)

> Where would we get these? If from /etc/pki/rpm-gpg, there are no symbolic references for the current key, so we'd have to either import everything we find there (boo) or have a heuristic for importing keys (boo).

> These are the roots of trust for install images, and IMO it's safer to manage them manually. We can always start that way and add dynamic roots or more automation later on.

A big part of this discussion is essentially a direct consequence of https://github.com/coreos/fedora-coreos-tracker/issues/187#issuecomment-540103433. WDYT about:
- ship in a container with `fedora-gpg-keys`
- parse the FCOS version to extract the major release, similar to what is proposed for RoboSignatory itself in that ticket
- verify download using *only that release's specific key* from `/etc/pki/rpm-gpg`

?

This is similar to how Fedora verifies RPMs: https://src.fedoraproject.org/rpms/fedora-repos/blob/a12c809ab2dcb730cbeb4a03c50d20e3d25cc048/f/fedora.repo#_10.

For RHEL, the keys don't rotate nearly as much, so we can just bake it in.

Anyway, I'm cool with how things set up for now, and discussing this stuff in more details later on!

jlebon:
Would be pretty cool to have the CI here actually do an install of FCOS. I think Travis CI can give you Ubuntu VMs, though we can also hook it up to CoreOS CI!

> @bgilbert @jlebon @arithx how do you want to proceed here?

Now that we've branched, I'm thinking let's just merge this, cut a 0.0.1 release, and iterate on top as we uncover issues?

src/signing-keys.asc
bgilbert: > So as an example right now we don't have any CI covering the coreos-installer path. So bumping from f30 to f31 without bumping the keys here would've possibly gone completely unnoticed until it shipped to users. (A gap we should fix of course!)

If nothing else, that'll be addressed when we start running kola on Packet, where every machine will bootstrap with coreos-installer.

> ship in a container with fedora-gpg-keys

I'd still prefer not to rely on external keys.  It'd be great if `cargo install coreos-installer` worked properly on e.g. Debian.

>  parse the FCOS version to extract the major release, similar to what is proposed for RoboSignatory itself in that ticket
> verify download using only that release's specific key from /etc/pki/rpm-gpg

I like the idea of binding to a specific key.  Will file an issue.

Hmm, maybe stream metadata should list a signing key ID for each artifact?

bgilbert: https://github.com/coreos/coreos-installer/issues/71

PR #70, opened 2019-10-29
Revert "Use `dd conv=sparse`"

cgwalters:
This reverts commit 33ba62ef154c32b6e5d8b4278e7f56180a31d8c5.

As bgilbert points out in
https://github.com/coreos/coreos-installer/pull/20#issuecomment-547561379

> This isn't safe. The image might contain zeroed blocks for reasons other
> than unallocated space, and if the target disk already has data on it,
> we'll write a corrupt image to disk.

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -682,7 +682,7 @@ write_image_to_disk() {
     fi
 
     $DECOMPRESSION_TOOL /mnt/dl/imagefile.compressed |\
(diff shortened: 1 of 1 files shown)


bgilbert:
Could you PR this to the `legacy` branch?  #65 will drop the script from master.

cgwalters:
It is done.

PR #72, opened 2019-10-30
Block device handling improvements

bgilbert:
- If the output path is not a block device, produce an error message that doesn't talk about typewriters
- Issue discard request for entire output device before writing to it, and also after a failure
- Minor cleanups


bgilbert:
BTW I've confirmed via `strace(1)` that the ioctl and parameters are the same as issued by `blkdiscard(8)`.

PR #78, opened 2019-11-06
Add download subcommand

bgilbert:
Features:
- Downloads an image from FCOS stream metadata or from a manually-specified URL
- Requires GPG signatures unless `--insecure` specified
- Avoids redownload if it can verify an existing GPG signature
- Optionally decompresses the downloaded image
- Can download arbitrary FCOS (stream, architecture, platform, format) tuple
- Automatically downloads all artifacts from multi-artifact images

Requires https://github.com/coreos/coreos-installer/pull/77.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -261,7 +261,7 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         )?)
     };
     // and report it to the user
-    println!("{}", location);
+    eprintln!("{}", location);
 
     // build configuration
     Ok(Config::Install(InstallConfig {
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -21,6 +21,7 @@ use crate::source::*;
 
 pub enum Config {
     Install(InstallConfig),
+    Download(DownloadConfig),
     IsoEmbed(IsoEmbedConfig),
     IsoShow(IsoShowConfig),
     IsoRemove(IsoRemoveConfig),
@@ -32,6 +33,13 @@ pub struct InstallConfig {
     pub ignition: Option<String>,
     pub platform: Option<String>,
     pub firstboot_kargs: Option<String>,
+    pub insecure: bool,
+}
+
+pub struct DownloadConfig {
+    pub location: Box<dyn ImageLocation>,
+    pub directory: String,
+    pub decompress: bool,
     pub insecure: bool,
 }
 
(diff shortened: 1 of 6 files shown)


bgilbert:
cc @lucab

src/download.rs
```diff
@@ -71,6 +82,47 @@ pub fn download(config: &DownloadConfig) -> Result<()> {
...
+        .open(path)
+        .chain_err(|| format!("opening {}", path.display()))?;
+
+    // perform GPG verification
+    let mut verify_reader = GpgReader::new(&mut file, signature)?;
+    let mut buf: [u8; 4096] = [0; 4096];
+    while verify_reader
+        .read(&mut buf)
+        .chain_err(|| "reading image file")?
+        > 0
```
lucab: It looks like this logic may be better as its own method attached to `GpgReader`.

src/source.rs
```diff
@@ -261,34 +261,35 @@ impl ImageLocation for StreamLocation {
...
-            &Url::parse(&artifact.signature)
-                .chain_err(|| "parsing signature URL from stream metadata")?,
-            &artifact_type,
-        )
-        .sources()
+        // build sources, letting UrlLocation handle the details
+        let mut sources: Vec<ImageSource> = Vec::new();
+        for (artifact_type, artifact) in artifacts.iter() {
+            let mut artifact_sources = UrlLocation::new_with_sig_and_type(
+                &Url::parse(&artifact.location)
```
lucab: Minor: readability here could be improved by using two variables for the URLs (removing one level of nesting).

src/download.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+/// Copy the image to disk and verify its signature.
+pub fn write_image(source: &mut ImageSource, dest: &mut File, decompress: bool) -> Result<()> {
+    // wrap source for GPG verification
+    let mut verify_reader: Box<dyn Read> = {
+        if let Some(signature) = source.signature.as_ref() {
+            Box::new(GpgReader::new(&mut source.reader, signature)?)
+        } else {
+            Box::new(&mut source.reader)
```
lucab: I have some (very minor) distress about this `&mut` and the one above, as they cause a long chain of upwards `mut` locations, only because here we want to read through the reader.
Do you think it could be feasible to avoid that by having an additional wrapping like `Cursor` or `BufReader`?

bgilbert: I'm not sure how?  `Read.read()` fundamentally expects `&mut self`.  We could get around that with a `RefCell` but that doesn't seem like the right approach.  Or am I misunderstanding your point?

lucab: I'm not totally sure (so it's fine to postpone and I'll experiment a bit, later on).
My gut feeling was that `Cursor` (or something akin) could provide the mutability for `Read` while keeping the inner type immutable.

src/source.rs
```diff
@@ -41,6 +41,7 @@ pub struct FileLocation {
 pub struct UrlLocation {
     image_url: Url,
     sig_url: Url,
+    artifact_type: String,
```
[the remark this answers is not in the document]
bgilbert: This field is only used by the progress bar to differentiate downloading multiple artifacts, e.g. the kernel and initrd.  I don't see a lot of benefit to making it an enum, and if we add additional multi-artifact images to the stream metadata, it'd be good for older coreos-installer releases to work transparently.

src/download.rs
bgilbert: I looked into it a bit.  `Cursor` only wraps a `[u8]`, and `BufReader` expects an inner `Read`.

src/cmdline.rs
```diff
@@ -146,6 +154,62 @@ pub fn parse_args() -> Result<Config> {
                         .takes_value(true),
                 ),
         )
+        .subcommand(
+            SubCommand::with_name("download")
+                .about("Download a CoreOS image")
```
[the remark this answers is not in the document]
bgilbert: I think this is the same situation as with `install`: the core of the `download` command is generic, but the data sources are not.  I was assuming we'd add additional command-line options to support RHCOS data sources.  We could have `download-fcos` and `download-rhcos`, but that seems kind of unpleasant, especially if there's eventually another downstream distro that uses FCOS- or RHCOS-compatible data sources.

bgilbert:
Updated.

src/download.rs
lucab: Ack. Then I think I'm thinking about something not directly possible. Let's keep this as it is.

src/cmdline.rs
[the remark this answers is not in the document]
lucab: Yes (and `list-stream` too). I'm just fearing that, by the time we get there, FCOS and RHCOS metadata will have diverged enough to have a different matrix of options.
If we are confident we aren't going in that direction (I'm not on top of RHCOS pipeline/metadata right now), then my concern is moot.

bgilbert:
Rebased.

lucab:
No further concerns from my side. I'll leave some space if @jlebon or others want to do a pass, otherwise I'll stamp tomorrow.

src/cmdline.rs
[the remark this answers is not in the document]
bgilbert: I'm not at all confident that we're not going in that direction.  I've been assuming that we'll "just" add more options for RHCOS.

I do expect `list-stream` to be FCOS-specific, FWIW.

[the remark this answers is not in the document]
lucab: Ack. As our crystal balls seems to be equally out of order, I think it's a fair approach to go with what we have in this PR (also because short-term UX is better).
If later we ever found out we cannot fit RHCOS in this command anymore, we'll figure out some split or a better idea. Today is not that day.

PR #79, opened 2019-11-06
Add list-stream subcommand

bgilbert:
Allow users to examine the architectures, platforms, and formats available within FCOS stream metadata.  Don't support listing artifacts within a format, since the `download` subcommand will download all of them.

Requires #78.

cc @lucab

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -22,6 +22,7 @@ use crate::source::*;
 pub enum Config {
     Install(InstallConfig),
     Download(DownloadConfig),
+    ListStream(ListStreamConfig),
     IsoEmbed(IsoEmbedConfig),
     IsoShow(IsoShowConfig),
     IsoRemove(IsoRemoveConfig),
@@ -41,6 +42,13 @@ pub struct DownloadConfig {
     pub directory: String,
     pub decompress: bool,
     pub insecure: bool,
+}
+
+pub struct ListStreamConfig {
+    pub stream_base_url: Option<Url>,
+    pub stream: String,
+    pub architecture: Option<String>,
+    pub platform: Option<String>,
 }
 
 pub struct IsoEmbedConfig {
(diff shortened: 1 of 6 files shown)


src/source.rs
```diff
@@ -276,6 +278,55 @@ impl ImageLocation for StreamLocation {
...
+                "Platforms for stream {}, architecture {}:",
+                &config.stream, architecture_name
+            );
+            result = Vec::from_iter(architecture.artifacts.keys());
+            if let Some(platform_name) = config.platform.as_ref() {
+                // the user also specified a platform
+                if let Some(platform) = architecture.artifacts.get(platform_name) {
+                    // ...and we found it
+                    header = format!(
+                        "Formats for stream {}, architecture {}, platform {}:",
```
lucab: Does "formats" mean media types here?

src/cmdline.rs
```diff
@@ -228,6 +237,45 @@ pub fn parse_args() -> Result<Config> {
                         .takes_value(true),
                 ),
         )
+        .subcommand(
+            SubCommand::with_name("list-stream")
+                .about("List available images in a Fedora CoreOS stream")
+                .usage("coreos-installer list-stream [OPTIONS] [<architecture> [<platform>]]")
```
lucab: I personally find the UX here confusing, as it seems to switch listing different entity types depending on the optional parameters.

I'd probably go for a different arrangement, where the different cases are split apart via flags or subcommands. A strawman could be:
 * `-t`: entity-type to list. Allowed values: `basearch`, `platform`, `format`. Default: all.
 * `-a`: basearch to select for filtering. Default: none (include all arch in results).
 * `-p`: platform to select for filtering. Default: none (include all platforms in results).

src/cmdline.rs
```diff
@@ -228,6 +237,45 @@ pub fn parse_args() -> Result<Config> {
...
+                        .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("platform")
+                        .value_name("platform")
+                        .help("Fedora CoreOS platform name")
+                        .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("quiet")
```
lucab: I think we want to keep this command for human usage only (as we have the original JSON for scripts), no? In that case, I don't think that we need this option.

lucab:
@bgilbert I think I have some general confusion about what we are doing here, so I'll take a couple of steps back and ask:
 * is this meant to be consumed by human interactively?
 * what do you expect is the main usecase for this? Getting the list of platforms? Getting the media types for a platform? Getting all media for all platforms?

Based on the above, I think the UI can be simplified to just list the type of entities we are interested in, with additional flags to filter/select a subset of results.

src/source.rs
bgilbert: "Formats" in the sense of [stream metadata](https://github.com/coreos/fedora-coreos-tracker/blob/master/metadata/stream/rationale.yaml).  They don't exactly map to content-types, since we have formats like `iso` and `pxe`.

bgilbert:
It's intended to help users figure out what arguments to pass to `coreos-installer download` without manually examining stream metadata.  The assumption is that users need to download one image, and downloading all images isn't a meaningful action.  I thought the subcommand might also have some value for scripts, but maybe you're right that scripts should use stream metadata directly.

I thought of the UX as descending a directory tree.  Because of the structure of stream metadata, each positional parameter only makes sense if everything to the left is also specified.  For example, if I want to list x86_64 metal artifacts, I think it makes more sense to say `coreos-installer list-stream x86_64 metal` than `coreos-installer list-stream -t format -a x86_64 -p metal`.  But that also requires users to run the command three times to get the information they want.  So maybe it'd be better to remove all the filtration arguments, and generate a multi-column table listing every format in stream metadata?

bgilbert:
Updated to drop `--quiet`.  I've also posted a tabular alternative in #84.

lucab:
Thanks for the explanation. If the goal is indeed to discover the parameters for `download` (i.e. architecture+platform+format) then I think the approach in #84 is more directly useful for users.

bgilbert:
Works for me.

PR #80, opened 2019-11-07
Add issue templates

bgilbert:
cc @lucab

PR #81, opened 2019-11-07
docs: add technical details about iso-embedding

lucab:
This adds a docs item, mostly meant for other developers, describing
what is the technical contract governing ISO embed area.

### DIFF
--- a/doc/iso-embed-ignition.md
+++ b/doc/iso-embed-ignition.md
@@ -1,0 +1,30 @@
+# ISO-embedded Ignition configuration
+
+CoreOS ISO images are typically used for live-booting machines directly from a read-only storage device (e.g. a CD-ROM or USB stick) on non-cloud platforms.
+
+First-boot provisioning via Ignition in such environments is difficult, as there are no well-defined metadata endpoints, there may not be any hypervisor back-channels or writable disks, and manual entry of an Ignition URL on the kernel command line is not ergonomic.
+
+For such reasons `coreos-installer` supports a special mode for ISO images, where an Ignition configuration file can be embedded as a user customization into a pristine image.
+The resulting image can be then used to boot a live system which is provisioned with the given Ignition configuration.
+
+## Technical details
+
+The ISO-embedding mechanism works by modifying some raw data directly on the image.
+The technical specifications are described below to help third-party logic.
+
+CoreOS ISO images come with some reserved empty space, the "embed area", which can be used to inject the Ignition configuration.
+The embed area is a block of zero-filled padding concatenated to the end of the initrd file.
+
+The location and size of the embed area depends on each image, and can be read from a header in the disk's ISO9660 System Area (i.e. the first 32KiB of the disk).
+The last 24 bytes of this 32 KiB area will contains the following fields, in this order and without special padding/alignment:
+
+ * magic value (8 bytes): the ASCII value "coreiso+" as a marker to identify CoreOS ISO images.
+ * location (8 bytes): little-endian unsigned 64-bits integer with the offset for the starting byte of the embed area.
+ * length (8 bytes): little-endian unsigned 64-bits integer with the total length of the embed area.
+
+The embed area is zeroed by default, meaning that the image is a pristine one without any user customization.
+
+User customization is performed by writing a gzip-compressed `newc` cpio archive (i.e. the equivalent of a `.cpio.gz` file) to the embed area.
+Such an archive must at least contain a regular file named `config.ign`, which can hold any custom Ignition configuration.
+
+This archive is then detected and mounted in the initrd for Ignition consumption.


lucab:
As I have this info still fresh from review, I've collected into a doc page for future quicker consumption.

/cc @bgilbert for review

docs/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
...
+
+CoreOS ISO images come with some reserved empty space, the "embed area", which can be used to inject the Ignition configuration.
+The embed area is a block of zero-filled padding concatenated to the end of the initrd file.
+
+The location and size of the embed area depends on each image, and can be read from each disk ISO9660 System Area (i.e. the first 32KiB of the disk).
+The last 24 bytes of this 32 KiB area will contains the following fields, in this order and without special padding/alignment:
+
+ * magic value (8 bytes): the ASCII value "coreiso+" as a marker to identify CoreOS ISO images.
+ * location (64 bits): little-endian unsigned 64-bits integer with the offset for the starting byte of the embed area.
+ * length (64 bits): little-endian unsigned 64-bits integer with the total lenght of the embed area.
```
miabbott: s/lenght/length/

miabbott:
More documentation is great!  :+1:

bgilbert:
Minor: https://github.com/coreos/coreos-installer/pull/80 creates `doc/` and this creates `docs/`.

lucab:
Ack, I've moved it to `doc/` then (and fixed the typo).

doc/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
+# ISO-embedded Ignition configuration
+
+CoreOS ISO images are typically used for live-booting machines directly from a read-only storage device (e.g. a CDROM) on non-cloud platforms.
```
bgilbert: s/CDROM/CD-ROM or USB stick/

doc/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
...
+
+First-boot provisioning via Ignition in such environments is difficult, as there are no reliable metadata endpoints, hypervisor back-channels, or writable disks.
```
bgilbert: Maybe `...as there are no well-defined metadata endpoints and may not be any hypervisor back-channels or writable disks, and manual entry of an Ignition URL on the kernel command line is not ergonomic.`?

doc/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
...
+
+For such reasons `coreos-installer` supports a special mode for ISO images, where an Ignition configuration file can be embedded as a user-customization into a pristine image.
```
bgilbert: `user customization`

doc/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
...
+The resulting image can be then used to boot a live system which is provisioned with the given Ignition configuration.
+
+## Technical details
+
+The ISO-embedding mechanism works by modifying some raw data directly on the image.
+The technical specifcations are described below to help third-party logic.
```
bgilbert: `specifications`

doc/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
...
+
+CoreOS ISO images come with some reserved empty space, the "embed area", which can be used to inject the Ignition configuration.
+The embed area is a block of zero-filled padding concatenated to the end of the initrd file.
+
+The location and size of the embed area depends on each image, and can be read from each disk ISO9660 System Area (i.e. the first 32KiB of the disk).
```
bgilbert: `...can be read from a header in the disk's...`

doc/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
...
+The last 24 bytes of this 32 KiB area will contains the following fields, in this order and without special padding/alignment:
+
+ * magic value (8 bytes): the ASCII value "coreiso+" as a marker to identify CoreOS ISO images.
+ * location (64 bits): little-endian unsigned 64-bits integer with the offset for the starting byte of the embed area.
```
bgilbert: It'd be more consistent to always use `(8 bytes)` rather than switching to `(64 bits)` for the integers.

doc/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
...
+ * length (64 bits): little-endian unsigned 64-bits integer with the total length of the embed area.
+
+The embed area is zero-ed by default, meaning that the image is a pristine one without any user customization.
```
bgilbert: `zeroed`

doc/iso-embed-ignition.md
```diff
@@ -0,0 +1,30 @@
...
+
+User customization is performed by writing a gzip-compressed cpio archive (i.e. the equivalent of a `.cpio.gz` file) to the embed area.
```
bgilbert: It must be a `newc` cpio archive.

lucab:
Thanks both for proof-reading! I've applied all suggestions (and realized my spell-checker stopped working at some point recently).

tuan-hoang1:
My understanding for ISO-embedding Ignition method is :
When installing a FCOS system, some users might not be able to get access to the boot console (i.e. no serial port, etc.) to add kernel parameters. Thus they wanted to prepare an ISO which is embedded with an Ignition file with all configs. When they boot, it would automatically go through coreos-installer, first boot, then second boot, which is where they would have, i.e. SSH access.

Is that the right use case ?

bgilbert:
> Is that the right use case ?

Not quite.  The feature is intended to support live systems, where the machine is booted via ISO or PXE and then runs completely from RAM.  In this model, the OS is never installed to disk at all.  In the case of PXE boots, the user can easily control the kernel arguments, so they can specify an Ignition config URL.  But with ISO boots, they'd need to type in the kernel argument by hand, which isn't practical.  So this feature allows users to embed an Ignition config directly into the ISO image.

The functionality _can_ be used to drive installation (by preparing an Ignition config that runs the installer) but it's not the primary use case.

PR #83, opened 2019-11-07
dracut/parse-coreos.sh: allow multiple networking karg values under same key

zonggen:
Only allowing single key-value correspondence would cause problems when customers
want to put more than 1 DNS server in their production instances. Previously only
the last value would be picked if multiple values under the same key are passed.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1763341

Signed-off-by: Allen Bai <abai@redhat.com>

### DIFF
--- a/dracut/30coreos-installer/parse-coreos.sh
+++ b/dracut/30coreos-installer/parse-coreos.sh
@@ -43,11 +43,14 @@ local NETWORKING_ARGS="rd.neednet=1"
 declare -a DRACUT_NET_ARGS=("ip=" "ifname=" "rd.route=" "bootdev=" "BOOTIF=" "rd.bootif=" "nameserver=" "rd.peerdns=" "biosdevname=" "vlan=" "bond=" "team=" "bridge=" "rd.net.timeout.carrier=")
 for NET_ARG in "${KERNEL_NET_ARGS[@]}" "${DRACUT_NET_ARGS[@]}"
 do
-    NET_OPT=$(getarg $NET_ARG)
+    NET_OPT=$(getargs $NET_ARG)
     if [ ! -z "$NET_OPT" ]
     then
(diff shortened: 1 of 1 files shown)


dracut/30coreos-installer/parse-coreos.sh
```diff
@@ -2,6 +2,36 @@
...
+# related bug: https://bugzilla.redhat.com/show_bug.cgi?id=1763341
+getargs_multiple() {
+    local _o _val _doecho
+    local CMDLINE="$(cat /proc/cmdline)"
+    unset _val
+    unset _o
+    unset _doecho
+
+    for _o in $CMDLINE; do
+        if [ "${_o%%=*}" = "${1%%=*}" ]; then
```
miabbott: This is some `bash` sorcery!  Could you add a comment here with a little bit more detail of what this is doing?

zonggen: For sure! Should've added comments in the first place.
Also this is mostly from [this line from `/dracut/dracut-lib.sh`](https://github.com/zfsonlinux/dracut/blob/df03141dbc9896b808cebdcc7358745dc2e8b979/modules.d/99base/dracut-lib.sh#L216) but with multiple matched karg concatenated together.

tuan-hoang1: > This is some `bash` sorcery! 
It's part of POSIX shell.
https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html.

tuan-hoang1:
I think we can use `getargs` function from dracut for this job.

zonggen:
@tuan-hoang1 That looks like the right function to use, thanks for the link!

zonggen:
Squashed to one commit

miabbott:
Nice!  The `getarg` is much easier to understand.  :+1:

PR #84, opened 2019-11-08
Add list-stream subcommand, tabular version

bgilbert:
Allow users to examine the architectures, platforms, and  formats available within FCOS stream metadata.  Don't support listing artifacts within a format, since the `download` subcommand will download all of them.

Unlike #79, this version generates a table with all arch/platform/format tuples, and doesn't provide arguments for filtering.  Sample output:

```
$ ./coreos-installer list-stream  -s testing
Architecture  Platform   Format
x86_64        aws        vmdk.xz
x86_64        metal      installer-pxe
x86_64        metal      installer.iso
x86_64        metal      iso
x86_64        metal      pxe
x86_64        metal      raw.xz
x86_64        openstack  qcow2.xz
x86_64        qemu       qcow2.xz
x86_64        vmware     ova
```

Requires #78.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -350,15 +350,8 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
             .chain_err(|| "parsing image URL")?;
         Box::new(UrlLocation::new(&image_url))
     } else {
-        let base_url = if matches.is_present("stream-base-url") {
-            Some(
-                Url::parse(
-                    matches
-                        .value_of("stream-base-url")
-                        .expect("stream-base-url missing"),
-                )
-                .chain_err(|| "parsing stream base URL")?,
-            )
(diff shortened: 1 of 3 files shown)


src/cmdline.rs
```diff
@@ -406,6 +433,28 @@ fn parse_download(matches: &ArgMatches) -> Result<Config> {
     }))
 }
 
+fn parse_list_stream(matches: &ArgMatches) -> Result<Config> {
+    let stream_base_url = if matches.is_present("stream-base-url") {
+        Some(
+            Url::parse(
+                matches
+                    .value_of("stream-base-url")
+                    .expect("stream-base-url missing"),
```
lucab: I think this can be avoided by adjusting the logic like:
```
let stream_base_url = match matches.value_of("stream-base-url") {
  Some(u) => { ... },
  None => None,
}
```

bgilbert:
Rebased and updated.

PR #88, opened 2019-11-09
Packaging and release fixes

bgilbert:


src/blockdev.rs
```diff
@@ -15,7 +15,7 @@
 use error_chain::bail;
 use nix::errno::Errno;
 use nix::{self, ioctl_none, ioctl_write_ptr_bad, mount, request_code_none};
-use serde::Deserialize;
+use serde_derive::Deserialize;
```
lucab: What is the problem you are fixing with this? It could be better to just opt-in the `derive` feature in `serde` (without further code changes):  
https://github.com/serde-rs/serde/blob/4eb580790dd6c96089b92942a5f481b21df4feaf/serde/Cargo.toml#L35-L36

bgilbert: The Fedora package, which pulls in an older serde version, won't build without it.  I can opt into `derive` if that's better.

lucab: :disappointed: 
If that's ok with the Fedora package (I didn't check), that would be my preference.

bgilbert:
Updated.

Issue #91, opened 2019-11-18
Non-partitionable target devices are not supported


r3boot:
# Bug #
I am unable to write an image to a block device, because coreos-installer fails to get an exclusive lock on the block device.

## Host Operating System Version ##
CentOS 7.7, kernel 3.10.0-1062.4.3.el7.x86_64

## Target Operating System Version ##
n/a

## coreos-installer Version ##
HEAD as of  Tue Nov 12 09:31:11 2019 -0500, commit hash: 48c4daf5ef50967a67e3f8ba78e9e788589e6130

## Expected Behavior ##
```
[16:14:08] sasami [/tmp] >> ./coreos-installer install /dev/mapper/storage-temp.as65342.net -s testing
Downloading testing image and signature
<verifying and writing a disk image>
```

## Actual Behavior ##
```
[16:14:08] sasami [/tmp] >> ./coreos-installer install /dev/mapper/storage-temp.as65342.net -s testing
Downloading testing image and signature
Error: checking for exclusive access to /dev/mapper/storage-temp.as65342.net
Caused by: rereading partition table
Caused by: EINVAL: Invalid argument
```

## Reproduction Steps ##
On centos 7:
```
yum install epel-release
yum install cargo openssl-devel
cargo build --release
./target/release/coreos-installer install /some/block/device -s testing
```

## Other Information ##
Trying to migrate my existing workflow based on CoreOS to FCOS, but running into this issue. This is with SELinux in enforcing mode, but this is not the cause for this bug.

lucab:
@bgilbert the following may be relevant for you. Quoting from https://fedoraproject.org/wiki/BlockDeviceOddities:
> The kernel's internal partitioning scheme is managed with some IOCTLs, such as BLKRRPART, BLKPG/BLKPG_ADD_PARTITION, and BLKPG/BLKPG_DEL_PARTITION, which only work on some devices. They work on hd* (ide without libata), sd* (scsi), md_d* (partitionable md), but not on md (non-partitionable md) or device-mapper devices (i.e. multipath or dmraid)

lucab:
@r3boot thanks for reporting this. I have some suspicion that this may happening because you are trying to use a devmapper device (see note above). Can you maybe try, on the same system, using a plain device?

Unrelated doubt: how do you plan to boot FCOS out of a devmapper device?

bgilbert:
[...]

PR #92, opened 2019-11-19
verify: Switch to `always` trust model

bgilbert:
Older gpg, such as 2.0.22 in CentOS 7.6, doesn't support the TOFU trust model.  Switch to `always`, which makes older hosts work at the cost of a worse spurious warning.  Old verification text:

```
gpg: Signature made Mon 14 Oct 2019 11:59:16 PM EDT
gpg:                using RSA key EF3C111FCFC659B9
gpg: Good signature from "Fedora (30) <fedora-30-primary@fedoraproject.org>" [full]
gpg: fedora-30-primary@fedoraproject.org: Verified 1 signatures in the past
     0 seconds.  Encrypted 0 messages.
```

New verification text:

```
gpg: Signature made Mon 14 Oct 2019 11:59:16 PM EDT
gpg:                using RSA key EF3C111FCFC659B9
gpg: Good signature from "Fedora (30) <fedora-30-primary@fedoraproject.org>" [unknown]
gpg: WARNING: Using untrusted key!
```

The `WARNING` line is intentional on the part of gpg (to note that the stored trust is being overridden) and there doesn't seem to be a good way to omit it.  We could switch to `gpgv`, but that doesn't check for key expiration, which seems a useful thing to warn about.

### DIFF
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -85,11 +85,8 @@ impl<R: Read> GpgReader<R> {
             .arg("--homedir")
             .arg(gpgdir.path())
             .arg("--batch")
-            // avoid warnings about untrusted keys
             .arg("--trust-model")
-            .arg("tofu")
-            .arg("--tofu-default-policy")
-            .arg("good")
+            .arg("always")
             .arg("--verify")
             .arg(&signature_path)
             .arg("-")


lucab:
> The `WARNING` line is intentional on the part of gpg (to note that the stored trust is being overridden)

To the best of my knowledge, I think the warning is showing up because the imported key is not marked as trusted.
If that's the case, we have several ways to silence that:
 * use `--trusted-keys $ID` when verifying
 * call an additional `gpg --import-ownertrust` before verifying

I think we can safely merge this to unblock #91, and circle back later to improve the UX wrt. this warning.

r3boot:
I can confirm that #92 fixes gpg on centos 7. Another error did pop up, but I will open a separate issue for that.
```
[root@neutralino r3boot]# ./coreos-installer install /dev/vdb -s testing -i ./temp.json 
Downloading testing image and signature
gpg: Signature made Tue 15 Oct 2019 05:59:16 AM CEST using RSA key ID CFC659B9
gpg: Good signature from "Fedora (30) <fedora-30-primary@fedoraproject.org>"
gpg: WARNING: Using untrusted key!
> Read disk 416.2 MiB/416.2 MiB (100%)   
lsblk: unrecognized option '--json'

Usage:
 lsblk [options] [<device> ...]

Options:
 -a, --all            print all devices
 -b, --bytes          print SIZE in bytes rather than in human readable format
 -d, --nodeps         don't print slaves or holders
 -D, --discard        print discard capabilities
 -e, --exclude <list> exclude devices by major number (default: RAM disks)
 -I, --include <list> show only devices with specified major numbers
 -f, --fs             output info about filesystems
 -h, --help           usage information (this)
 -i, --ascii          use ascii characters only
 -m, --perms          output info about permissions
 -l, --list           use list format output
 -n, --noheadings     don't print headings
 -o, --output <list>  output columns
 -p, --paths          print complate device path
 -P, --pairs          use key="value" output format
 -r, --raw            use raw output format
 -s, --inverse        inverse dependencies
 -t, --topology       output info about topology
 -S, --scsi           output info about SCSI devices

 -h, --help     display this help and exit
 -V, --version  output version information and exit

Available columns (for --output):
        NAME  device name
       KNAME  internal kernel device name
     MAJ:MIN  major:minor device number
      FSTYPE  filesystem type
  MOUNTPOINT  where the device is mounted
       LABEL  filesystem LABEL
        UUID  filesystem UUID
   PARTLABEL  partition LABEL
    PARTUUID  partition UUID
          RA  read-ahead of the device
          RO  read-only device
          RM  removable device
       MODEL  device identifier
      SERIAL  disk serial number
        SIZE  size of the device
       STATE  state of the device
       OWNER  user name
       GROUP  group name
        MODE  device node permissions
   ALIGNMENT  alignment offset
      MIN-IO  minimum I/O size
      OPT-IO  optimal I/O size
     PHY-SEC  physical sector size
     LOG-SEC  logical sector size
        ROTA  rotational device
       SCHED  I/O scheduler name
     RQ-SIZE  request queue size
        TYPE  device type
    DISC-ALN  discard alignment offset
   DISC-GRAN  discard granularity
    DISC-MAX  discard max bytes
   DISC-ZERO  discard zeroes data
       WSAME  write same max bytes
         WWN  unique storage identifier
        RAND  adds randomness
      PKNAME  internal parent kernel device name
        HCTL  Host:Channel:Target:Lun for SCSI
        TRAN  device transport type
         REV  device revision
      VENDOR  device vendor

For more details see lsblk(8).
Error: lsblk of /dev/vdb failed
Clearing partition table
Error: install failed
```

bgilbert:
@r3boot Ah, great, thanks for reporting the `lsblk` problem.

bgilbert:
https://github.com/coreos/coreos-installer/pull/94 should fix this properly.

bgilbert:
#95 should fix the `lsblk` problem.

PR #94, opened 2019-11-19
verify: Explicitly trust imported keys

bgilbert:
After importing GPG keys, extract their key IDs and explicitly mark them trusted.  Avoids warning message on verify:

    gpg: WARNING: Using untrusted key!


bgilbert:
Updated to copy the `--check-trustdb` stderr to our stderr if the command fails.

bgilbert:
Merging per @lucab.

PR #95, opened 2019-11-20
Fix mounting boot device on CentOS 7.6

bgilbert:
The CentOS 7.6 `lsblk` doesn't support `--json` or the PATH key.  Switch to `--pairs`, and form the device path from `NAME`.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -15,7 +15,8 @@
 use error_chain::bail;
 use nix::errno::Errno;
 use nix::{self, ioctl_none, ioctl_write_ptr_bad, mount, request_code_none};
-use serde::Deserialize;
+use regex::Regex;
+use std::collections::HashMap;
 use std::fs::{remove_dir, File};
 use std::io::{Seek, SeekFrom};
 use std::os::unix::io::AsRawFd;
@@ -36,23 +37,17 @@ pub fn mount_boot(device: &str) -> Result<Mount> {
     }
 }
 
-#[derive(Deserialize)]
-struct LsBlk {
-    blockdevices: Vec<BlkDev>,
-}
-
-#[derive(Deserialize)]
 struct BlkDev {
     path: String,
-    label: Option<String>,
     fstype: Option<String>,
 }
 
 fn get_partition_with_label(device: &str, label: &str) -> Result<Option<BlkDev>> {
+    // have lsblk enumerate partitions on the device
     let result = Command::new("lsblk")
-        .arg("--json")
+        .arg("--pairs")
         .arg("--output")
-        .arg("PATH,LABEL,FSTYPE")
+        .arg("NAME,LABEL,FSTYPE")
         .arg(device)
         .output()
         .chain_err(|| "running lsblk")?;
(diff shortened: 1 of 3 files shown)


src/blockdev.rs
```diff
@@ -61,17 +56,48 @@ fn get_partition_with_label(device: &str, label: &str) -> Result<Option<BlkDev>>
...
+        let mut path = Path::new(device)
+            .parent()
+            .chain_err(|| format!("couldn't get parent directory of {}", device))?
+            .to_path_buf();
+        match fields.get("NAME") {
+            None => continue,
+            Some(name) => path.push(name),
+        }
+
+        // accept
```
lucab: Spurious line?

src/blockdev.rs
```diff
@@ -61,17 +56,48 @@ fn get_partition_with_label(device: &str, label: &str) -> Result<Option<BlkDev>>
         eprint!("{}", String::from_utf8_lossy(&*result.stderr));
         bail!("lsblk of {} failed", device);
     }
-    let output: LsBlk =
-        serde_json::from_slice(&*result.stdout).chain_err(|| "decoding lsblk JSON")?;
+    let output = String::from_utf8(result.stdout).chain_err(|| "decoding lsblk output")?;
+
+    // walk each device in the output
+    let re = Regex::new("([A-Z-]+)=\"([^\"]+)\"").unwrap();
```
lucab: You can use raw strings here if you prefer avoiding quote-escapes: `r#"a "raw" string"#`

src/blockdev.rs
```diff
@@ -61,17 +56,48 @@ fn get_partition_with_label(device: &str, label: &str) -> Result<Option<BlkDev>>
...
     let mut found: Option<BlkDev> = None;
-    for dev in output.blockdevices {
-        if dev.label.is_none() || dev.label.as_ref().unwrap() != label {
-            continue;
+    for line in output.lines() {
+        // parse key-value pairs
+        // newer versions of lsblk support JSON but CentOS 7 doesn't
+        let mut fields: HashMap<String, String> = HashMap::new();
+        for cap in re.captures_iter(line) {
+            fields.insert(cap[1].to_string(), cap[2].to_string());
```
lucab: Should we check-and-continue for `cap.len() < 2` here?

src/blockdev.rs
```diff
@@ -61,17 +56,48 @@ fn get_partition_with_label(device: &str, label: &str) -> Result<Option<BlkDev>>
...
+    // walk each device in the output
+    let re = Regex::new("([A-Z-]+)=\"([^\"]+)\"").unwrap();
     let mut found: Option<BlkDev> = None;
-    for dev in output.blockdevices {
-        if dev.label.is_none() || dev.label.as_ref().unwrap() != label {
-            continue;
+    for line in output.lines() {
+        // parse key-value pairs
+        // newer versions of lsblk support JSON but CentOS 7 doesn't
+        let mut fields: HashMap<String, String> = HashMap::new();
```
lucab: If you have time for it here, I would move the output parsing logic to its own function, returning this hashmap.
That way we can have a test covering it with some real-world `lsblk` output, in turn making it easier to see what's going on.

src/blockdev.rs
bgilbert: Uh, no?

src/blockdev.rs
bgilbert: Neither of the match groups in the regex are optional, so we shouldn't end up here if `cap.len() < 3`.  I'm okay with the panic in that case.

src/blockdev.rs
lucab: Does it spark joy?

On a serious note: :+1:, sorry, it was hanging a bit alone so my brain put it in the garbage pile.

bgilbert:
Updated.

PR #100, opened 2019-12-01
systemd: add scripts and systemd units for running at boot

bgilbert:
Should be compatible with the legacy kargs syntax, with the following adjustments:

- Drop `coreos.inst=yes`, since it's redundant.  `coreos.inst.install_dev` is mandatory, so run only if that karg is specified.
- Expect full device paths in `coreos.inst.install_dev`.  For compatibility, prepend `/dev/` if a bare device name is specified.
- Make `coreos.inst.ignition_url` optional.  For compatibility, ignore the karg if `skip` is specified as the value.
- Add kargs for Fedora CoreOS stream, stream base URL, and to enable insecure mode.

Untested.  Marking draft for now.


dustymabe:
>     * Drop `coreos.inst=yes`, since it's redundant.  `coreos.inst.install_dev` is mandatory, so run only if that karg is specified.

In the past specifying just `coreos.inst=yes` would get us into the initramfs and then print out the usage and drop to a prompt so the user could run the installer interactively.

I'm just trying to make sure that's no longer desirable. Forgive me if this point is moot now with the rewrite - I haven't had time to digest the rewrite yet.

dustymabe:
> Forgive me if this point is moot now with the rewrite - I haven't had time to digest the rewrite yet.

yeah I think it's moot ignore

dustymabe:
Thanks @bgilbert. This mostly looks good to me, but there were quite a few things that needed tweaking. I finally got it all working and tested. I opened up #119 as a followup PR that builds on top of this one. 

I also opened #117, #118, and #120 for us to discuss regarding the behavior of the installer.

jlebon:
This was merged as part of #119.

PR #105, opened 2019-12-03
s390x: use common SCSI code path for zFCP SCSI devices

tuan-hoang1:
While working on the multipath support, zFCP SCSI experts say that
zFCP-attached SCSI devices on s390x are just the normal SCSI devices on
x86_64. Any special handling for SCSI on s390x is not needed and maybe
more error prone. The reason I thought `dd` the whole SCSI image did not
work for zFCP SCSI is because I messed up some `dd` parameters somehow.

Now tested again on both FCOS/RHCOS and it works fine with common code
path.

With this commit, we only need special code path for ECKD DASD devices.

Suggested-by: Steffen Maier <maier@linux.ibm.com>
https://github.com/steffen-maier

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -724,6 +724,7 @@ write_image_to_dasd() {
     # low-level format the ECKD DASD using dasdfmt, if needed
     DEST_DEV_BUS=$(lszdev --by-node ${DEST_DEV} | tail -n 1  | awk '{print $2}')
     if [ "$(< /sys/bus/ccw/devices/${DEST_DEV_BUS}/status)" = "unformatted" ]; then
+        log "Formatting DASD device ${DEST_DEV_BUS}"
         dasdfmt --blocksize 4096 --disk_layout cdl --mode full -ypv "${DEST_DEV}"
     fi
     block_per_tracks=$(fdasd -p ${DEST_DEV} | grep blocks\ per\ track | awk '{print $5}')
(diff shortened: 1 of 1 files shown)


tuan-hoang1:
cc zFCP SCSI master https://github.com/steffen-maier
cc @jlebon per our discussion

coreos-installer
```diff
@@ -724,6 +724,7 @@ write_image_to_dasd() {
     # low-level format the ECKD DASD using dasdfmt, if needed
     DEST_DEV_BUS=$(lszdev --by-node ${DEST_DEV} | tail -n 1  | awk '{print $2}')
     if [ "$(< /sys/bus/ccw/devices/${DEST_DEV_BUS}/status)" = "unformatted" ]; then
+        log "Formatting DASD device ${DEST_DEV_BUS}"
```
steffen-maier: minor nit: seems topic-wise a bit unrelated, but I'm OK with this additional changed line in the DASD special handling

coreos-installer
```diff
@@ -764,52 +765,16 @@ EOF
...
 #And Write the image to disk
 #########################################################
 write_image_to_disk() {
     log "Writing disk image"
 
     set -o pipefail
-    if [[ "${arch}" =~ (aarch64|ppc64le|x86_64) ]]
+    # only s390x DASD devices require a special code path
+    if [[ "${arch}" =~ (aarch64|ppc64le|x86_64) || \
+         ! ( "${arch}" = s390x && "${DEST_DEV}" =~ ^.*dasd.*$ ) ]]
```
steffen-maier: I think, any kind of zfcp-attached SCSI disk (single or multipath) should never have "dasd" as substring in its block device path string, so the check should be OK.

coreos-installer
tuan-hoang1: Yep :D hijacking this PR a bit for this line.

coreos-installer
jlebon: Hmm, is this conditional correct? Seems like it'll be true for *any* basearch which isn't s390x, not just the specific whitelisted ones. I think we want something like:

```
    if [[ "${arch}" =~ (aarch64|ppc64le|x86_64) || \
        ( "${arch}" == s390x && [[ ! "${DEST_DEV}" =~ ^.*dasd.*$ ]] ) ]]
```

right?

jlebon: Is there a more solid way using e.g. `lsblk` to check whether it's a DASD device?

PR #107, opened 2019-12-04
[legacy] Wait for device, drop dependency on initrd-root-device

cgwalters:
Haven't tested this yet.

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -487,14 +487,14 @@ get_img_url() {
 }
 
 ############################################################
-# Validate that selected device exists
+# Wait for selected device to appear
 ############################################################
 check_selected_device() {
-    if [ ! -b $DEST_DEV ]
-    then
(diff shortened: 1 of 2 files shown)


coreos-installer
```diff
@@ -487,14 +487,14 @@ get_img_url() {
...
-    if [ ! -b $DEST_DEV ]
-    then
-        log "${DEST_DEV} does not exist."
-        exit 1
-    fi
+
+    while [ ! -b $DEST_DEV ]; do
+        log "Awaiting ${DEST_DEV}"
+        sleep 3
+    done
```
jlebon: Hmm, should we put a time limit on this, e.g. 5m or something?

Another more systemd-y approach is dynamically adding a `Requires` & `After` on `$(systemd-escape $DESTDEV).device` in the generator.

bgilbert:

PR #110, opened 2019-12-12
Change Root partition index to 4 by default for DASD installs

Prashanth684:
Moving forward, both FCOS and RHCOS use the same partitioning scheme
Only 4.2 RHCOS uses anaconda where the root parition index is 2. So,
making that an exception

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -730,11 +730,11 @@ write_image_to_dasd() {
 
     # the first 2 tracks of the ECKD DASD are reserved
     first_track=2
-    # in RHCOS4.2 using anaconda, we have root partition at index 2 while in FCOS it's 4
+    # in RHCOS 4.2 using anaconda, root partition index is 2
     BOOTPN=1
-    ROOTPN=2
-    if [[ -n $(grep -i fedora /usr/lib/initrd-release) ]]; then
-        ROOTPN=4
+    ROOTPN=4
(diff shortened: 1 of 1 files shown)


Prashanth684:
cc tuan-hoang1

coreos-installer
```diff
@@ -730,11 +730,11 @@ write_image_to_dasd() {
...
     # the first 2 tracks of the ECKD DASD are reserved
     first_track=2
-    # in RHCOS4.2 using anaconda, we have root partition at index 2 while in FCOS it's 4
+    # in RHCOS 4.2 using anaconda, root partition index is 2
     BOOTPN=1
-    ROOTPN=2
-    if [[ -n $(grep -i fedora /usr/lib/initrd-release) ]]; then
-        ROOTPN=4
+    ROOTPN=4
+    if [[ -n $(grep -i 'Red Hat Enterprise Linux CoreOS 42' /usr/lib/initrd-release) ]]; then
```
Prashanth684: not sure if we need this. i believe we are not going to make changes to 4.2 for s390x anymore. @tuan-hoang1 - thoughts?

tuan-hoang1: Even if we have change for 4.2.z, this should still be fine. Also please test for both 4.2 and 4.3.

Prashanth684: yes, tested on 4.2 and 4.3

Issue #118, opened 2019-12-21
allow for `--ignition` to grab from a url; check before downloading disk


dustymabe:
Can we allow for ignition to be either a local file or a URL that can be retrieved? Right now it only accepts a local file. Would be nice if we can use either a local file or grab from a URL if a URL is specified.

Also can we verify we can grab the ignition config (from either local file or URL) before we try to download the disk image? I hit a case where the installer downloaded the disk image and then failed to locate the ignition config. It would have saved some bandwidth if we checked the ignition file first.

```
Downloading image from http://192.168.122.1:9000/fedora-coreos-31.20191220.dev.3-metal.x86_64.raw.gz
Downloading signature from http://192.168.122.1:9000/fedora-coreos-31.20191220.dev.3-metal.x86_64.raw.gz.sig
Signature fetch failed: 404 Not Found
Signature not found; skipping verification as requested
> Read disk 684.9 MiB/684.9 MiB (100%)   
Writing Ignition config
Error: opening source Ignition config http://192.168.122.1:8000/user-systemd.ign
Caused by: No such file or directory (os error 2)
Clearing partition table
Error: install failed
```

dustymabe:
@jlebon, so you hit this too. Do you support the feature request?

lucab:
My 2¢ on this: the second part of the ticket sounds fair, but I'm a bit wary on the first part (grab content from URL).

In particular, I think we are heading toward the usual spectrum of nuances around remote content (checking integrity, checking authenticity, endpoints behind authentication, cloud buckets, etc).
As we are now running in a real OS and those details change across usecases, I'd rather leave that part to the users so that they can range from a simple `curl` to more complex `podman run $downloader` stuff.

jlebon:
> In particular, I think we are heading toward the usual spectrum of nuances around remote content (checking integrity, checking authenticity, endpoints behind authentication, cloud buckets, etc).

Hmm, I think at least supporting HTTP seems reasonable given that we already today support downloading disk images that way too. (In fact, when I ran into this, I was serving the image and the Ignition config via the same httpd thinking it would Just Work.)

I totally agree though we don't want to start expanding this much farther (just like we wouldn't for the disk images themselves). Plus, users can also leverage the Ignition fetcher too to download from e.g. S3 and tftp.

jlebon:
https://github.com/coreos/coreos-installer/pull/126

bgilbert:
[...]

PR #119, opened 2019-12-21
systemd: add scripts and systemd units for running at boot

dustymabe:
This builds on #100. It fixes up the units with changes from testing and adds a `coreos-installer-noreboot.service` that is used when the user requests no reboot.


systemd/coreos-installer-generator
```diff
@@ -17,7 +17,20 @@ karg() {
...
+    esac
+}
+
 if [ -n "$(karg coreos.inst.install_dev)" ]; then
     ln -sf "/usr/lib/systemd/system/coreos-installer.target" \
         "${UNIT_DIR}/default.target"
+
+    # Create precondition for coreos-installer-reboot.service if requested
+    if ! karg_bool coreos.inst.skip_reboot; then
+        touch /run/coreos-installer-reboot
```
jlebon: Hmm, I'm confused why we need the reboot to happen in a separate service entirely. Can't we just do it from the installer directly after a successful install? E.g. we add a `--reboot` flag and have `coreos-installer-service` just pass that in if the karg is specified.

systemd/coreos-installer-reboot.service
```diff
@@ -2,7 +2,7 @@
 Description=Reboot after CoreOS Installer
 After=coreos-installer.service
 OnFailure=emergency.target
-OnFailureJobMode=isolate
+OnFailureJobMode=replace-irreversibly
```
jlebon: Huh weird, `replace-irreversibly` didn't work for me in the past (hence why the ignition-dracut services use `isolate`: https://github.com/coreos/ignition-dracut/blob/c4790bc283c8cc5318019ef8d5ad3929c8259d0c/dracut/30ignition/ignition-complete.target#L12). Though maybe it's time to retest this again and clean things up if it works again.

systemd/coreos-installer-noreboot.service
```diff
@@ -0,0 +1,13 @@
...
+After=coreos-installer.service
+ConditionPathExists=!/run/coreos-installer-reboot
+
+[Service]
+Type=oneshot
+StandardInput=tty-force
+StandardOutput=inherit
+StandardError=inherit
+ExecStartPre=/usr/bin/echo -e "\nCoreOS install complete. Starting login shell\n"
+ExecStart=/usr/sbin/sulogin --force
```
jlebon: Ahh, interesting. I think this overlaps with my previous question re. the reboot service.

So I guess the question is: do we want this boot to be special or just like any other boot (but with code that runs the installer in the background). IOW, if users want to use a shell after installation, they could just provision the system as usual via Ignition and login, right?

I guess, I'm trying to wrap my head around whether we want to do this, when IIUC part of the idea was to move towards `coreos-installer` just be a regular binary you can run from a normal FCOS/RHCOS boot (or script to run via Ignition), and leave the kargs approach as a "legacy" thing.

systemd/coreos-installer-reboot.service
```diff
@@ -8,3 +8,5 @@ ConditionPathExists=/run/coreos-installer-reboot
 [Service]
 Type=simple
 ExecStart=/usr/bin/systemctl --no-block reboot
+StandardOutput=kmsg+console
+StandardError=kmsg+console
```
jlebon: Is this just to be consistent, or did you actually encounter an issue with the service?

systemd/coreos-installer-noreboot.service
jlebon: > and leave the kargs approach as a "legacy" thing.

Though I notice Benjamin's patch adds more kargs to the list for e.g. FCOS streams, so maybe I'm mistaken here.

And OK yeah, I totally missed that we're still overriding `default.target` here. So this WFM! (I thought we were just running this in the background of an otherwise normal boot.)

jlebon: OK, I've split out some thoughts on this in https://github.com/coreos/coreos-installer/issues/124. But anyway, it doesn't have to block on this, though... can we not add new kargs for the time being?

systemd/coreos-installer-generator
dustymabe: I think @bgilbert was trying to keep the utility separate from the service, but I'm not opposed to adding a --reboot flag to the utility. Though maybe we can open a separate issue and do that in a followup so we don't block?

systemd/coreos-installer-reboot.service
dustymabe: I think there was a case where something wasn't working and I had to add this to find out why

dustymabe: ahh yes. I think it was fab2fcc

systemd/coreos-installer-noreboot.service
dustymabe: >  can we not add new kargs for the time being?

it looks like the ones he added were `coreos.inst.insecure` `coreos.inst.stream` `coreos.inst.stream_base_url`. Those were mostly convenience based on the new functionality within the rust installer. I'd prefer not to rip them out, but also want to get this merged ASAP, so I'll go either way.

dustymabe: Note that the tool defaults to the stable stream. If we don't provide any way to change the stream then anyone trying the installer today (when no stable stream exists) and not providing their own disk image url will experience a failure.

dustymabe: and I think we need coreos.inst.insecure too because otherwise you can't even locally test the kargs approach right now without it.

jlebon: Hmm, at least `coreos.inst.stream_base_url`, it seems like we can drop, right? If you're targeting your own image, you can just use `coreos.inst.image_url` for now.

After that, yeah let me finish testing and then get this in!

dustymabe: yes, that one options seems to be more useful for us testing out the other parts of the metadata we publish. If you'd like I can try to drop that option.

systemd/coreos-installer-reboot.service
dustymabe: Let me know what you find

jlebon:
OK, tested this locally both in the interactive mode and via kargs. :heavy_check_mark: (I fell right in https://github.com/coreos/coreos-installer/issues/118 too. Will address that one as a follow-up.)

Anyway, mind addressing https://github.com/coreos/coreos-installer/pull/119#discussion_r363794237 before we merge this? I really don't want to add more kargs than necessary until we sort out #124.

dustymabe:
Rebased on top of latest master. Will merge shortly

dustymabe:
> Dusty confirmed he'll address it in a follow-up.

this was addressed in #125

bgilbert:
From https://github.com/coreos/coreos-installer/pull/119/commits/cb6dc6148c6c627cd1e98d40414db57383470ff5:
> We can't wait until the service runs to create the flag file (`/run/coreos-installer-reboot`) because systemd unit conditions are evaluated way before the coreos-installer.service runs.

They're evaluated at the moment a service is about to be started.  Did the original code not work for you in testing?

bgilbert:
This looks great.  Thanks for getting it to the finish line!

dustymabe:
> From [cb6dc61](https://github.com/coreos/coreos-installer/commit/cb6dc6148c6c627cd1e98d40414db57383470ff5):
> 
> > We can't wait until the service runs to create the flag file (`/run/coreos-installer-reboot`) because systemd unit conditions are evaluated way before the coreos-installer.service runs.
> 
> They're evaluated at the moment a service is about to be started. Did the original code not work for you in testing?

Nope. It didn't work for me but now that I look at the commits I suspect that the very next commit (87c8c31) is related to why it wasn't working for me. 

> systemd: service: Make reboot service run after installer
>
> We make the reboot service as After=coreos-installer.service and we
> mark coreos-installer.serivce as `oneshot` that will make systemd not
> start any other units until the service has completed.

dustymabe:
@bgilbert Care to post a follow-up to fix that?

PR #123, opened 2020-01-06
Add Makefile

jlebon:
A bit unfortunate to wrap `cargo build` like this, though AFAIK `cargo`
doesn't have support for installing other data files today. (And plus,
we're making `make && make install DESTDIR=/path/to/cosa/workdir/rootfs`
the standard API for our projects.)

Works on top of https://github.com/coreos/coreos-installer/pull/119.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,26 @@
+RELEASE ?= 0
+
+ifeq ($(RELEASE),1)
+	PROFILE ?= release
+	CARGO_ARGS = --release
+else
+	PROFILE ?= debug
+	CARGO_ARGS =
+endif
+
+.PHONY: all
+all:
+	cargo build ${CARGO_ARGS}
+
+.PHONY: install
+install: install-bin install-systemd
+
+.PHONY: install-bin
+install-bin: all
+	install -D -t ${DESTDIR}/usr/bin target/${PROFILE}/coreos-installer
+
+.PHONY: install-systemd
+install-systemd: all
+	install -D -m 644 -t $(DESTDIR)/usr/lib/systemd/system systemd/*.{service,target}
+	install -D -t $(DESTDIR)/usr/lib/systemd/system-generators systemd/coreos-installer-generator
+	install -D -t $(DESTDIR)/usr/libexec systemd/coreos-installer-service
--- a/README.md
+++ b/README.md
@@ -38,9 +38,26 @@ usual way:
 sudo dnf install coreos-installer
 ```
 
+# Build and install from source tree
+
+To build from the source tree:
+
+```sh
+make
+```
+
+To install the binary and systemd units to a target rootfs
+(e.g. under a
+[coreos-assembler](https://github.com/coreos/coreos-assembler)
+workdir):
+
+```sh
+make install DESTDIR=/my/dest/dir
+```
+
 # Install with Cargo
 
-You can also install coreos-installer with Rust's Cargo package manager:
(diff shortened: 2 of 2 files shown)


dustymabe:
should we update the README to mention the makefile?

dustymabe:
> (And plus,
> we're making `make && make install DESTDIR=/path/to/cosa/workdir/rootfs`
> the standard API for our projects.)

+1. Though I'd prefer if we put protections in place like requiring DESTDIR:

```
ifndef DESTDIR                                                                                                                                                                               
$(error DESTDIR is not set)                                                                                                                                                                  
endif
```

i.e. if someone is going to modify their running system then make them be explicit by passing `DESTDIR=/`

lucab:
Indeed, `cargo` does not support installing additional files. I'm +1 on this with the further `DESTDIR` suggestion from @dustymabe.

The only issue I see here is that for most people "install" means "give me the binary", while this sets up a bunch system services and kargs compat.
How will the final specfile be arranged? Do we keep the split between binary and other stuff? Can we split the install steps too?

dustymabe:
> How will the final specfile be arranged? Do we keep the split between binary and other stuff? Can we split the install steps too?

The current [PR](https://src.fedoraproject.org/rpms/rust-coreos-installer/pull-request/1) for the spec file just delivers the files directly, but I do like splitting the service out to a separate sub package so I'll update the PR to create a `coreos-installer-systemd` subpackage.

jlebon:
OK, updated this to split out `install` into separate `install-bin` and `install-systemd` targets add added instructions to the README!

> +1. Though I'd prefer if we put protections in place like requiring DESTDIR:

Hmm, do we still need this now that the `install` target is split out? I can do this, though requiring `DESTDIR` I don't think is very common. I've also added a message about what it does in the `README` blurb.

dustymabe:
> > +1. Though I'd prefer if we put protections in place like requiring DESTDIR:
> 
> Hmm, do we still need this now that the `install` target is split out? I can do this, though requiring `DESTDIR` I don't think is very common. I've also added a message about what it does in the `README` blurb.

I'd prefer it, but optional. It's way more common to target some sort of chroot directory set up rather than `/`. If someone didn't intend `/` then we just saved them some headache. If someone really wants `/` then they can specify it.

dustymabe:
#119 merged, please rebase

jlebon:
Done!

dustymabe:
One thing I notice is that I don't think I can use the makefile in the rpm. The rpm already has `%cargo_build`, which expands to:

```
/usr/bin/env CARGO_HOME=.cargo RUSTC_BOOTSTRAP=1 /usr/bin/cargo build -j4 -Z avoid-dev-deps --release
```
which I think is a good idea to re-use to follow what fedora does. Since `make install` includes the `all` target we would be compiling twice (unnecessary).

jlebon:
> One thing I notice is that I don't think I can use the makefile in the rpm. The rpm already has `%cargo_build`, which expands to:

Ahh right, OK. Hmm, I think we can still use `make install-systemd` at least, right?

dustymabe:
> Ahh right, OK. Hmm, I think we can still use `make install-systemd` at least, right?

I don't think so. Same problem:

https://github.com/coreos/coreos-installer/blob/e76635c41fda3bfcc29f23fed20172657dd37d5c/Makefile#L22-L23

dustymabe:
Unless there is some way to opt out of specified targets.

Issue #117, opened 2019-12-21
combine image options - apply sanity checking


dustymabe:
Right now we have: 

```
    -u, --image-url <URL>          Manually specify the image URL                   
    -f, --image-file <path>        Manually specify a local image file
```

It seems like it would be reasonable to use a single option and support either a URL or a local path to a file being passed.

dustymabe:
@lucab @jlebon @bgilbert - WDYT? I'm thinking we just have: 

```
-f, --image-file <path|URL> Manually specify a local image file path or remote image file URL
```

jlebon:
Seems reasonable to me.

lucab:
Not against this on principle, but I'm not sure that `--image-file https://...` is less confusing.

What's the problem we are having with the current CLI? Is it a UX hurdle or just a cosmetic for `--help`?

dustymabe:
> Not against this on principle, but I'm not sure that `--image-file https://...` is less confusing.

Maybe `--image <path|URL>` would be better, but ignition is currently squatting on the `-i` short option. 

> 
> What's the problem we are having with the current CLI? Is it a UX hurdle or just a cosmetic for `--help`?

I think both. You'd never specify both `--image-file` and `--image-url` to a single invocation and they both specify the image, just different ways of retrieving it, which is encoded in the value passed to the option.

bgilbert:
`https://example.com/config.ign` is a valid filename.  `./https://example.com/config.ign` is a valid workaround, but that creates obligations for the calling code.

More generally: if we combined the two options, there are conceivable use cases where calling code might need to do extra validation of user-provided filenames and URLs, and that validation would need to match coreos-installer's parser in all cases.  That doesn't seem worth it to me.

dustymabe:
> `https://example.com/config.ign` is a valid filename. `./https://example.com/config.ign` is a valid workaround, but that creates obligations for the calling code.
> 
> More generally: if we combined the two options, there are conceivable use cases where calling code might need to do extra validation of user-provided filenames and URLs, and that validation would need to match coreos-installer's parser in all cases. That doesn't seem worth it to me.

It feels like we're overcomplicating things. I was thinking all that we'd do is the equivalent of `test -f $image && imageisfile=true || imageisurl=true`. if the value passed in by the user doesn't match a file that exists on disk then assume it's a URL. This logic should sufficiently handle a file at `./https://example.com/config.ign`

bgilbert:
[...]

PR #126, opened 2020-01-07
install: add --ignition-url

jlebon:
This matches our support for it for the disk image.

A big difference though is that we don't usually have GPG signatures for
integrity checking and authenticity. At the very least, we refuse to
fetch over pure HTTP without passing `--ignition-hash` or
`--insecure-ignition`. (I didn't want to wrap this under the same
`--insecure` switch since one may still want to verify signatures on the
disk image itself).

Closes: #118

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -119,6 +119,8 @@ pub fn parse_args() -> Result<Config> {
                         .short("s")
                         .long("stream")
                         .value_name("name")
+                        .conflicts_with("image-file")
+                        .conflicts_with("image-url")
                         .help("Fedora CoreOS stream")
                         .takes_value(true),
                 )
@@ -126,6 +128,8 @@ pub fn parse_args() -> Result<Config> {
                     Arg::with_name("image-url")
                         .short("u")
                         .long("image-url")
+                        .conflicts_with("stream")
+                        .conflicts_with("image-file")
                         .value_name("URL")
                         .help("Manually specify the image URL")
                         .takes_value(true),
@@ -134,6 +138,8 @@ pub fn parse_args() -> Result<Config> {
                     Arg::with_name("image-file")
                         .short("f")
                         .long("image-file")
+                        .conflicts_with("stream")
+                        .conflicts_with("image-url")
                         .value_name("path")
                         .help("Manually specify a local image file")
                         .takes_value(true),
@@ -513,9 +519,6 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         .chain_err(|| format!("getting sector size of {}", &device))?
         .get();
 
-    // Build image location.  Ideally we'd use conflicts_with (and an
-    // ArgGroup for streams), but that doesn't play well with default
-    // arguments, so we manually prioritize modes.
     let location: Box<dyn ImageLocation> = if matches.is_present("image-file") {
         Box::new(FileLocation::new(
             matches.value_of("image-file").expect("image-file missing"),
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -15,6 +15,7 @@
 use clap::{crate_version, App, AppSettings, Arg, ArgMatches, SubCommand};
 use error_chain::bail;
 use reqwest::Url;
+use std::fs::{File, OpenOptions};
 use std::path::Path;
 
 use crate::blockdev::*;
@@ -37,7 +38,7 @@ pub enum Config {
 pub struct InstallConfig {
     pub device: String,
     pub location: Box<dyn ImageLocation>,
-    pub ignition: Option<String>,
+    pub ignition: Option<File>,
     pub ignition_hash: Option<IgnitionHash>,
     pub platform: Option<String>,
     pub firstboot_kargs: Option<String>,
@@ -572,6 +573,17 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
             )?)
         }
     };
+
+    let ignition = matches
+        .value_of("ignition-file")
+        .map(|file| {
+            OpenOptions::new()
+                .read(true)
+                .open(file)
+                .chain_err(|| format!("opening source Ignition config {}", file))
+        })
+        .transpose()?;
+
     // and report it to the user
     eprintln!("{}", location);
 
@@ -588,7 +600,7 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
     Ok(Config::Install(InstallConfig {
         device,
         location,
-        ignition: matches.value_of("ignition-file").map(String::from),
+        ignition,
         ignition_hash: matches
             .value_of("ignition-hash")
             .map(IgnitionHash::try_parse)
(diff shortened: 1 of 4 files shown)


src/source.rs
```diff
@@ -379,3 +379,93 @@ struct Artifact {
...
+/// Local Ignition config
+#[derive(Debug)]
+pub struct IgnitionFileLocation(PathBuf);
+
+/// Remote Ignition config via URL
+#[derive(Debug)]
+pub struct IgnitionUrlLocation(Url);
+
+impl IgnitionFileLocation {
+    fn read(&self) -> Result<std::fs::File> {
```
cgwalters: Confusing to have a trait and struct method with the same name.  Why not just have this code in the trait `read()`?

jlebon: Yeah thanks, it's confusing. I renamed it to `open_file()`, which then made me displeased with the name `read()`, which I just changed to `open()`.

It's not in the trait implementation because it's used in both `check()` and `open()`.

cgwalters: Right, but the `check()` trait method can call the trait `read()`.

jlebon: Right, but then we're allocating memory for the `Box` when all we need is `open(2)`.

dustymabe:
Thanks @jlebon for working on this!

That being said we should probably hold this until we resolve the conversation in https://github.com/coreos/coreos-installer/issues/118 . There were some concerns voiced.

Also another comment:

> A big difference though is that we don't usually have GPG signatures for
integrity checking and authenticity. At the very least, we refuse for
now to fetch over pure HTTP without passing `--insecure-ignition`. (I
didn't want to wrap this under the same `--insecure` switch since one
may still want to verify signatures on the disk image itself).

I think `--insecure` still checks the image signature (if I'm parsing rust correctly) if it exists, but doesn't require it to exist. I think my vote is to not add a new option and fold it under `--insecure`. That is unless there's more that I don't understand.

jlebon:
> I think my vote is to not add a new option and fold it under `--insecure`.

But then we'd no longer be failing the install if the signature is missing somehow. The way I see it is that the disk image and the Ignition config are at separate "levels". E.g. you could be fetching the disk image from a more public/central location (like the canonical `builds.coreos.fp.o` one) but wanting to apply an Ignition config that's more local.

dustymabe:
> But then we'd no longer be failing the install if the signature is missing somehow. The way I see it is that the disk image and the Ignition config are at separate "levels". E.g. you could be fetching the disk image from a more public/central location (like the canonical `builds.coreos.fp.o` one) but wanting to apply an Ignition config that's more local.

You have a good point. The implementation just feels a bit icky. Let's brainstorm. We have a `--insecure` option already (this only controls gpg checking, not SSL noverify). We want to add a new option. Should we add a new option for disabling SSL checking, but make it apply for grabbing ignition or talking to the remote to grab a disk image? Basically I'm trying to minimize the number of options we add and if we have a `--insecure-ignition` and then later a user says "but I want to download a disk image without SSL verification too" then we'll have to add another one. I say we either lump them all into `--insecure` or we make two classes (`--sslnoverify`, `--gpgnoverify`) and have those two classes apply to all types of operations related to those technologies.

lucab:
> That being said we should probably hold this until we resolve the conversation in #118 . There were some concerns voiced.

Not very strong concerns from my side. If you folks think this improves UX, let's land it.

> I say we either lump them all into --insecure

I have some prior art on this here: https://coreos.com/rkt/docs/latest/commands.html#global-options. We can tweak the current CLI to get `--insecure http,gpg-unsigned`. Those would affect both disks and config fetching.

bgilbert:
I don't feel great about this change; see https://github.com/coreos/coreos-installer/issues/118#issuecomment-572182986.  But if we pursue it, we should have separate e.g. `--ignition-file` and `--ignition-url` options for the reasons in https://github.com/coreos/coreos-installer/issues/117#issuecomment-572185348.

jlebon:
OK, let's hold this for now. I'll make a separate PR for just the "check the Ignition upfront" bit.

jlebon:
OK, rebased this now!

I still left `--insecure-ignition` there, though now that we have `--ignition-hash`, I made erroring on HTTP conditional on missing *both* `--insecure-ignition` and `--ignition-hash`.

> We can tweak the current CLI to get `--insecure http,gpg-unsigned`. Those would affect both disks and config fetching.

I'm not a fan of that because it doesn't allow you to express different levels of trust for different parts of the operation. E.g. I might be totally OK fetching the Ignition config over HTTP (because e.g. it's just a local address back to the hypervisor) but not for the image payload.

I think it would also require making it a new switch, e.g. `--insecure-flags`, because having options with optional (for backcompat) arguments is a footgun IMO.

src/download.rs
```diff
@@ -306,6 +306,36 @@ pub fn write_image(
...
+pub fn download_to_tempfile(url: &str, allow_insecure: bool, has_hash: bool) -> Result<File> {
+    if url.starts_with("http://") {
+        if !has_hash && !allow_insecure {
+            bail!("refusing to fetch HTTP Ignition config without --ignition-hash or --insecure-ignition")
+        }
+    } else if !url.starts_with("https://") {
+        bail!("unknown protocol for URL '{}'", url)
+    }
+
+    let mut f = tempfile::Builder::new()
```
bgilbert, line 318: Could we just use `tempfile::tempfile()` here to avoid the named file entirely?

src/download.rs
```diff
@@ -306,6 +306,36 @@ pub fn write_image(
     Ok(())
 }
 
+pub fn download_to_tempfile(url: &str, allow_insecure: bool, has_hash: bool) -> Result<File> {
+    if url.starts_with("http://") {
```
bgilbert, line 310: reqwest doesn't insist on the trailing slashes.  Let's just match against `http:` or `https:`.

src/cmdline.rs
```diff
@@ -126,6 +126,7 @@ pub fn parse_args() -> Result<Config> {
                     Arg::with_name("image-url")
                         .short("u")
                         .long("image-url")
+                        .conflicts_with("image-file")
```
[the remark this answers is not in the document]
jlebon: We removed the default stream argument in #187 since the default is now offline installs. Hmm, I think that means we can now make all three conflict with each other. Let me try that out and update this comment if that works well.

src/download.rs
jlebon: That's... interesting. I'm not sure if we want to leak that through. I'd rather just be more strict here.

jlebon:
Updated!

src/cmdline.rs
```diff
@@ -145,19 +146,33 @@ pub fn parse_args() -> Result<Config> {
...
                     Arg::with_name("ignition-hash")
                         .long("ignition-hash")
                         .value_name("digest")
                         .help("Digest (type-value) of the Ignition config")
                         .takes_value(true),
                 )
+                .arg(
+                    Arg::with_name("insecure-ignition")
+                        .long("insecure-ignition")
+                        .help("Allow HTTP Ignition config fetch without a hash"),
```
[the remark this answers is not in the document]
bgilbert: I'd still prefer this to be moved after `--insecure`.

PR #127, opened 2020-01-07
systemd: remove rudimentary network checking code; depend on NetworkManager

dustymabe:
Previously you could prematurely continue into /usr/bin/coreos-installer
before DNS was up. This commit improves network checking such that DNS
is checked (if needed) before continuing.

### DIFF
--- a/systemd/coreos-installer-service
+++ b/systemd/coreos-installer-service
@@ -42,19 +42,35 @@ copy_arg() {
 }
 
 # Wait for the network to be up by trying to HEAD a URL.
-# If an image URL was specified, use that.
-check_url="$(karg coreos.inst.image_url)"
-# Otherwise, an Ignition config URL.
-check_url="${check_url:-$(karg coreos.inst.ignition_url)}"
-# Otherwise, fall back to a globally accessible URL.  Try this last because
-# we might not have access to the Internet.
-check_url="${check_url:-http://fedoraproject.org/static/hotspot.txt}"
-# We'd like to disable the progress meter but we can only do that completely
-# with --silent, which also disables status updates.  Use the compact
-# progress bar instead, since at least it's smaller.
-curl --head --progress-bar --fail \
-    --retry 60 --retry-delay 1 --retry-connrefused \
-    -o /dev/null "${check_url}"
+# There are a few corner cases we have to account for:
+#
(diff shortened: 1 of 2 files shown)


dustymabe:
hold for now, testing

systemd/coreos-installer-service
```diff
@@ -42,19 +42,35 @@ copy_arg() {
...
+#       - We need to try both URLs because one could be IP
+#         based and the other could be DNS based. If both are
+#         IP based then we don't need to wait on DNS.
+#   - The user provided only image_url
+#       - This case is simple. Check image_url and continue.
+#   - The user did not provide image_url
+#       - This case implies we are going to contact a remote
+#         service (most likely fedora infra) so we need to
+#         test DNS is working befrore continuing. In this case
+#         we'll check the fedora hotspot globally accessible URL:
```
cgwalters: Hmm.  NetworkManager has this functionality built in.  And on RHEL it uses a RHEL endpoint.

Not blocking this but...is it really worth re-coding that check here?

I think a much simpler common network test is "do we have a default route" - GLib supports this.

What's the harm in having the pull fail later?

dustymabe: @cgwalters I'm +1 for letting networkmanager handle things here (especially since we're in the real root), but what is a sufficient requirement to add? Currently we have:

https://github.com/coreos/coreos-installer/blob/005d691026e3c08ccd9c2c08071724bb1ff1d331/systemd/coreos-installer.service#L4

Which doesn't seem to be enough.

dustymabe: Also will the NM check block on having DNS set up?

dustymabe: > What's the harm in having the pull fail later?

The installer seems to barf and not do any retries (probably should be fixed). I get a failed install.

cgwalters: https://github.com/coreos/coreos-installer/pull/128

dustymabe: > #128

Thanks! I included that commit in this PR and tested everything together. I'd like to keep my original commit for history, even though it's not needed, in case we need to come back to it for any reason.

dustymabe:
This is ready for further review

dustymabe:
Thanks @cgwalters for the help with this

PR #128, opened 2020-01-07
systemd: Add After=network-online.target

cgwalters:
From https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/
"Cut the crap! How do I make sure that my service starts after the network is really online?"

We were just missing this one.  Came out of a discussion
about testing for network availability in
https://github.com/coreos/coreos-installer/pull/127

### DIFF
--- a/systemd/coreos-installer.service
+++ b/systemd/coreos-installer.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=CoreOS Installer
 Before=coreos-installer.target
+After=network-online.target
 Wants=network-online.target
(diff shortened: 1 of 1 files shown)


cgwalters:
(Not tested locally)

dustymabe:
I'll give this a spin

dustymabe:
Thanks @cgwalters - I included this commit into #127 and tested that everything worked . I'd like to include it over there instead of merging this PR because I'd like to keep my first commit over there to preserve it in the git history if that's OK.

PR #131, opened 2020-01-08
install: check for Ignition config upfront

jlebon:
Otherwise it's a painful experience to fail the install after having
already downloaded the image and written it to disk because of a typo in
the filename.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -40,6 +40,19 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         }
     }
 
+    // if an Ignition config is provided, open it upfront so we don't waste time downloading the
+    // main image and writing it to disk first
+    let ignition = config
+        .ignition
+        .as_ref()
+        .map(|f| {
+            OpenOptions::new()
+                .read(true)
+                .open(f)
+                .chain_err(|| format!("opening source Ignition config {}", f))
+        })
+        .transpose()?;
+
     // open output; ensure it's a block device and we have exclusive access
     let mut dest = OpenOptions::new()
         .write(true)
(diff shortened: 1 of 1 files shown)


jlebon:
I didn't feel it was worth deduping with the other bit of code later down that does the same thing, but I can if folks prefer.

src/install.rs
```diff
@@ -40,6 +40,14 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         }
     }
 
+    // if an Ignition config is provided, check upfront that we can open it
```
dustymabe: optional: maybe augment the comment with something like: "so we don't waste time downloading the disk image first"

src/install.rs
```diff
@@ -40,6 +40,14 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+    if let Some(ignition) = &config.ignition {
+        OpenOptions::new()
+            .read(true)
+            .open(ignition)
+            .chain_err(|| format!("opening source Ignition config {}", ignition))?;
```
lucab: What about keeping this around as an `Option<File>` and passing it directly to `write_disk`?

(Not strictly relevant in this case, but current approach is a pattern prone to TOCTOU, thus I try to avoid it)

jlebon: Thanks, done!

src/install.rs
jlebon: Augmented!

src/install.rs
```diff
@@ -40,6 +40,19 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+    let ignition = config
+        .ignition
+        .as_ref()
+        .map(|f| {
+            OpenOptions::new()
+                .read(true)
+                .open(f)
+                .chain_err(|| format!("opening source Ignition config {}", f))
+        })
+        .transpose()?;
```
lucab: :heart:

PR #132, opened 2020-01-08
Cargo.toml: remove package.metadata.release.upload-doc option

dustymabe:
Support for this was removed upstream [1] and now I get can error
like this when running cargo-release:

```
Fatal: Invalid TOML file format: unknown field `upload-doc`
```

[1] https://github.com/sunng87/cargo-release/commit/739d746


bgilbert:
LGTM

PR #135, opened 2020-01-09
Just retry image fetches 5 times

cgwalters:
OpenShift CI hit a provisioning failure in Packet.

Reading `man curl`, it's not just code `22` we need to retry on
but also things like:

- `7      Failed to connect to host.`
- `18     Partial file. Only a part of the file was transferred.`
- `92     Stream error in HTTP/2 framing layer.`

And the man page explicitly says others could be added in the future.
We could make this nicer but this is the legacy branch.

Let's just retry unconditionally a few times.

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -463,10 +463,12 @@ get_img_url() {
         curl -LsIf $IMAGE_URL >/tmp/image_info 2>&1
         RETCODE=$?
         if [ $RETCODE -ne 0 ]
-        then
-            if [[ $RETCODE -eq 22 && $retry -lt 5 ]]
+            if [[ $retry -lt 5 ]]
             then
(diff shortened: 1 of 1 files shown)


cgwalters:
And based on https://github.com/coreos/coreos-installer/pull/136 we should probably just be using `curl --retry`...will look at that

lucab:
@cgwalters thoughts on just using `curl --retries X` (here and in the other invocations too)?

dustymabe:
I believe this was obsoleted by #136

lucab:
#136 was for `master` while this is targeting the `legacy` branch of the installer. I think we are interested in both.

dustymabe:
oops. I missed that point. Sorry.

bgilbert:

PR #136, opened 2020-01-09
service: add some retries when downloading

lucab:
This adds some retries to curl when downloading Ignition config.
It should give us some room for resiliency in case of transient
network hiccups.
Proactive measure, this has not yet been observed/reported as a
direct cause of failures in the wild.

Related: https://github.com/coreos/coreos-installer/pull/127


lucab:
@cgwalters it looks like we independently picked N=3 and N=5, which one do you prefer? Upper bounds for the waiting time are 7s for the former and 31s for the latter (if I did the math right).

dustymabe:
> we independently picked N=3 and N=5, which one do you prefer?

5 works for me. doesn't hurt and we only wait if it isn't available.

lucab:
Updated from N=3 to N=5. Keeping previous stamps and merging.

PR #140, opened 2020-01-14
tests/iso: test Ignition cpio read/write roundtrip

lucab:
This adds a basic smoke test to check that reading/writing logic for
embedded Ignition config can roundtrip without issues.


lucab:
CI was red due to a network issue between Travis and crates.io. Re-triggered, it's fine now. Merging.

PR #141, opened 2020-01-16
cargo: update to reqwest 0.10 and new API

lucab:
This updates to `reqwest` 0.10. The default API is now async, but I simply updated and sticked to the blocking API which is good enough for our usecases.

Closes: https://github.com/coreos/coreos-installer/pull/139

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: rust
 rust:
-  - 1.37.0  # pinned toolchain for clippy
-  - 1.37.0  # minimum supported toolchain
+  - 1.40.0  # pinned toolchain for clippy
+  - 1.39.0  # minimum supported toolchain
   - stable
   - beta
   - nightly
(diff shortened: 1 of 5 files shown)


lucab:
This also requires bumping minimum supported Rust version to 1.39.0 as the `bytes` dependency [needs that](https://github.com/tokio-rs/bytes/blob/v0.5.3/azure-pipelines.yml#L23) now.

dustymabe:
> This also requires bumping minimum supported Rust version to 1.39.0 as the `bytes` dependency [needs that](https://github.com/tokio-rs/bytes/blob/v0.5.3/azure-pipelines.yml#L23) now.

Looks like Fedora has 1.40.0. https://apps.fedoraproject.org/packages/rust

However I wonder if we should consider RHEL here since we eventually want to drop the old installer.

jlebon:
> However I wonder if we should consider RHEL here since we eventually want to drop the old installer.

The minimum version across our Rust projects track the latest available in RHEL. See e.g. https://github.com/coreos/rpm-ostree/commit/13f554ea5c4b296b00f3f774756afff4bde994ca.

PR #153, opened 2020-02-11
s390x: add support for IBM Z DASD disks 

nikita-dubrovskii:
Installation on s390 dasd disks is a bit tricky,
this patch makes coreos-installer able to format,
create partitions, and finally install CoreOS on s390.

### DIFF
--- a/scripts/coreos-installer-service
+++ b/scripts/coreos-installer-service
@@ -10,6 +10,9 @@ PERSIST_KERNEL_NET_PARAMS=("ipv6.disable" "net.ifnames" "net.naming-scheme")
 # List from https://www.mankier.com/7/dracut.cmdline#Description-Network
 PERSIST_DRACUT_NET_PARAMS=("ip" "ifname" "rd.route" "bootdev" "BOOTIF" "rd.bootif" "nameserver" "rd.peerdns" "biosdevname" "vlan" "bond" "team" "bridge" "rd.net.timeout.carrier" "coreos.no_persist_ip")
 
+# IBM S390X params to persist
+PERSIST_S390X_PARAMS=("rd.dasd" "rd.zfcp" "rd.znet" "zfcp.allow_lun_scan" "cio_ignore")
+
 args=("install")
 
 cmdline=( $(</proc/cmdline) )
@@ -61,8 +64,11 @@ if [ -n "${ignition_url}" -a "${ignition_url}" != "skip" ]; then
     args+=("--ignition-url" "${ignition_url}" "--insecure-ignition")
 fi
 
-# First-boot kernel arguments. Filter out any args that aren't in our
-# list of networking arguments to forward and store them in $firstboot_args
+# Forward whitelisted kernel arguments to the installed system. We have
+# separate sets of whitelists for first-boot kargs and persistent kargs.
+# Only pass --firstboot-kargs if additional networking options have been
+# specified, since the default in coreos-assembler specifies
+# `rd.neednet=1 ip=dhcp` when no options are persisted.
 firstboot_args=""
 for item in "${cmdline[@]}"; do
     for param in "${PERSIST_KERNEL_NET_PARAMS[@]}" "${PERSIST_DRACUT_NET_PARAMS[@]}"; do
@@ -70,10 +76,12 @@ for item in "${cmdline[@]}"; do
             firstboot_args+="${item} "
         fi
     done
+    for param in "${PERSIST_S390X_PARAMS[@]}"; do
+        if [[ $item =~ ^$param(=.*)?$ ]]; then
+            args+=("--append-karg" "${item}")
+        fi
+    done
 done
-# Only pass firstboot-kargs if additional networking options have been
-# specified, since the default in ignition-dracut specifies
-# `rd.neednet=1 ip=dhcp` when no options are persisted
 if [ -n "${firstboot_args}" ]; then
     args+=("--firstboot-args" "rd.neednet=1 ${firstboot_args}")
 fi
@@ -94,4 +102,3 @@ udevadm settle
 # Install
 echo "coreos-installer ${args[@]}"
 coreos-installer "${args[@]}"
-
--- a/src/install.rs
+++ b/src/install.rs
@@ -24,6 +24,8 @@ use crate::cmdline::*;
 use crate::download::*;
 use crate::errors::*;
 use crate::io::*;
+#[cfg(target_arch = "s390x")]
+use crate::s390x;
 use crate::source::*;
 
 /// Integrity verification hash for an Ignition config.
@@ -96,6 +98,13 @@ pub fn install(config: &InstallConfig) -> Result<()> {
             eprintln!("Signature not found; skipping verification as requested");
         } else {
             bail!("--insecure not specified and signature not found");
+        }
+    }
+
+    #[cfg(target_arch = "s390x")]
+    {
+        if is_dasd(config)? {
+            s390x::prepare_dasd(&config)?;
         }
     }
 
@@ -179,11 +188,17 @@ fn write_disk(
     let sector_size = get_sector_size(dest)?;
 
     // copy the image
+    #[allow(clippy::match_bool, clippy::match_single_binding)]
+    let image_copy = match is_dasd(config)? {
+        #[cfg(target_arch = "s390x")]
+        true => s390x::image_copy_s390x,
+        _ => image_copy_default,
+    };
     write_image(
         source,
         dest,
         Path::new(&config.device),
-        image_copy_default,
+        image_copy,
         true,
         Some(sector_size),
     )?;
@@ -196,6 +211,7 @@ fn write_disk(
         || config.delete_kargs.is_some()
         || config.platform.is_some()
         || config.network_config.is_some()
+        || cfg!(target_arch = "s390x")
     {
         let mount =
             Disk::new(&config.device).mount_partition_by_label("boot", mount::MsFlags::empty())?;
@@ -225,6 +241,8 @@ fn write_disk(
         if let Some(network_config) = config.network_config.as_ref() {
             copy_network_config(mount.mountpoint(), network_config)?;
         }
+        #[cfg(target_arch = "s390x")]
+        s390x::install_bootloader(mount.mountpoint(), &config.device)?;
     }
 
     Ok(())
@@ -461,6 +479,11 @@ fn clear_partition_table(dest: &mut File, table: &mut dyn PartTable) -> Result<(
         .chain_err(|| "syncing partition table to disk")?;
     table.reread()?;
     Ok(())
+}
+
+fn is_dasd(config: &InstallConfig) -> Result<bool> {
+    let (target, _) = resolve_link(&config.device)?;
+    Ok(target.to_string_lossy().starts_with("/dev/dasd"))
 }
 
 #[cfg(test)]
--- a/src/install.rs
+++ b/src/install.rs
@@ -179,7 +179,14 @@ fn write_disk(
     let sector_size = get_sector_size(dest)?;
 
     // copy the image
-    write_image(source, dest, true, Some(sector_size))?;
+    write_image(
+        source,
+        dest,
+        Path::new(&config.device),
+        image_copy_default,
+        true,
+        Some(sector_size),
+    )?;
     table.reread()?;
 
     // postprocess
--- a/src/s390x/dasd.rs
+++ b/src/s390x/dasd.rs
@@ -1,0 +1,314 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use error_chain::bail;
+use gptman::GPT;
+use std::fs::{read_to_string, File};
+use std::io::{self, Cursor, Read, Seek, SeekFrom, Write};
+use std::num::NonZeroU32;
+use std::os::unix::io::AsRawFd;
+use std::path::Path;
+use std::process::{Command, Stdio};
+
+use crate::blockdev::{get_sector_size, udev_settle};
+use crate::cmdline::*;
+use crate::errors::*;
+use crate::io::{copy, copy_exactly_n};
+
+/////////////////////////////////////////////////////////////////////////////
+// IBM DASD Support
+/////////////////////////////////////////////////////////////////////////////
+
+#[derive(Debug)]
+struct Range {
+    in_offset: u64,
+    out_offset: u64,
+    length: u64,
+}
+
+pub fn prepare_dasd(config: &InstallConfig) -> Result<()> {
+    low_level_format(&config.device)?;
+    if is_invalid(&config.device)? {
+        eprintln!("Disk {} is invalid, formatting", &config.device);
+        default_format(&config.device)?
+    }
+    Ok(())
+}
+
+pub fn image_copy_s390x(
+    first_mb: &[u8],
+    source: &mut dyn Read,
+    dest: &mut File,
+    dest_path: &Path,
+) -> Result<()> {
+    let (ranges, partitions) = partition_ranges(first_mb, dest)?;
+    make_partitions(
+        dest_path
+            .to_str()
+            .chain_err(|| format!("couldn't encode path {}", dest_path.display()))?,
+        &partitions,
+    )?;
+
+    // copy each partition
+    eprintln!("Installing to {}", dest_path.display());
+    let mut buf = [0u8; 1024 * 1024];
+    // there shouldn't be any partition data in the first MiB, so don't
+    // worry about copying first_mb
+    let mut cursor: u64 = 1024 * 1024;
+    let sink = &mut io::sink();
+    for range in ranges.iter() {
+        if range.in_offset < cursor {
+            bail!(
+                "found partition at {} when current stream location is {}",
+                range.in_offset,
+                cursor
+            );
+        }
+        if range.in_offset > cursor {
+            copy_exactly_n(source, sink, range.in_offset - cursor, &mut buf)
+                .chain_err(|| "sinking input data")?;
+            cursor = range.in_offset;
+        }
+        dest.seek(SeekFrom::Start(range.out_offset))
+            .chain_err(|| "seeking output")?;
+        copy_exactly_n(source, dest, range.length, &mut buf).chain_err(|| "copying partition")?;
+        cursor += range.length;
+    }
+
+    // close out the stream
+    copy(source, sink).chain_err(|| "reading remainder of stream")?;
+
+    Ok(())
+}
+
+/// Generate partition table entries and byte ranges to copy
+fn partition_ranges(header: &[u8], device: &mut File) -> Result<(Vec<Range>, Vec<String>)> {
+    let bytes_per_block: u64 = get_sector_size(device)?.get().into();
+    let blocks_per_track: u64 = get_sectors_per_track(device)?.get().into();
+
+    let gpt = GPT::read_from(&mut Cursor::new(header), bytes_per_block)
+        .chain_err(|| "reading GPT of source image")?;
+
+    let mut ranges = Vec::new();
+    let mut partitions = Vec::new();
+    let mut start_track: u64 = 2; // the first 2 tracks of the ECKD DASD are reserved
+    let entries = || gpt.iter().filter(|(_, pt)| pt.is_used());
+    let (last_partition, _) = entries()
+        .last()
+        .chain_err(|| "source image has no partitions")?;
+
+    for (i, pt) in entries() {
+        let blocks = pt.ending_lba - pt.starting_lba + 1;
+        let end_track = start_track + (blocks + blocks_per_track - 1) / blocks_per_track - 1;
+
+        ranges.push(Range {
+            in_offset: pt.starting_lba * bytes_per_block,
+            out_offset: start_track * blocks_per_track * bytes_per_block,
+            length: blocks * bytes_per_block,
+        });
+
+        if i == last_partition {
+            partitions.push(format!("[{}, last, native]", start_track));
+        } else {
+            partitions.push(format!("[{}, {}, native]", start_track, end_track));
+        };
+        start_track = end_track + 1;
+    }
+    // partitions should be in offset order, but just to be sure
+    ranges.sort_unstable_by_key(|r| r.in_offset);
+    Ok((ranges, partitions))
+}
+
+/// Get disk bus id
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+fn bus_id(dasd: &str) -> Result<String> {
+    let cmd = Command::new("lszdev")
+        .arg("-n")
+        .arg("-c")
+        .arg("ID")
+        .arg("--by-node")
+        .arg(dasd)
+        .stderr(Stdio::inherit())
+        .output()
+        .chain_err(|| format!("executing lszdev on {}", dasd))?;
+    if !cmd.status.success() {
+        bail!("lszdev on {} failed", dasd);
+    }
+    Ok(std::str::from_utf8(&cmd.stdout)
+        .chain_err(|| "decoding lszdev output")?
+        .trim_end()
+        .to_string())
+}
+
+/// Check if disk is already formatted or not
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+fn is_formatted(dasd: &str) -> Result<bool> {
+    let id = bus_id(dasd)?;
+    let path = format!("/sys/bus/ccw/devices/{}/status", id);
+    let contents = read_to_string(&path).chain_err(|| format!("reading {}", path))?;
+    Ok(!contents.contains("unformatted"))
+}
+
+/// Check if disk is valid or not
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+fn is_invalid(dasd: &str) -> Result<bool> {
+    let cmd = Command::new("fdasd")
+        // we're looking for a hardcoded string in the output
+        .env("LC_ALL", "C")
+        .arg("-p")
+        .arg(dasd)
+        .output()
+        .chain_err(|| format!("executing fdasd -p {}", dasd))?;
+    if !cmd.status.success() {
+        bail!("fdasd -p {} failed", dasd);
+    }
+    Ok(std::str::from_utf8(&cmd.stdout)
+        .chain_err(|| "decoding fdasd output")?
+        .contains("disk label block is invalid"))
+}
+
+/// Perform low-level format. This step is necessary before any further disk usage
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+fn low_level_format(dasd: &str) -> Result<()> {
+    if is_formatted(dasd)? {
+        eprintln!("Skipping low-level format for {}", dasd);
+        return Ok(());
+    }
+    eprintln!("Performing low-level format for {}", dasd);
+    let status = Command::new("dasdfmt")
+        .arg("--blocksize")
+        .arg("4096")
+        .arg("--disk_layout")
+        .arg("cdl")
+        .arg("--mode")
+        .arg("full")
+        .arg("-y")
+        .arg("-p")
+        .arg(dasd)
+        .status()
+        .chain_err(|| format!("failed to execute dasdfmt on {}", dasd))?;
+    if !status.success() {
+        bail!("dasdfmt on {} failed", dasd);
+    }
+    udev_settle()?;
+    Ok(())
+}
+
+/// Format disk and create partitions
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+/// * `partitions` - configuration strings
+fn make_partitions(dasd: &str, partitions: &[String]) -> Result<()> {
+    if partitions.len() > 3 {
+        // fdasd silently ignores partitions after the first 3
+        bail!("Can't create {} partitions, maximum 3", partitions.len());
+    }
+    let mut config = partitions.join("\n");
+    config.push('\n');
+    if try_format(dasd, &config).is_err() {
+        default_format(dasd)?;
+        try_format(dasd, &config)?;
+    }
+    Ok(())
+}
+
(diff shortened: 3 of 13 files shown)


nikita-dubrovskii:
Hi Collin. We had a small discussion in Brno (DevConf). Could you please review this patch?

src/s390x/zipl.rs
```diff
@@ -0,0 +1,148 @@
+use super::dasd::check;
```
bgilbert: Needs license block.

src/s390x/zipl.rs
```diff
@@ -0,0 +1,148 @@
...
+// IBM Z Bootloader Support
+/////////////////////////////////////////////////////////////////////////////
+
+/// Generate zipl config and set dasd as boot device
+///
+/// # Arguments
+/// * `boot` - path to boot partition mountpoint, i.e. smth like /boot
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+pub fn install_bootloader<P: AsRef<Path>>(boot: P, dasd: &str) -> Result<()> {
+    eprintln!("Installing_bootloader");
```
bgilbert: s/_/ /

src/s390x/zipl.rs
```diff
@@ -0,0 +1,148 @@
...
+    let boot_config = boot_config.as_ref();
+    let zipl_config = zipl_config.as_ref();
+    let mut contents = read_options(&boot_config)?;
+    let pat = "options ";
+    let idx = contents
+        .find(pat)
+        .chain_err(|| format!("cannot find: \'{}\'", pat))?
+        + pat.len();
+
+    let s390_options = read_options("/tmp/s390x_opts")?;
```
bgilbert: Options should come from the coreos-installer command line.  The `/tmp` files used by the old installer script are no longer used.

src/s390x/zipl.rs
```diff
@@ -0,0 +1,148 @@
...
+    contents.insert_str(idx, &s390_options);
+
+    eprintln!("Writing boot options");
+    std::fs::write(&boot_config, &contents)
+        .chain_err(|| format!("writing {}", boot_config.display()))?;
+
+    eprintln!("Generating zipl config");
+    let mut contents = &contents[idx..contents.len()];
+    if let Some(idx) = contents.find('\n') {
+        contents = &contents[0..idx];
```
bgilbert: We're already pulling in `regex`, so let's use it rather than manual string manipulation.

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+    Ok(())
+}
+
+/// Return disk's "blocks per track" and "bytes per block" config values as tuple
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+fn disk_config(dasd: &str) -> Result<(u64, u64)> {
```
bgilbert: Let's return a proper struct here for clarity.

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+    let find = |data: &str, pat: &str| -> Result<u64> {
+        let s = data
+            .find(pat)
+            .chain_err(|| format!("cannot find: \'{}\'", pat))?
+            + pat.len();
+        let e = data[s..].find('\n').chain_err(|| "cannot find newline")? + s;
+        u64::from_str(&data[s..e]).chain_err(|| format!("parsing: {}", data))
+    };
+    let cmd = Command::new("fdasd").arg("-p").arg(dasd).output()?;
```
bgilbert: In general, all errors should be filtered through `chain_err()`.

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+fn disk_config(dasd: &str) -> Result<(u64, u64)> {
+    let find = |data: &str, pat: &str| -> Result<u64> {
+        let s = data
+            .find(pat)
+            .chain_err(|| format!("cannot find: \'{}\'", pat))?
+            + pat.len();
+        let e = data[s..].find('\n').chain_err(|| "cannot find newline")? + s;
```
bgilbert: Might make more sense to use `regex` here?

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+    check(status).chain_err(|| "low-level format")
+}
+
+/// Format disk and create partitions
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+/// * `partitions` - configuration strings
+fn make_partitions(dasd: &str, partitions: &[String]) -> Result<()> {
+    let config_filename = "/tmp/fdasd.conf";
```
bgilbert: This enables symlink attacks, and will cause problems if multiple instances of coreos-installer run at the same time.  Should use a tempfile.

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    write(config_filename, partitions.join("\n"))
+        .chain_err(|| format!("writing {}", config_filename))?;
+
+    if try_format(dasd, &config_filename).is_err() {
+        default_format(dasd)?;
+        try_format(dasd, &config_filename)?;
```
bgilbert: Is an auto-format followed by a config-based format likely to actually fix the problem?

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/// Check if disk is valid or not
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+fn is_invalid(dasd: &str) -> Result<bool> {
+    let cmd = Command::new("fdasd").arg("-p").arg(dasd).output()?;
+    let output = cmd.stdout.iter().map(|&c| c as char).collect::<String>();
+    Ok(output.contains("disk label block is invalid"))
```
bgilbert: Does `fdasd` output localized strings?

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+/// * `source` - CoreOS image source
+fn save_image(source: &mut ImageSource) -> Result<PathBuf> {
+    let mut image_filename = PathBuf::new();
+    image_filename.push("/tmp/");
+    image_filename.push(
+        source
+            .filename
+            .trim_end_matches(".gz")
+            .trim_end_matches(".xz"),
+    );
```
bgilbert: If we absolutely must save to a file, we should use a tempfile.

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+        let mut opts = Opts::default();
+        opts.input_block_size = bytes as usize;
+        opts.output_block_size = bytes as usize;
+        opts.iflags = dd_lib::opts::IFlag::FULLBLOCK;
+        opts.input_file = Some(image.to_str().unwrap().to_string());
+        opts.output_file = Some(dasd.to_string());
+        opts.input_seek = Some(skip as usize);
+        opts.count = Some(size as usize);
+        opts.output_seek = Some(seek as usize);
+        dd_options.push(opts);
```
bgilbert: It looks like we're doing a straightforward copy, without any fancy conversions.  Do we need to pull in `dd_lib` rather than just doing the I/O ourselves?

src/s390x/dasd.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/// Download and unpack CoreOS image to file.
+/// Return path to raw image or error
+///
+/// # Arguments
+/// * `source` - CoreOS image source
+fn save_image(source: &mut ImageSource) -> Result<PathBuf> {
+    let mut image_filename = PathBuf::new();
+    image_filename.push("/tmp/");
```
bgilbert: This requires us to have enough RAM for the entire uncompressed image, assuming `/tmp` is a tmpfs.

Is it actually necessary for us to prefetch the file?  It seems as though streaming should work: we can fetch enough of the image to get the partition table, do our partition table processing, and then stream each partition to disk in turn.

src/s390x/dasd.rs
nikita-dubrovskii: unfortunately now. if disk has no valid table, than it should be auto-formatted

src/s390x/dasd.rs
nikita-dubrovskii: it's hardcoded c-string:
 `printf("\nCannot show requested information because "
			       "the disk label block is invalid\n");`

src/s390x/dasd.rs
nikita-dubrovskii: sorry, misunderstood the question. yes, auto-format fixes all issues, and than config-based format  works fine. (it may fail only if config itslef is not valid )

src/s390x/dasd.rs
nikita-dubrovskii: i'll check it.

src/s390x/dasd.rs
nikita-dubrovskii: dd is very convenient and less err-prone  way of copying. that's why it used here.

src/s390x/dasd.rs
nikita-dubrovskii: what's about making a separate ISSUE/patch and replace all manual string processing with regex if/when required?

src/install.rs
```diff
@@ -118,6 +125,9 @@ fn write_disk(
         if let Some(platform) = config.platform.as_ref() {
             write_platform(mount.mountpoint(), platform)?;
         }
+
+        #[cfg(target_arch = "s390x")]
+        crate::s390x::zipl::install_bootloader(mount.mountpoint(), &config.device)?;
```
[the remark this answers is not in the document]
nikita-dubrovskii: it's zVM specific: we have to mark dasd as next boot-device.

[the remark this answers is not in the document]
bgilbert: That doesn't answer my question, though.  Currently this function is only called if we're writing an Ignition config or changing kernel arguments or changing the platform ID.  Are those exactly the cases in which we need to install a bootloader?

src/s390x/dasd.rs
bgilbert: I think it'd be better to do things cleanly from the start.  Is there some reason regexes would be difficult to apply here?

src/s390x/dasd.rs
bgilbert: In cases where we care about specific output strings, we should run the program with the `LC_ALL` environment variable set to `C` in case localization is added in the future.

src/s390x/zipl.rs
```diff
@@ -0,0 +1,191 @@
...
+    contents.retain(|c| c != '\n' && c != '\r');
+    contents.push_str(" ");
+    Ok(contents)
+}
+
+fn get_s390x_kargs() -> Result<String> {
+    let re = Regex::new(
+        r#"rd.dasd=[[[:alnum:]].]*|rd.zfcp=[[[:alnum:]].,=:!]*|rd.znet=[[[:alnum:]].,=]*|zfcp.allow_lun_scan=[\d]|cio_ignore=[[[:alnum:]].,=:!]*"#)
+        .chain_err(|| "compiling regex")?;
+    let cmdline = read_options("/proc/cmdline")?;
```
bgilbert: In general, coreos-installer doesn't rely on external state like this.  Instead, it takes configuration explicitly from its own command line.  We can do otherwise if there's a good reason, though.  What's the purpose of these kernel arguments?  Do we always want to use exactly the ones from the current boot?

src/s390x/dasd.rs
nikita-dubrovskii: so, i've checked truncated images - it doesn't work (fdisk fails with `GPT PMBR size mismatch`). we have to fetch the entire image

src/install.rs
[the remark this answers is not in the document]
nikita-dubrovskii: I removed all changes to /boot, similar code was in legacy branch. Now functions generates only config for zipl

src/s390x/dasd.rs
nikita-dubrovskii: np, now regex are in use

src/s390x/zipl.rs
tuan-hoang1: @nikita-dubrovskii: I think some thing like this : https://github.com/coreos/coreos-installer/blob/ce8e9a8ccbf4fb0b5703d367edea5662e3137f8a/dracut/30coreos-installer/parse-coreos.sh#L77 instead of `/proc/cmdline`.

tuan-hoang1: And for the coreos-installer bin itself : https://github.com/coreos/coreos-installer/blob/ce8e9a8ccbf4fb0b5703d367edea5662e3137f8a/coreos-installer#L702

nikita-dubrovskii: files in /tmp are no longer available

tuan-hoang1: another way is parse those kcmdline you want in /proc/cmdline in `systemd/coreos-installer-generator`

tuan-hoang1: `systemd/coreos-installer-service` : so that you can make the s390x parameters in /proc/cmdline you want as an option for `coreos-installer`. Like ``--dasd-id 0123`.

nikita-dubrovskii: why should i? don't see any benefits in doing this. main question - do we need all s390x specific + networking (firstboot) args at all? for some reasons legacy was using all of them. @tuan-hoang1 why it was like that in legacy?

nikita-dubrovskii: i've checked CoreOS on zVM without s390x specific args - it fails during boot. But firstboot args could be ignored

cgwalters:
Does this have any intersection with the discussions around coreos-installer and multipath?  I think that one was going to end up on the legacy branch first?

src/s390x/zipl.rs
```diff
@@ -0,0 +1,172 @@
...
+
+fn initramfs<P: AsRef<Path>>(boot: P) -> Result<PathBuf> {
+    let mut path = PathBuf::new();
+    path.push(boot);
+    path.push("ostree");
+    find_file(&path, "initram")
+}
+
+fn get_s390x_kargs() -> Result<String> {
+    let cmdline = read_to_string("/proc/cmdline").chain_err(|| "reading cmdline")?;
```
cgwalters: Parsing the kernel cmdline can be subtle, I think https://github.com/coreos/afterburn/ has some Rust code relevant for that?   We also have [C code in libostree](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-kernel-args.c).

src/s390x/zipl.rs
```diff
@@ -0,0 +1,172 @@
...
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use tempfile::NamedTempFile;
+use walkdir::WalkDir;
+
+/////////////////////////////////////////////////////////////////////////////
+// IBM Z Bootloader Support
+/////////////////////////////////////////////////////////////////////////////
+
+/// Generate zipl config and set dasd as boot device
```
cgwalters: Hmm interesting.  We have code for this in coreos-assembler too.  One thing perhaps we could do is factor out a hidden `coreos-installer internals s390x-update-zipl` command to share it?

cgwalters:
My overall feeling here is if you've tested this and the rest of the s390x team is happy with it, I'm inclined to merge.

nikita-dubrovskii:
> Does this have any intersection with the discussions around coreos-installer and multipath? I think that one was going to end up on the legacy branch first?

No intersection, Tuan is working on multipath for legacy branch. And i've started working on multipath for master branch

nikita-dubrovskii:
> OK so I only skimmed this, and I only have superficial knowledge of what we need for dasd; I know we've talked about it a few times and my recollection is that it boils down to the kernel needing to "activate" devices effectively?
Exactly, we have to prepare (format & make partitions) DASD device and then tell kernel about it
 `rd.dasd=xxxxx ....`  (and networking as well `rd.znet=xxxxx ..`)

> Is there a good reference for this stuff? OK from a quick search the top two (here) results are:
> 
>     * https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/installation_guide/chap-post-installation-configuration-s390
> 
>     * https://www.kernel.org/doc/Documentation/s390/DASD
Yes, i read this. 

> Which look useful. Though what would help me in this PR is perhaps a little bit more elaboration on what "support" here means. Perhaps a block comment in the `s390x.rs` file or a `README-s390x.md`?
Ok, makes sense

> Also, did you learn Rust just for this? If so (or actually regardless) it's excellent code and really well done!
Yep, i've tried several times before, but was too busy with C++ projects and had no chance to do smth in Rust. thx ))

src/s390x/zipl.rs
nikita-dubrovskii: Indeed, but here it's not about parsing, just copying s390x specific `key=value` options

src/s390x/zipl.rs
nikita-dubrovskii: Hmm. Maybe in the future? I don't have a strong opinion about that

tuan-hoang1:
LGTM !

src/s390x/dasd.rs
```diff
@@ -0,0 +1,336 @@
...
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+fn get_disk_config(dasd: &str) -> Result<DiskConfig> {
+    let cmd = Command::new("fdasd")
+        .env("LC_ALL", "C")
+        .arg("-p")
+        .arg(dasd)
+        .output()
+        .chain_err(|| format!("executing fdasd -p {}", dasd))?;
+    let output = cmd.stdout.iter().map(|&c| c as char).collect::<String>();
```
cgwalters, line 298:
```
let output = str::from_utf8(&cmd.stdout);
```
or so.

src/s390x/dasd.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+impl FromStr for DiskConfig {
+    type Err = Error;
+
+    fn from_str(data: &str) -> Result<Self> {
+        fn get_value(s: &str, data: &str) -> Result<u32> {
+            let pt = format!(r"{}\D+(?P<v>\d+)", s);
+            Regex::new(&pt)
+                .chain_err(|| format!("compiling {}", pt))?
```
cgwalters, line 311: Personally for a regexp that's static in the code I think it's ok to use `.unwrap()` or `.expect("regex")`.  (But this is fine as is!)

src/s390x/zipl.rs
```diff
@@ -0,0 +1,201 @@
...
+    let mut zipl_config = NamedTempFile::new().chain_err(|| "creating zipl config file")?;
+    generate_zipl_config(&bootconf, &mut zipl_config, &firstboot_kargs)
+        .chain_err(|| "writing boot options")?;
+
+    let status = Command::new("zipl")
+        .arg("--verbose")
+        .arg("-p")
+        .arg(zipl_config.path())
+        .arg("-i")
+        .arg(vmlinuz.to_str().unwrap())
```
cgwalters, line 56: This is probably another case where we want to use
`vmlinuz.as_os_str()`

src/s390x/dasd.rs
```diff
@@ -0,0 +1,336 @@
...
+    let mut table = ctx.get_partitions().chain_err(|| "reading partitions")?;
+    let mut count = table.nents();
+
+    for mut pt in &mut table {
+        let uuid = pt.uuid().unwrap_or_else(|_| "UUID unknown".to_string());
+        let skip = pt.start().chain_err(|| format!("{}: no start", uuid))?;
+        let size = pt.size().chain_err(|| format!("{}: no size", uuid))?;
+        let seek = config.blocks_per_track as u64 * start;
+        let end = (size + config.blocks_per_track as u64) / 12;
+        eprintln!("{} start: {} size: {}", uuid, skip, size);
```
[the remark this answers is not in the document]
nikita-dubrovskii: actually now, all `eprinln!` are here to make installation progress more explicit/detailed. 
but i can remove it. should i?

src/s390x/dasd.rs
```diff
@@ -0,0 +1,336 @@
...
+        .arg(dasd)
+        .output()
+        .chain_err(|| "execututing lszdev")?;
+    let s = cmd
+        .stdout
+        .iter()
+        .map(|&c| c as char)
+        .filter(|&c| c != '\n' && c != '\r')
+        .collect::<String>();
+    eprintln!("Bus ID: {}", s);
```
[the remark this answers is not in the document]
nikita-dubrovskii: same as above

src/s390x/dasd.rs
```diff
@@ -0,0 +1,336 @@
...
+
+/// Copy each partition from CoreOS raw image to corresponding /dev/<disk-name>
+///
+/// # Arguments
+/// * `dds` - list of dd options.
+fn copy_partitions(dds: &[Opts]) -> Result<()> {
+    for opts in dds {
+        match dd(&opts) {
+            Ok(s) => {
+                eprintln!("{}", s);
```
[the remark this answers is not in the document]
nikita-dubrovskii: same as above

src/s390x/dasd.rs
```diff
@@ -0,0 +1,336 @@
...
+    ctx.assign_device(&image, true)
+        .chain_err(|| format!("opening image {}", image.display()))?;
+
+    let mut dd_options = Vec::new();
+    let mut partitions = Vec::new();
+    let mut start = 2; //the first 2 tracks of the ECKD DASD are reserved;
+    let mut table = ctx.get_partitions().chain_err(|| "reading partitions")?;
+    let mut count = table.nents();
+
+    for mut pt in &mut table {
```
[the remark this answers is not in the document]
nikita-dubrovskii: fixed

src/s390x/dasd.rs
```diff
@@ -0,0 +1,336 @@
...
+        let uuid = pt.uuid().unwrap_or_else(|_| "UUID unknown".to_string());
```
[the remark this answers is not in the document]
nikita-dubrovskii: done

src/s390x/dasd.rs
```diff
@@ -0,0 +1,336 @@
...
+///
+/// # Arguments
+/// * `dasd` - dasd device, i.e. smth like /dev/dasda
+/// * `file` - configuration file
+fn try_format<P: AsRef<Path>>(dasd: &str, file: P) -> Result<()> {
+    eprintln!("Formatting {} ", dasd);
+    let status = Command::new("fdasd")
+        .env("LC_ALL", "C")
+        .arg("--config")
+        .arg(file.as_ref().to_str().unwrap())
```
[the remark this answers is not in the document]
nikita-dubrovskii: done

src/s390x/dasd.rs
nikita-dubrovskii: done

src/s390x/zipl.rs
nikita-dubrovskii: done

bgilbert:
I've pushed some WIP cleanup commits.  I'm planning to do a little more cleanup work, and this will need rebasing on #285, #290, and #291.  Converted the PR to draft for now.

@nikita-dubrovskii, I've tested this in Hercules, but PTAL and let me know if I've introduced any issues.

nikita-dubrovskii:
@bgilbert , i've tested and it's not working:
```
[    7.952529] coreos-installer-service[874]: Error: copying partition 
[    7.952784] coreos-installer-service[874]: Caused by: Invalid character '    ' at position 18374690876376778512 
```

now debugging

src/s390x/zipl.rs
```diff
@@ -58,14 +59,18 @@ pub fn install_bootloader<P: AsRef<Path>>(
...
+        bail!("couldn't install bootloader on {}", disk);
+    }
 
     let status = Command::new("chreipl")
         .arg(disk)
         .status()
-        .chain_err(|| "failed to execute chreipl")?;
-    error_chain::ensure!(status.success(), "setting {} as boot-device", disk);
+        .chain_err(|| format!("failed to execute chreipl on {}", disk))?;
+    if !status.success() {
```
nikita-dubrovskii, line 77: is there any reason to use `if - bail` instead of `ensure` ?

bgilbert: IMO it's better to make control flow explicit, rather than hiding it in a macro argument.

bgilbert:
I think this is ready for review!

bgilbert:
@jlebon proposed squashing this series down to one commit before merging.  Any preferences?

nikita-dubrovskii:
> @jlebon proposed squashing this series down to one commit before merging. Any preferences?

hmm, it's too big for one commit. maybe we can have smth like:

1. s390x: add support for IBM Z DASD disks
2. clean-ups
3. s390x: use ioctls to get DASD parameters 
4. s390x: switch from fdisk to gptman
5. s390x: drop dd-lib
6. s390x: stream download directly to disk 
7. clean-ups
8. s390x: trigger udev_settle after disk format
9. s390x: remove variable substitutions from kargs
10.  s390x: use '/dev/dasd' as a pattern

bgilbert:
I'll let @jlebon chime in here, but I think his idea is to avoid leaking development history into the master branch.  The merged commit would list you as the primary author, and me as a [co-author](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line).

I think the prep changes that affect the rest of the codebase (e.g. adding a callback to `write_image`) should still be separate commits, though.

src/s390x/zipl.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// Generate zipl config and set boot device
+///
+/// # Arguments
+/// * `boot` - path to boot partition mountpoint, i.e. smth like /boot
+/// * `disk` - smth like /dev/dasda
+pub fn install_bootloader<P: AsRef<Path>>(boot: P, disk: &str) -> Result<()> {
+    eprintln!("Installing bootloader");
+
+    let bootconf = boot_config(&boot)?;
```
jlebon, line 36: `get_bls_config_path` ?

src/s390x/zipl.rs
```diff
@@ -0,0 +1,130 @@
...
+    let vmlinuz = kernel(&boot)?;
```
jlebon, line 37: `get_kernel_path`?

src/s390x/zipl.rs
```diff
@@ -0,0 +1,130 @@
...
+    let ramfs = initramfs(&boot)?;
```
jlebon, line 38: `get_initramfs_path`?

src/s390x/zipl.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+fn initramfs<P: AsRef<Path>>(boot: P) -> Result<PathBuf> {
+    find_file(boot.as_ref().join("ostree"), "initram")
+}
+
+fn get_boot_kargs<P: AsRef<Path>>(boot_config: P) -> Result<String> {
+    let contents = read_to_string(&boot_config)
+        .chain_err(|| format!("reading {}", boot_config.as_ref().display()))?;
+    // read kargs from options line
```
jlebon, line 112: Regex is fine, though it'd be cleaner and probably less lines overall to just iterate over the lines and find the one that starts with `options`. See e.g. `bls_entry_delete_and_append_kargs`.

jlebon:
> I'll let @jlebon chime in here, but I think his idea is to avoid leaking development history into the master branch. The merged commit would list you as the primary author, and me as a co-author.
>
> I think the prep changes that affect the rest of the codebase (e.g. adding a callback to write_image) should still be separate commits, though.

Yeah exactly. I think the rule of thumb I personally use is something like: any commit that's just changing content that's all new in the same PR should normally just be squashed in. To take a random example, is [this commit](https://github.com/coreos/coreos-installer/pull/153/commits/c4853c6d128fc77933e2aa53c32af05d995008a5) valuable to have on its own on the master branch? It's just modifying a `README.md` file which master hasn't even seen yet.

Commits which refactor existing code as prep should definitely still be kept.

src/s390x/zipl.rs
bgilbert: Eh, I find this cleaner than open-coding an iteration and `starts_with()`.  Can change it if there's a strong preference though.

jlebon: Not a strong preference. Regex just seemed like overkill for the task at hand, but meh...

scripts/coreos-installer-service
```diff
@@ -10,6 +10,9 @@ PERSIST_KERNEL_NET_PARAMS=("ipv6.disable" "net.ifnames" "net.naming-scheme")
 # List from https://www.mankier.com/7/dracut.cmdline#Description-Network
 PERSIST_DRACUT_NET_PARAMS=("ip" "ifname" "rd.route" "bootdev" "BOOTIF" "rd.bootif" "nameserver" "rd.peerdns" "biosdevname" "vlan" "bond" "team" "bridge" "rd.net.timeout.carrier" "coreos.no_persist_ip")
 
+# IBM S390X params to persist
+PERSIST_S390X_PARAMS=("rd.dasd" "rd.zfcp" "rd.znet" "zfcp.allow_lun_scan" "cio_ignore")
```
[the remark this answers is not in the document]
bgilbert: As I understand it:
- `rd.dasd` activates a DASD device, without which a DASD boot device won't show up
- `rd.zfcp` activates a Fibre Channel device, without which an FC boot device won't show up
- `rd.znet` activates a network device, without which it won't be available from the initramfs
- `zfcp.allow_lun_scan` might be needed to configure activation of certain FC devices
- `cio_ignore` tells the kernel not to attach to specified devices on boot

So I suspect they might all be needed.  @nikita-dubrovskii, is that correct?  Any additional details?

src/install.rs
```diff
@@ -99,6 +101,11 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         }
     }
 
+    if is_dasd(config)? {
+        #[cfg(target_arch = "s390x")]
+        s390x::prepare_dasd(&config)?;
+    }
```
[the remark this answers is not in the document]
bgilbert: That requires Rust &ge; 1.43.0 (2020-04-23).

bgilbert:
@nikita-dubrovskii If you agree, I'll refactor into three commits:
- The generic parts of "s390x: stream download directly to disk"
- The generic parts of "install: correctly detect DASDs if they're behind a symlink"
- Everything else, with `Author: you` and `Co-authored-by: me`

I can do that in a new PR if you'd like the history to be preserved here.

scripts/coreos-installer-service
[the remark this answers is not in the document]
nikita-dubrovskii: >So I suspect they might all be needed.

Exactly, we have to persist those kargs.

src/s390x/zipl.rs
nikita-dubrovskii: As i remember, some of you asked me to use regex for substring-search operations )

nikita-dubrovskii:
> I can do that in a new PR if you'd like the history to be preserved here.

I don't mind, for me any approach is Ok.

src/s390x/zipl.rs
bgilbert: Yup, that was me.  :slightly_smiling_face:

bgilbert:
Updated and squashed!

PR #155, opened 2020-02-12
blockdev: specifically wait for /boot to show up

jlebon:
We want to be resilient to systems where the kernel may take a while
to send the updates out, or udevd hasn't reacted to them yet, or there's
some other IO stuttering.

So we keep running `udevadm settle` until the boot partition shows up.
Note we're not actually waiting a long time in the loop itself in total
(just 1s). The real waiting happens at `udevadm settle` (which doesn't
really _do_ anything else than wait for udev rules to be processed). The
loop here is just to have a better chance to have `udevadm settle` watch
for the actual udev events we care about (see the related comment about
this that I deleted).

See also: https://github.com/coreos/fedora-coreos-tracker/issues/385

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -225,6 +225,22 @@ pub fn try_discard_all(file: &mut File) -> Result<bool> {
 ioctl_none!(blkrrpart, 0x12, 95);
 ioctl_write_ptr_bad!(blkdiscard, request_code_none!(0x12, 119), [u64; 2]);
 
+pub fn udev_settle_until_partition_appears(device: &str, label: &str) -> Result<()> {
+    eprint!("Waiting for partition label {} to come online", label);
+    for _ in (0..10).rev() {
+        udev_settle()?;
+
+        let dev = get_partition_with_label(device, label)?;
+        if dev.is_some() {
+            return Ok(());
+        }
+
+        sleep(Duration::from_millis(100));
+    }
+
+    return Err("timed out waiting for boot partition to show up".into());
+}
+
 pub fn udev_settle() -> Result<()> {
     // "udevadm settle" silently no-ops if the udev socket is missing, and
     // then lsblk can't find partition labels.  Catch this early.
@@ -233,12 +249,6 @@ pub fn udev_settle() -> Result<()> {
             "udevd socket missing; are we running in a container without /run/udev mounted?".into(),
         );
     }
-
-    // There's a potential window after rereading the partition table where
-    // udevd hasn't yet received updates from the kernel, settle will return
-    // immediately, and lsblk won't pick up partition labels.  Try to sleep
-    // our way out of this.
-    sleep(Duration::from_millis(200));
 
     let status = Command::new("udevadm")
         .arg("settle")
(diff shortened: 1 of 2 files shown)


jlebon:
(Haven't tested this yet...)

src/blockdev.rs
```diff
@@ -225,6 +225,22 @@ pub fn try_discard_all(file: &mut File) -> Result<bool> {
...
+
+        let dev = get_partition_with_label(device, label)?;
+        if dev.is_some() {
+            return Ok(());
+        }
+
+        sleep(Duration::from_millis(100));
+    }
+
+    return Err("timed out waiting for boot partition to show up".into());
```
cgwalters: We're waiting a total of 1s?  Maybe add that to the error; something like
`
let maxwait = Duration::from_secs(1);
for dur in (0..maxwait/10) {
   ...
   sleep(dur)
}
return Err("timed out waiting {} for partition label '{}' of device {}", maxwait, label, device);
```

File: src/blockdev.rs [hunk 2]
```diff
@@ -225,6 +225,22 @@ pub fn try_discard_all(file: &mut File) -> Result<bool> {
 ioctl_none!(blkrrpart, 0x12, 95);
 ioctl_write_ptr_bad!(blkdiscard, request_code_none!(0x12, 119), [u64; 2]);
 
+pub fn udev_settle_until_partition_appears(device: &str, label: &str) -> Result<()> {
```
cgwalters: Preserve the comment that was deleted here?

src/blockdev.rs
```diff
@@ -225,6 +225,22 @@ pub fn try_discard_all(file: &mut File) -> Result<bool> {
...
+    eprint!("Waiting for partition label {} to come online", label);
```
bgilbert: If we want to report this to the user (and I'm not 100% on that) it should be in `write_disk()` (and an `eprintln!`) rather than in utility code.

Probably quotes around the `{}`, too.

src/install.rs
```diff
@@ -104,10 +104,10 @@ fn write_disk(
     // copy the image
     write_image(source, dest, true)?;
     reread_partition_table(dest)?;
-    udev_settle()?;
 
     // postprocess
     if ignition.is_some() || config.firstboot_kargs.is_some() || config.platform.is_some() {
+        udev_settle_until_partition_appears(&config.device, "boot")?;
```
bgilbert: Let's not hardcode the partition label both inline in `blockdev.rs` and inline here.

src/blockdev.rs
```diff
@@ -225,6 +225,22 @@ pub fn try_discard_all(file: &mut File) -> Result<bool> {
...
+    for _ in (0..10).rev() {
```
bgilbert: If we're going to do this, we might as well increase the number of iterations.  Timeout leads to global failure anyway, and most systems should return on the first iteration.

jlebon:
I think this may still be worth doing, but I'd rather wait until we actually see this race in the wild. Let's close for now and we can revive then.

PR #160, opened 2020-02-13
source: add an upper bound timeout to HTTP requests

lucab:
This adds a very large timeout (4 hours) to all HTTP requests, as they
currently default to none.
It is meant to almost never trigger, except under pathological network
conditions. In such cases, it should help detecting issues by explicitly
erroring out instead of hanging forever.

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -20,10 +20,15 @@ use std::fmt::{Display, Formatter};
 use std::fs::OpenOptions;
 use std::io::{Read, Seek, SeekFrom};
 use std::path::Path;
+use std::time::Duration;
 
 use crate::cmdline::*;
 use crate::errors::*;
 
+/// Completion timeout for HTTP requests (4 hours).
+const HTTP_COMPLETION_TIMEOUT: Duration = Duration::from_secs(4 * 60 * 60);
+
+/// Default base URL to Fedora CoreOS streams metadata.
 const DEFAULT_STREAM_BASE_URL: &str = "https://builds.coreos.fedoraproject.org/streams/";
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Does this apply even in the face of retrying requests?  (Do we retry requests?)

lucab:
This is a completion timeout on each request. In case of N retries, the overall waiting time would be `N * timeout` in the worst case.

Additionally, I think we aren't currently retrying HTTP downloads. I guess this is in part because we directly stream the response/image to disk. I'm unsure whether there should be a top-level retry logic with a `--retry` flag, or if we should simply leave that to a looping caller.

/cc @bgilbert

src/source.rs
```diff
@@ -20,10 +20,15 @@ use std::fmt::{Display, Formatter};
 use std::fs::OpenOptions;
 use std::io::{Read, Seek, SeekFrom};
 use std::path::Path;
+use std::time::Duration;
 
 use crate::cmdline::*;
 use crate::errors::*;
 
+/// Default timeout for HTTP requests (4 hours).
+const DEFAULT_HTTP_REQ_TIMEOUT: Duration = Duration::from_secs(4 * 60 * 60);
```
bgilbert: Let's not call this the "default" since it's not configurable.

lucab: Ack, I've renamed to `HTTP_COMPLETION_TIMEOUT`. Were you hinting at keeping a ticket around to maybe later add a CLI flag to configure this, or do you think it isn't useful to have that?

bgilbert: Nah, I don't think it'd be useful.  We can always add later if needed.

Issue #161, opened 2020-02-13
Consider checking that disk sector size matches image


bgilbert:
Rather than failing later with an obscure error, maybe we should check that the disk sector size matches the partition table and fail if not.

PR #167, opened 2020-02-22
install: Fail early if image sector size doesn't match destination

bgilbert:
Otherwise we'd copy the whole image, then complain that we couldn't find any partitions.  Detect the image sector size by checking both possible offsets for a GPT, detect the destination sector size by `ioctl`, and fail if they don't match.

If we don't detect any GPT, continue without error.  In the future, detection could be expanded to cover additional partition table types.

In addition, if no partitions are detected after writing the image, complain about it rather than saying `couldn't find boot device for /dev/sda`.  That should make it easier to track down non-sector-size issues with similar symptoms.

Closes https://github.com/coreos/coreos-installer/issues/161.


src/blockdev.rs
```diff
@@ -195,6 +208,18 @@ pub fn reread_partition_table(file: &mut File) -> Result<()> {
...
 }
 
+/// Get the logical sector size of a block device.
+pub fn get_sector_size(file: &mut File) -> Result<u32> {
+    let fd = file.as_raw_fd();
+    let mut size: c_int = 0;
+    match unsafe { blksszget(fd, &mut size) } {
+        Ok(_) => Ok(size
+            .try_into()
+            .chain_err(|| "sector size doesn't fit in u32")?),
```
lucab: I don't expect this case to happen except for very weird cases, where I guess it could be helpful to display `size` somewhere in the error.

src/blockdev.rs
```diff
@@ -250,6 +276,51 @@ pub fn udev_settle() -> Result<()> {
...
 }
 
+/// A Reader of a disk image that fails if the formatted sector size of the
+/// image doesn't match a specified value.
+pub struct DiskSectorSizeReader<R: Read> {
+    source: BufReader<R>,
+}
+
+impl<R: Read> DiskSectorSizeReader<R> {
+    pub fn new(source: R, expected_size: u32) -> Result<Self> {
```
lucab: If it doesn't bother you too much, I'd add a quick regression-test covering the 512/4096/garbage cases.

src/blockdev.rs
```diff
@@ -14,35 +14,57 @@
...
+    }
+
+    // find the boot partition
+    let boot_partitions = partitions
+        .iter()
+        .filter(|d| d.label.as_ref().unwrap_or(&"".to_string()) == "boot")
+        .collect::<Vec<&BlkDev>>();
+    if boot_partitions.is_empty() {
+        bail!("couldn't find boot device for {}", device);
+    } else if boot_partitions.len() > 1 {
```
lucab: Self-note: `itertools` has a few helper methods to collapse a collection to a single-matching-item, but none of them with ergonomic empty-vs-many logic (and we don't currently bring in that crate, not even transitively).

src/blockdev.rs
```diff
@@ -250,6 +276,51 @@ pub fn udev_settle() -> Result<()> {
...
 
+/// A Reader of a disk image that fails if the formatted sector size of the
+/// image doesn't match a specified value.
+pub struct DiskSectorSizeReader<R: Read> {
+    source: BufReader<R>,
+}
+
+impl<R: Read> DiskSectorSizeReader<R> {
+    pub fn new(source: R, expected_size: u32) -> Result<Self> {
+        let gpt_magic: &[u8; 12] = b"EFI PART\x00\x00\x01\x00";
```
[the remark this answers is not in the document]
bgilbert: The idea was that we may not understand the semantics of future versions and shouldn't make assumptions about them.  I don't feel strongly about it, and can drop the version part if you want.

bgilbert:
Updated.

src/blockdev.rs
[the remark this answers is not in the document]
lucab: I'm fine with sticking to your approach and keeping it, I was mostly wondering whether it was an intentional decision.

lucab:
CI is red due to a clippy warning. Other than that, PR looks fine and good to go.

src/blockdev.rs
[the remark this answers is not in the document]
bgilbert: On second thought, we're getting the information we need from the location of the GPT, not its contents, so we probably don't care so much about the details of the latter.  I'll stick to just the signature.

bgilbert:
Updated for :paperclip:.

src/blockdev.rs
```diff
@@ -260,6 +280,51 @@ pub fn udev_settle() -> Result<()> {
     Ok(())
 }
 
+/// A Reader of a disk image that fails if the formatted sector size of the
+/// image doesn't match a specified value.
+pub struct DiskSectorSizeReader<R: Read> {
+    source: BufReader<R>,
+}
+
+impl<R: Read> DiskSectorSizeReader<R> {
```
jlebon: We're already reading and keeping the first 1M of the source image, right? Couldn't we just check that buffer? Doing it in a wrapping reader is fine, though it seems like unnecessary cognitive load/architecturing.

bgilbert: Good call.  That's much, much better.

bgilbert:
Updated.

src/download.rs
```diff
@@ -239,15 +247,32 @@ pub fn write_image(source: &mut ImageSource, dest: &mut File, decompress: bool)
...
     let mut first_mb: [u8; 1024 * 1024] = [0; 1024 * 1024];
-    dest.write_all(&first_mb)
-        .chain_err(|| "clearing first MiB of disk")?;
     decompress_reader
         .read_exact(&mut first_mb)
         .chain_err(|| "decoding first MiB of image")?;
+    // Were we asked to check sector size?
+    if let Some(expected) = expected_sector_size {
+        // Can we derive one from source data?
+        if let Some(actual) = detect_formatted_sector_size(&first_mb) {
```
jlebon: If we're asked to expect a sector size, shouldn't we fail if we can't query the sector size from the image?

bgilbert: It's a best-effort check to improve UX.  We don't currently detect sector size on non-EFI platforms.

bgilbert:
Thanks all!

Issue #165, opened 2020-02-13
couldn't find boot device for /dev/sda


ssoor:
# Bug
## Host Operating System Version
Microsoft Windows [version 10.0.18362.592]

VM: Hyper-V

## Target Operating System Version
fedora-coreos-31.20200210.2.0-live.x86_64.iso

## coreos-installer Version
coreos-installer 0.1.2

Expected Behavior
boot drive part partitioned

## Actual Behavior
```
Clearing partition table
Error: install failed
```
## Reproduction Steps
```
# /dev/sda == os.vhd(1024G)
# bootiso == fedora-coreos-31.20200127.3.0-live.x86_64.iso
[core@localhost ~]$ sudo coreos-installer install --image-url=http://172.18.52.212/_iso/fedora-coreos-31.20200210.2.0-live.x86_64.iso --ignition=http://172.18.52.212/config.ign /dev/sda
Downloading image from http://172.18.52.212/_iso/fedora-coreos-31.20200210.2.0-live.x86_64.iso
Downloading signature from http://172.18.52.212/_iso/fedora-coreos-31.20200210.2.0-live.x86_64.iso.sig
gpg: Signature made Mon Feb 10 23:21:03 2020 UTC
gpg:                using RSA key 50CB390B3C3359C4
gpg: Good signature from "Fedora (31) <fedora-31-primary@fedoraproject.org>" [ultimate]
> Read disk 634.0 MiB/634.0 MiB (100%)
Error: couldn't find boot device for /dev/sda
Clearing partition table
Error: install failed
[core@localhost ~]$ sudo fdisk -l /dev/sda
Disk /dev/sda: 1 TiB, 1099511627776 bytes, 2147483648 sectors
Disk model: Virtual Disk
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
```

```
# /dev/sda == os.vhdx(127G)
# bootiso == fedora-coreos-31.20200210.2.0-live.x86_64.iso
[root@localhost core]# coreos-installer install /dev/sda --image-url=http://172.18.52.212/_iso/fedora-coreos-31.20200210.2.0-live.x86_64.iso --ignition=http://172.18.52.212/config.ign
Downloading image from http://172.18.52.212/_iso/fedora-coreos-31.20200210.2.0-live.x86_64.iso
[...]

PR #168, opened 2020-02-22
install: Add option to skip clearing partition table on error

bgilbert:
If `--preserve-on-error` is specified, don't clear the destination partition table on error, as a debugging aid.

For https://github.com/coreos/coreos-installer/issues/165#issuecomment-588380848.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -35,6 +35,7 @@ pub struct InstallConfig {
     pub platform: Option<String>,
     pub firstboot_kargs: Option<String>,
     pub insecure: bool,
+    pub preserve_on_error: bool,
 }
 
 pub struct DownloadConfig {
@@ -151,6 +152,11 @@ pub fn parse_args() -> Result<Config> {
                         .help("Target CPU architecture")
                         .default_value(uname.machine())
                         .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("preserve-on-error")
+                        .long("preserve-on-error")
+                        .help("Don't clear partition table on error"),
                 )
                 // positional args
                 .arg(
@@ -379,6 +385,7 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         platform: matches.value_of("platform").map(String::from),
         firstboot_kargs: matches.value_of("firstboot-kargs").map(String::from),
         insecure: matches.is_present("insecure"),
+        preserve_on_error: matches.is_present("preserve-on-error"),
     }))
 }
 
--- a/src/install.rs
+++ b/src/install.rs
@@ -78,7 +78,11 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         eprint!("{}", ChainedError::display_chain(&err));
 
         // clean up
-        clear_partition_table(&mut dest)?;
+        if config.preserve_on_error {
+            eprintln!("Preserving partition table as requested");
+        } else {
+            clear_partition_table(&mut dest)?;
+        }
 
         // return a generic error so our exit status is right
         bail!("install failed");


src/cmdline.rs
```diff
@@ -152,6 +153,11 @@ pub fn parse_args() -> Result<Config> {
                         .default_value(uname.machine())
                         .takes_value(true),
                 )
+                .arg(
+                    Arg::with_name("preserve-on-error")
+                        .long("preserve-on-error")
```
lucab: Beware, bike-shedding ahead.

If this is mostly (only?) meant for partition-table debugging, should we maybe:
 * mention `table` or `part` somewhere in the visible name
 * mark as `.hidden(true)` XOR move to an undocumented env-flag

bgilbert: I think the option is useful for debugging generally, e.g. if there's a failure writing to something in the `/boot` partition.  I thought about mentioning the partition table, but the fact that we clear it (rather than the whole disk) feels like an implementation detail we shouldn't commit to.

On balance I think there's also some value in discoverability.  Otherwise users are more likely to have to ask us for help debugging their setups.

That said, I'm not completely happy with the option name and help text.  Thoughts, preferences, and proposals welcome.

lucab: Fair. It seems like you'd want this to be generic and useful enough on its own.

I honestly don't have very good proposals. An alternative could be `--lazy-failures: Do not perform cleanups on install failures` but I'm not strongly convinced. Maybe @jlebon has something.

If we don't find something more convincing, I'm also fine ending the bikeshedding here.

jlebon: I think "cleanups" might be misunderstood to just mean temporary/cache files the installer needs during the install. I'd go with something more explicit like `--skip-wipe-on-error: Don't wipe the device on install failure`?

bgilbert: @jlebon That's basically what I started with, but:

- Re option name, it seemed better to have the name describe what it does, rather than what it doesn't do.
- Re help text, as of #172 we don't wipe the entire device on failure, and implying otherwise might scare someone who wants to preserve a data partition.  (Which undermines my point above about not mentioning the partition table...)

Thoughts?

jlebon: > Re help text, as of #172 we don't wipe the entire device on failure, and implying otherwise might scare someone who wants to preserve a data partition. (Which undermines my point above about not mentioning the partition table...)

Hmm, so maybe a prerequisite before this patch is figuring out what exactly we *should* do on failure if we don't actually want to wipe the partition table. Maybe something like remember the original 1MiB block and restore it on failure?

bgilbert: I think we're doing the right thing already.  Restoring the old partition table and boot sector implies that the old partitions are still usable, but we've probably overwritten several GB of the disk so that's not actually the case.  We'll have deleted the user's _data_ partitions, but we won't have overwritten their contents, so an Ignition config that recreates those partitions and conditionally formats their filesystems will work as expected.

In principle we could recreate only the unclobbered partitions on failure, but that seems like a lot of work.  I could file an issue though.

Thoughts?

jlebon: Hmm OK, so I think this also intersects with https://github.com/coreos/coreos-assembler/pull/924#issuecomment-589268886 actually. Because while the disk might have fit before a certain partition offset in the original version, a newer version may not. I'm wondering now if we should do something like: if we detect that the disk already has FCOS installed, and it has an additional partition, and the new disk will not fit, then we error out and require a `--force` flag to proceed. This gives users a chance to move the partition forward before having it be clobbered.

I guess we could make this more generic so as to support not just FCOS. E.g. some kind of `--preserve-partition` flag which (1) ensures that we won't clobber it, and (2) recreates it as is after a successful install?

Anyway, don't mean to derail this PR too much. Keeping it as `preserve-on-error` WFM.

cgwalters:
This looks sane to me as is.

But another option: Rather than clearing the partition table if something goes wrong, we could just neuter bootability - toggle off the MBR boot flag and change the ESP GPT type e.g.

bgilbert:
IIUC the MBR boot flag isn't relevant if GRUB is installed at the MBR level, but we could clear the stage 1 bootloader code.  This more nuanced approach is arch-specific, though.  I'm also worried that it could lead to confusion, e.g. users trying to fix bootability on an image that failed GPG verification.

cgwalters:
Right.  A bit more elaborate idea is to have coreos-installer write an error message to `/boot/coreos-installer.failed` and our systemd units detect this and print it to the console on the subsequent boot and fail.

bgilbert:
That's assuming we were successful enough to be able to do so.  Also, if GPG verification failed, the image is untrusted and we shouldn't be mounting its filesystems.

PR #177, opened 2020-02-28
source: keep going on signature-fetching errors

lucab:
This tweaks the signature fetching logic, so that it log-and-continues
on any kind of errors (e.g. timeouts or broken redirects).
This is mostly useful for the "insecure" mode, where remote images
can be consumed even without a corresponding signature.
In "normal" mode, such case will instead result in an hard-failure
when trying to validate the (missing) signature.
Additionally, this prints more error details on non-OK HTTP responses.

Ref: coreos/fedora-coreos-tracker#402

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -147,6 +147,29 @@ impl UrlLocation {
             artifact_type: artifact_type.to_string(),
         }
     }
+
+    /// Fetch signature content from URL.
+    fn fetch_signature(sig_url: &Url) -> Result<Vec<u8>> {
+        let client = new_http_client()?;
+        let mut resp = client
+            .get(sig_url.clone())
+            .send()
+            .chain_err(|| "sending signature request")?;
+        match resp.status() {
+            StatusCode::OK => {
+                let mut sig_vec = Vec::new();
+                resp.read_to_end(&mut sig_vec)
+                    .chain_err(|| "reading signature URL")?;
+                Ok(sig_vec)
+            }
+            _ => {
+                bail!(
+                    "error response from signature URL: {}",
+                    resp.error_for_status().unwrap_err()
+                );
+            }
+        }
+    }
 }
 
 impl Display for UrlLocation {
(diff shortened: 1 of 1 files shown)


src/source.rs
```diff
@@ -161,26 +184,12 @@ impl Display for UrlLocation {
...
-                    .chain_err(|| "reading signature URL")?;
-                Some(sig_vec)
-            }
-            s => {
-                eprintln!("Signature fetch failed: {}", s);
-                None
-            }
-        };
+        let signature = Self::fetch_signature(&self.sig_url)
+            .map_err(|e| eprintln!("failed to fetch signature: {}", e))
```
bgilbert: Non-fatal log messages should start with a capital letter.

lucab:
Sorry that the commit message was not easy to decipher, your reading of the patch is correct though.

Reworded along the lines of your suggestion, I hope it is less obscure now.

lucab:
@bgilbert yeah, added one more sentence to that effect. Feel free to merge when it looks sufficiently complete to you.

src/source.rs
```diff
@@ -147,6 +147,29 @@ impl UrlLocation {
...
     }
+
+    /// Fetch signature content from URL.
+    fn fetch_signature(sig_url: &Url) -> Result<Vec<u8>> {
+        let client = new_http_client()?;
+        let mut resp = client
+            .get(sig_url.clone())
+            .send()
+            .chain_err(|| "sending signature request")?;
+        match resp.status() {
```
cgwalters: Can just use [error_for_status](https://docs.rs/reqwest/0.10.1/reqwest/struct.Response.html#method.error_for_status) ?

cgwalters:
Related to this, I don't understand the code around `--insecure` today at all.  It looks to me like at https://github.com/coreos/coreos-installer/blob/master/src/download.rs#L45 we print...and then just continue?

cgwalters:
OK I think I see, this will turn into `None` for signature which looks like it will end up being handled.  But I'm testing this now too.

cgwalters:
OK yep confirmed this fixes the no-signature case.  Let's ship this and revisit anything else as followup.

src/source.rs
lucab: Yes, good point (sorry I was offline yesterday, just seen your comment).
Followup at https://github.com/coreos/coreos-installer/pull/181.

PR #182, opened 2020-03-06
ci: Add a CoreOS CI test

cgwalters:
Part of https://github.com/coreos/fedora-coreos-tracker/issues/263

This makes use of the new `kola testiso` command too.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,0 +1,15 @@
+// Documentation: https://github.com/coreos/coreos-ci/blob/master/README-upstream-ci.md
+
+cosaPod(buildroot: true) {
+    checkout scm
+
+    fcosBuild(make: true)
+
+    stage("Build metal+live") {
+        shwrap("cd /srv/fcos && cosa build metal")
+        shwrap("cd /srv/fcos && cosa buildextend-live")
+    }
+    stage("Test ISO") {
+        shwrap("cd /srv/fcos && kola testiso --cosa-build builds/latest/x86_64/meta.json -S")
+    }
+}


.cci.jenkinsfile
```diff
@@ -0,0 +1,17 @@
+// Documentation: https://github.com/coreos/coreos-ci/blob/master/README-upstream-ci.md
+
+cosaPod(buildroot: true) {
+    checkout scm
+
+    fcosBuild(make: true)
+
+    stage("Build metal") {
+        shwrap("cosa build metal")
```
jlebon: Oh right, so one thing here is that the FCOS build actually happens in `/srv/fcos` which is on an `emptyDir`. Easiest is to just wrap all the additional stages in a `dir("/srv/fcos") { ... }`.

.cci.jenkinsfile
```diff
@@ -0,0 +1,21 @@
...
+    fcosBuild(make: true)
+
+    dir("/srv/fcos") {
+        stage("Build metal/live") {
+            parallel metal: {
+                shwrap("cosa build metal")
+            },
+            live: {
+                shwrap("cosa buildextend-live")
+            }
```
jlebon: We don't do any locking right now on `meta.json` though. Worse, most code paths I think read it in on start, so it's likely the last one to finish will clobber modifications.

I've thought about supporting this in cosa though. It would make the FCOS pipeline much faster. (And related here of course is https://github.com/coreos/coreos-assembler/issues/531).

cgwalters: Ohh, right.

.cci.jenkinsfile
```diff
@@ -0,0 +1,15 @@
...
+    checkout scm
+
+    fcosBuild(make: true)
+
+    stage("Build metal+live") {
+        shwrap("cd /srv/fcos && cosa build metal")
+        shwrap("cd /srv/fcos && cosa buildextend-live")
+    }
+    stage("Test ISO") {
+        shwrap("cd /srv/fcos && kola testiso --cosa-build builds/latest/x86_64/meta.json -S")
```
jlebon: Hmm, I think we should be able to just `cosa kola testiso -S` here to let cosa pass the `--cosa-build` flag automatically.

cgwalters: Yeah you're right, though I am hoping soon to teach `kola` directly to do the right thing.

Let's get this in since it's green and working and do other cleanups/enhancements as followups?

jlebon: Yup, didn't mean it to be blocking.

PR #183, opened 2020-03-06
download: Avoid using `is_some()`

cgwalters:
I think using that API call is a bit of a code smell; it isn't
*wrong* but I think it's better to use matching like `if let Some(x)`.
Among other things it avoids an `unwrap()` call which is *definitely*
a code smell.  It's harder for the compiler to optimize out the
`unwrap()` too since it has to reason about the `is_some()` logic.

If agreed I'll change a few other places like this too.

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -147,7 +147,7 @@ fn write_image_and_sig(
     write_image(source, &mut dest, decompress, None)?;
 
     // write signature, if relevant
-    if !decompress && source.signature.is_some() {
+    if let (false, Some(signature)) = (decompress, source.signature.as_ref()) {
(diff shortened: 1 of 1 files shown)


bgilbert:
In general I agree.  In this particular case I find the new logic harder to read, but I'm willing to be overruled.

lucab:
I also agree in the general case.

On the other hand, I usually try to avoid chaining multiple `let`/`if let` as I think it's a heavier on  casual readers and lang-learners.

In this specific case, I'd probably go for a 
```
let (have_length, length_hint) = match source.length_hint { ... };
```

cgwalters:
OK I picked a different place for this.

PR #191, opened 2020-03-19
blockdev: preparatory patches for osmet

jlebon:
Split out of https://github.com/coreos/coreos-installer/pull/187. See individual commit messages.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -30,28 +30,36 @@ use tempfile;
 
 use crate::errors::*;
 
-pub fn mount_boot(device: &str) -> Result<Mount> {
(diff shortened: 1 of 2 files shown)


src/blockdev.rs
```diff
@@ -30,28 +30,36 @@ use tempfile;
 
 use crate::errors::*;
 
-pub fn mount_boot(device: &str) -> Result<Mount> {
+pub fn mount_partition_by_label(device: &str, label: &str, flags: mount::MsFlags) -> Result<Mount> {
```
bgilbert: This requires the caller to know about `nix::mount`.  That seems like a reasonable tradeoff, but worth noting.

PR #194, opened 2020-03-23
README.md: Reorder, and note RHCOS has legacy

cgwalters:
Reorder the "how to get/use" instructions to demote "via Fedora RPM"
to below "via container".  And also the "on other editions" verbiage
was wrong, because "other editions" include e.g. Silverblue and IoT
which don't use `dnf` for the root filesystem either.

While we're here, note that `dnf install coreos-installer`
*also* works in a `--privileged` container - don't need to mutate
your host's rootfs.

Move "install with cargo" above "build from source" too.

Finally, add a note that older RHCOS uses the `legacy` branch
so people are hopefully less confused why it's different.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,19 +23,47 @@ coreos-installer.
 
 # On Fedora CoreOS or RHEL CoreOS
 
-coreos-installer is included in Fedora CoreOS and RHEL CoreOS.  Just run
+coreos-installer is included in Fedora CoreOS.  Just run
 `coreos-installer` from the command line.  Fedora CoreOS provides
 [live CD and network boot images](https://getfedora.org/coreos/download/)
 you can run from RAM; you can use these to run coreos-installer to install
 Fedora CoreOS or RHEL CoreOS to disk.
 
-# On other Fedora editions
+RHEL CoreOS also includes this, but note that versions &le; 4.4 contain
+the `legacy` branch of this software.
 
-On other Fedora editions, you can install coreos-installer with DNF in the
-usual way:
+# Run from a container
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,37 +23,38 @@ coreos-installer.
...
 
-coreos-installer is included in Fedora CoreOS and RHEL CoreOS.  Just run
+coreos-installer is included in Fedora CoreOS.  Just run
 `coreos-installer` from the command line.  Fedora CoreOS provides
 [live CD and network boot images](https://getfedora.org/coreos/download/)
 you can run from RAM; you can use these to run coreos-installer to install
 Fedora CoreOS or RHEL CoreOS to disk.
 
-# On other Fedora editions
+RHEL CoreOS also includes this, but note that versions &lt;= 4.4 contain
```
bgilbert: Might as well use `&le;` :slightly_smiling_face:

README.md
```diff
@@ -23,37 +23,38 @@ coreos-installer.
...
-To install the binary and systemd units to a target rootfs
-(e.g. under a
-[coreos-assembler](https://github.com/coreos/coreos-assembler)
-workdir):
-
-```sh
-make install DESTDIR=/my/dest/dir
-```
+(Note you can do this inside a `--privileged` container, not necessarily
+ on a host system's root filesystem)
```
bgilbert: Could we be clearer here?  For people used to traditional Fedora Workstation/Server, this statement might be confusing.

cgwalters: I tweaked this a bit but ultimately I think we want to capture more of this at https://github.com/containers/toolbox  (once it's not completely broken in the case of running privileged)

PR #197, opened 2020-03-26
install: default to offline installs via osmet

jlebon:
Change the default `install` behaviour to always install from the osmet
files if present. In live CoreOS systems that should always be the case.
In other contexts, like containers, we automatically default back to
online installs as before.

Patch best viewed with whitespaces ignored.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -517,31 +517,44 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
             None => bail!("cannot perform offline install; metadata missing"),
         }
     } else {
-        let base_url = if let Some(stream_base_url) = matches.value_of("stream-base-url") {
-            Some(Url::parse(stream_base_url).chain_err(|| "parsing stream base URL")?)
-        } else {
+        // For now, using --stream automatically will cause a download. In the future, we could
+        // opportunistically use osmet if the version and stream match an osmet file/the live ISO.
+
+        let maybe_osmet = if matches.is_present("stream") {
             None
+        } else {
+            OsmetLocation::new(architecture, sector_size)?
         };
-        let format = match sector_size {
-            4096 => "4k.raw.xz",
-            512 => "raw.xz",
-            n => {
-                // could bail on non-512, but let's be optimistic and just warn but try the regular
-                // 512b image
-                eprintln!(
-                    "Found non-standard sector size {} for {}, assuming 512b-compatible",
-                    n, &device
-                );
-                "raw.xz"
-            }
-        };
-        Box::new(StreamLocation::new(
-            matches.value_of("stream").unwrap_or("stable"),
-            architecture,
-            "metal",
-            format,
-            base_url.as_ref(),
-        )?)
+
+        if let Some(osmet) = maybe_osmet {
+            Box::new(osmet)
+        } else {
+            let format = match sector_size {
+                4096 => "4k.raw.xz",
+                512 => "raw.xz",
+                n => {
+                    // could bail on non-512, but let's be optimistic and just warn but try the regular
+                    // 512b image
+                    eprintln!(
+                        "Found non-standard sector size {} for {}, assuming 512b-compatible",
+                        n, &device
+                    );
+                    "raw.xz"
+                }
+            };
+            let base_url = if let Some(stream_base_url) = matches.value_of("stream-base-url") {
+                Some(Url::parse(stream_base_url).chain_err(|| "parsing stream base URL")?)
+            } else {
+                None
+            };
+            Box::new(StreamLocation::new(
+                matches.value_of("stream").unwrap_or("stable"),
+                architecture,
+                "metal",
+                format,
+                base_url.as_ref(),
+            )?)
+        }
     };
     // and report it to the user
     eprintln!("{}", location);


jlebon:
Works on top of https://github.com/coreos/coreos-installer/pull/187, but also requires https://github.com/coreos/coreos-assembler/pull/1244. See https://github.com/coreos/coreos-installer/pull/187#issuecomment-604560787.

src/source.rs
```diff
@@ -19,21 +19,37 @@ use std::collections::HashMap;
...
 const DEFAULT_STREAM_BASE_URL: &str = "https://builds.coreos.fedoraproject.org/streams/";
 
+/// Marker file for whether we're in a live OSTree system.
+const LIVE_SYSTEM_MARKER: &str = "/run/ostree-live";
+
+/// Marker file for whether we're in a CoreOS system.
+const COREOS_SYSTEM_MARKER: &str = "/sysroot/.coreos-aleph-version.json";
+
+/// Directory in which we look for osmet files.
+const OSMET_FILES_DIR: &str = "/run/coreos-installer/osmet";
```
bgilbert, line 43: `/run` implies that some external process needs to put them there.  Why not `/usr/share`?

src/source.rs
```diff
@@ -282,6 +303,53 @@ impl ImageLocation for StreamLocation {
...
+            // we really always expect it to be there. We could just log and fallback... though in
+            // an automated context, suddenly pulling down lots of data should raise some alarm
+            // bells.
+            if Path::new(LIVE_SYSTEM_MARKER).exists() && Path::new(COREOS_SYSTEM_MARKER).exists() {
+                bail!("{} not found in live CoreOS system; please report this as a bug");
+            }
+            return Ok(None);
+        }
+
+        let sector_size = get_sector_size_for_path(Path::new(device))?;
```
bgilbert, line 321: chain_err

src/osmet/mod.rs
```diff
@@ -150,6 +151,23 @@ pub fn osmet_unpack(config: &OsmetUnpackConfig) -> Result<()> {
     Ok(())
 }
 
+pub fn find_matching_osmet_in_dir(sector_size: NonZeroU32, osmet_dir: &Path) -> Result<PathBuf> {
```
bgilbert, line 154: Per https://github.com/coreos/coreos-installer/pull/187#discussion_r402681315 we need to care about target CPU arch also.

I'd put `osmet_dir` first.  We'll probably have more matching parameters over time.

src/osmet/mod.rs
```diff
@@ -150,6 +151,23 @@ pub fn osmet_unpack(config: &OsmetUnpackConfig) -> Result<()> {
...
+            continue;
+        }
+
+        let (osmet, _) = osmet_file_read(entry.path())?;
+        if osmet.sector_size == sector_size.get() {
+            return Ok(entry.into_path());
+        }
+    }
+
+    Err(format!("no matching osmet file found in {:?}", osmet_dir).into())
```
bgilbert, line 168: Is this really a failure?  I'd think we'd return `None` and fall back to streaming in this case.

src/osmet/mod.rs
```diff
@@ -150,6 +151,23 @@ pub fn osmet_unpack(config: &OsmetUnpackConfig) -> Result<()> {
     Ok(())
 }
 
+pub fn find_matching_osmet_in_dir(sector_size: NonZeroU32, osmet_dir: &Path) -> Result<PathBuf> {
+    for entry in WalkDir::new(osmet_dir) {
```
bgilbert, line 155: Are we intentionally recursing into subdirs here?  We might not want to commit to that.  (The directory format isn't really a public API, so maybe this doesn't matter.)

src/osmet/mod.rs
```diff
@@ -150,6 +151,23 @@ pub fn osmet_unpack(config: &OsmetUnpackConfig) -> Result<()> {
...
+        let entry = entry?;
```
bgilbert, line 156: chain_err

src/source.rs
```diff
@@ -282,6 +303,53 @@ impl ImageLocation for StreamLocation {
...
+        let osmet_path = find_matching_osmet_in_dir(sector_size, osmet_dir)?;
+        Ok(Some(Self { osmet_path }))
+    }
+}
+
+impl Display for OsmetLocation {
+    fn fmt(&self, f: &mut Formatter<'_>) -> ::std::fmt::Result {
+        write!(f, "Installing matching version in offline mode")
```
bgilbert, line 329: Let's print the image properties (arch and sector size) or filename.

src/osmet/unpacker.rs
```diff
@@ -52,8 +62,13 @@ impl OsmetUnpacker {
         Ok(Self {
             thread_handle,
             reader,
+            length,
         })
     }
+
+    pub fn length_hint(&self) -> u64 {
```
bgilbert, line 69: This returns a guaranteed size, so I'd be inclined to drop the `_hint`.

src/source.rs
```diff
@@ -282,6 +303,53 @@ impl ImageLocation for StreamLocation {
...
+        Ok(vec![ImageSource {
+            reader: Box::new(unpacker),
+            length_hint: Some(length_hint),
+            signature: None,
+            filename: "image.raw".to_string(),
+            artifact_type: "disk".to_string(),
+        }])
+    }
+
+    // For osmet, we don't require GPG verification since it comes from a trusted source.
```
bgilbert, line 347: `Don't require GPG verification, since we trust images in the OSMET_FILES_DIR.`

yrobla:
Will it be useful for our use case? We need to install RHCOS from an all-in-one USB. So in RHCOS right now, there is an small coreos-installer image, and a RHCOS raw.gz image. We can create an USB from coreos-install, and then we fill some entries such as the image-url and ignition-url. But this is forcing us to use http, and in our situation that is not possible for the image, due to flaky connection.
With that improvement, could it be possible to supply image_url as a local file, placed in the usb? I tried modifying the iso, placing rhcos.raw.gz inside images folder of coreos-installer iso, and then passing coreos.inst.image_url=file:///images/rhcos.raw.gz, but this doesn't seem to work.

src/source.rs
jlebon: We copy it into `/run` from the initrd. It can't be part of `/usr` since that'd require making it part of the metal image we've just packed. :)  See https://github.com/coreos/fedora-coreos-config/pull/322.

src/osmet/mod.rs
jlebon: Yeah, changed to return an `Option` instead. Relatedly, in #187, I've added a new `--offline` switch so that (1) we can test the offline path without first defaulting to it as done in this PR, and (2) one can assert that the install will be offline and error out otherwise. (Kinda like cargo's `--frozen` tag).

jlebon:
> Will it be useful for our use case?

The short answer is "yes". :)

> With that improvement, could it be possible to supply image_url as a local file, placed in the usb?

With this improvement, it won't be necessary to pass an image URL at all. The app will know to install the version of RHCOS which matches the RHCOS version of the live ISO you've put on the USB stick.

jlebon:
Updated! As mentioned in https://github.com/coreos/coreos-installer/pull/187#issuecomment-616661610, only the key part is kept here (the defaulting to osmet by default). Everything else is now in #187.

yrobla:
How will i inject the RHCOS image in the usb? Currently RHCOS has the coreos-installer and the .raw.gz image in different files. I am still unsure on how can i integrate both coreos-installer and the rw.gz in the same usb. Shall i need to just extract the .raw.gz and burn an iso with it, and also embed the coreos-installer binary and run it when booting from usb?

cgwalters:
> How will i inject the RHCOS image in the usb?

You won't need to.  There will be a single `.iso` file which contains the `-metal.img` via this PR.

jlebon:
OK, I rebased this!

Though I'd like to get https://github.com/coreos/coreos-assembler/pull/1417 in first, so that we can actually exercise the new path in CI here.

jlebon:
Restarted CI on this now that we have https://github.com/coreos/coreos-assembler/pull/1417!

I made no change to this PR during the rebase. Given it's already approved, I'm going to merge it once CI is green!

jlebon:
> Successfully tested scenario iso-offline-install for 31.20200430.dev.0 on bios (metal)

:tada:

Issue #201, opened 2020-03-31
Automatically select 4k metal image when needed


bgilbert:
Fedora CoreOS is now shipping a separate 4k metal image.  Automatically select it from stream metadata when presented with a 4k drive.

jlebon:
I started on this yesterday. Should have something up soon!

jlebon:
https://github.com/coreos/coreos-installer/pull/203

PR #203, opened 2020-03-31
install: use 4k image when when targeting 4k drives

jlebon:
Detect the sector size of the target device, then select the right
format from the stream metadata based on that.

Closes: #201

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -17,9 +17,10 @@ use nix::{errno::Errno, mount};
 use regex::Regex;
 use std::collections::HashMap;
 use std::convert::TryInto;
-use std::fs::{remove_dir, File};
+use std::fs::{remove_dir, File, OpenOptions};
 use std::num::NonZeroU32;
 use std::os::raw::c_int;
+use std::os::unix::fs::FileTypeExt;
 use std::os::unix::io::AsRawFd;
 use std::path::{Path, PathBuf};
 use std::process::Command;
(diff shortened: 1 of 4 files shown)


jlebon:
(Working on enhancing `kola testiso` to also test 4k installs.)

src/cmdline.rs
```diff
@@ -363,13 +371,19 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         } else {
             None
         };
+        let format = match get_sector_size_for_path(Path::new(&device))?.get() {
+            4096 => "4k.raw.xz",
+            // could bail on non-512, but let's be optimistic and just try the regular 512b image
+            _ => "raw.xz",
```
cgwalters, line 377:
```suggestion
            512 => "raw.xz",
            format => { eprintln!("Warning: Unhandled sector size '{}'; assuming 512", format); "raw.xz" },
```

src/blockdev.rs
```diff
@@ -209,6 +210,19 @@ pub fn reread_partition_table(file: &mut File) -> Result<()> {
     Ok(())
 }
 
+pub fn get_sector_size_for_path(device: &Path) -> Result<NonZeroU32> {
+    let dev = OpenOptions::new()
+        .read(true)
+        .open(Path::new(device))
+        .chain_err(|| format!("opening {:?}", device))?;
+
+    if !dev.metadata()?.file_type().is_block_device() {
```
bgilbert, line 219: Missing `chain_err` on `metadata()`.

src/cmdline.rs
```diff
@@ -12,10 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::path::Path;
```
bgilbert, line 15: The codebase generally groups `std` with other imports.

src/cmdline.rs
```diff
@@ -363,13 +371,19 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         } else {
             None
         };
+        let format = match get_sector_size_for_path(Path::new(&device))?.get() {
```
bgilbert, line 374: `chain_err`?

jlebon:
OK so https://github.com/coreos/coreos-assembler/pull/1312 enhances `kola testiso` for testing metal4k image installs. Will hook it up here and in the pipeline too afterwards, though it doesn't actually test the new codepath here since we're not doing a stream install. We could work on mocking that though... not sure if we should block on this. I *did* test this code on the current stable stream metadata though and it did the right thing. :)

Will address review comments tomorrow!

src/cmdline.rs
jlebon: Thanks, fixed! (Though I personally prefer separating std imports.)

jlebon:
I added another commit here which adds metal4k testing in CI!

Issue #145, opened 2020-01-20
Install with kargs should report failure to console


bgilbert:
# Feature Request #

## Desired Feature ##

If install is invoked via kargs and fails, the user has to manually press Enter and type `systemctl status coreos-installer.service` to see what happened.  We should automatically replay the install log to the console, similar to how we do for Ignition failures.

bgilbert:
Closing in favor of doing this distro-wide, which is https://github.com/coreos/fedora-coreos-tracker/issues/924.

PR #204, opened 2020-04-01
systemd: report failures to console before emergency shell

zonggen:
Currently users have to manually type `systemctl status coreos-installer.service` after pressing Enter when hitting error state.

Adds a service `coreos-installer-emergency.service` before entering `emergency.service` that reports failures to console when `coreos-installer.service` failed.

The script `systemd-coreos-installer-emergency` is a fork from `timeout.sh` of `ignition-dracut` with minor adjustment for coreos-installer. It is placed under `/usr/lib/systemd/systemd-coreos-installer-emergency`.

Currently `coreos-installer-emergency.service` is enabled when running `coreos-installer-service` before executing `coreo-installer` command.

Closes: https://github.com/coreos/coreos-installer/issues/145
Signed-off-by: Allen Bai <abai@redhat.com>
References: https://github.com/coreos/ignition-dracut/blob/master/dracut/99emergency-timeout/timeout.sh

### DIFF
--- a/systemd/coreos-installer-emergency.service
+++ b/systemd/coreos-installer-emergency.service
@@ -1,0 +1,22 @@
+# Forked and modified from /lib/systemd/system/emergency.service
+[Unit]
+Description=Report CoreOS Installer Failures to Console
+DefaultDependencies=no
+Conflicts=shutdown.target
+Conflicts=rescue.service
+Before=emergency.target
+Before=emergency.service
+
+[Service]
+Type=forking
+ExecStart=/usr/lib/systemd/systemd-coreos-installer-emergency
+StandardInput=tty-force
+StandardOutput=inherit
+StandardError=inherit
+KillMode=process
+IgnoreSIGPIPE=no
+SendSIGHUP=yes
+TimeoutSec=330
+
+[Install]
+WantedBy=emergency.target
--- a/systemd/coreos-installer-service
+++ b/systemd/coreos-installer-service
@@ -92,6 +92,10 @@ fi
 # Ensure device nodes have been created
 udevadm settle
 
+# Enable the service for reporting failures to console
+echo "systemctl enable coreos-installer-emergency.service"
+systemctl enable coreos-installer-emergency.service
+
 # Install
 echo "coreos-installer ${args[@]}"
 coreos-installer "${args[@]}"
--- a/systemd/systemd-coreos-installer-emergency
+++ b/systemd/systemd-coreos-installer-emergency
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+# Fored and modified from https://github.com/coreos/ignition-dracut/blob/master/dracut/99emergency-timeout/timeout.sh
+
+_wait_for_journalctl_to_stop() {
+    local time_since_last_log=0
+
+    local time_started="$(date '+%s')"
+    local now="$(date '+%s')"
+
+    while [ ${time_since_last_log} -lt 5 -a $((now-time_started)) -lt 15 ]; do
+        sleep 1
+
+        local last_log_timestamp="$(journalctl -e -n 1 -q -o short-unix | cut -d '.' -f 1)"
+        local now="$(date '+%s')"
+
+        local time_since_last_log=$((now-last_log_timestamp))
+    done
+}
+
+_prompt_for_timeout() {
+    local timeout=300
+    local interval=15
+
+    if systemctl show "coreos-installer.service" | grep -q "^ActiveState=failed$"; then
+        # coreos-installer has failed, suppress kernel logs so that coreos-installer logs stay
+        # on the screen
+        dmesg --console-off
+
+        # There's a couple straggler systemd messages. Wait until it's been 5
+        # seconds since something was written to the journal.
+        _wait_for_journalctl_to_stop
+
+        # Print CoreOS Installer logs
+        cat <<EOF
+-------------------------------------------------------------------------------
+CoreOS Installer has failed.
+Here are the CoreOS Installer logs:
+
+EOF
+        journalctl -u coreos-installer.service -b --no-pager --no-hostname -o cat
+        echo -n -e "\n"
+    fi
+
+    # Regularly prompt with time remaining.  This ensures the prompt doesn't
+    # get lost among kernel and systemd messages, and makes it clear what's
+    # going on if the user just connected a serial console.
+    while [[ $timeout > 0 ]]; do
+        local m=$(( $timeout / 60 ))
+        local s=$(( $timeout % 60 ))
+        local m_label="minutes"
+        if [[ $m = 1 ]]; then
+            m_label="minute"
+        fi
+
+        if [[ $s != 0 ]]; then
+            echo -n -e "Press Enter for emergency shell or wait $m $m_label $s seconds for reboot.      \r"
+        else
+            echo -n -e "Press Enter for emergency shell or wait $m $m_label for reboot.                 \r"
+        fi
+
+        local anything
+        if read -t $interval anything; then
+            return
+        fi
+
+        timeout=$(( $timeout - $interval ))
+    done
+
+    echo -e "\nRebooting."
+    # This is not very nice, but since reboot.target likely conflicts with
+    # the existing goal target wrt the desired state of shutdown.target,
+    # there doesn't seem to be a better option.
+    systemctl reboot --force
+    exit 0
+}
+
+# If we're invoked from a dracut breakpoint rather than
+# dracut-emergency.service, we won't have a controlling terminal and stdio
+# won't be connected to it. Explicitly read/write /dev/console.
+_prompt_for_timeout < /dev/console > /dev/console


zonggen:
```
[  OK  ] Reached target Network is Online.
         Starting CoreOS Installer...
[    6.210158] coreos-installer-service[769]: coreos-installer install /dev/sda
[    6.222847] coreos-installer-service[769]: Downloading stable image and signature
[   14.838394] coreos-installer-service[769]: ● coreos-installer.service - CoreOS Installer
[   14.839504] coreos-installer-service[769]:    Loaded: loaded (/usr/lib/systemd/system/coreos-installer.service; static; vendor preset: d)
[   14.841110] coreos-installer-service[769]:    Active: activating (start) since Wed 2020-04-01 13:45:48 UTC; 8s ago
[   14.842307] coreos-installer-service[769]:  Main PID: 769 (coreos-installe)
[   14.843194] coreos-installer-service[769]:     Tasks: 3 (limit: 2295)
[   14.844018] coreos-installer-service[769]:    Memory: 17.4M
[   14.844763] coreos-installer-service[769]:    CGroup: /system.slice/coreos-installer.service
[   14.846028] coreos-installer-service[769]:            ├─769 /usr/bin/bash /usr/libexec/coreos-installer-service
[   14.847346] coreos-installer-service[769]:            ├─807 coreos-installer install /dev/sda
[   14.848551] coreos-installer-service[769]:            └─814 systemctl status coreos-installer.service
[   14.849971] coreos-installer-service[769]: Apr 01 13:45:48 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com systemd[1]: Starting .
[   14.851890] coreos-installer-service[769]: Apr 01 13:45:48 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com coreos-installer-serva
[   14.853718] coreos-installer-service[769]: Apr 01 13:45:48 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com coreos-installer-serve
[   14.856099] coreos-installer-service[769]: Error: opening /dev/sda
[   14.857097] coreos-installer-service[769]: Caused by: No such file or directory (os error 2)
[FAILED] Failed to start CoreOS Installer.
See 'systemctl status coreos-installer.service' for details.
[DEPEND] Dependency failed for CoreOS Installer Target.
[  OK  ] Stopped Dispatch Password …ts to Console Directory Watch.
[  OK  ] Stopped Daily rotation of log files.
[  OK  ] Stopped Daily Cleanup of Temporary Directories.
[  OK  ] Closed Docker Socket for the API.
[  OK  ] Stopped Network Manager Wait Online.
         Stopping Network Manager...
         Stopping Hostname Service...
[  OK  ] Stopped Hostname Service.
[  OK  ] Stopped Network Manager.
         Stopping Network Manager Script Dispatcher Service...
[  OK  ] Stopped Network Manager Script Dispatcher Service.
[  OK  ] Stopped target Basic System.
         Stopping D-Bus System Message Bus...
[  OK  ] Stopped D-Bus System Message Bus.
[  OK  ] Closed D-Bus System Message Bus Socket.
[  OK  ] Stopped target System Initialization.
[  OK  ] Started Emergency Shell.
[  OK  ] Reached target Emergency Mode.
You are in emergency mode. After logging in, type "journalctl -xb" to view
system logs, "systemctl reboot" to reboot, "systemctl default" or "exit"
to boot into default mode.
Press Enter for maintenance

```

cgwalters:
This needs to be done conditionally on whether we were *invoked* via the systemd unit.  It's also supported to e.g. run from the container image, or by copying the binary to e.g. an Ubuntu system or whatever.  In both of those cases we aren't run via our systemd unit.

zonggen:
Not sure if this the approach we want to report failure to console (like in ignition) but if not, any hint would be helpful.

cc @bgilbert

cgwalters:
Also I think we should do a more general fix of having emergency mode just do the same thing that we have for the main login with failed units via https://github.com/rfairley/console-login-helper-messages/blob/master/usr/share/console-login-helper-messages/profile.sh

bgilbert (dismissed):
The problem we're trying to solve is in the glue that invokes coreos-installer, so we shouldn't fix it by adding a workaround inside coreos-installer itself.  Have you looked at modifying the `coreos-installer-service` script or the unit that invokes it?

zonggen:
After follow-up conversation with bgilbert, I believe we would need to

> something like: add a unit to emergency.target before emergency.service that checks for an installer failure and prints out logs

 ,which cgwalters also pointed out in the previous comment https://github.com/coreos/coreos-installer/pull/204#issuecomment-607270050. And this would also make sure `coreos-installer` is run via our systemd unit.

zonggen:
Demo:

demo_coreos_installer

systemd/coreos-installer-service
```diff
@@ -92,6 +92,10 @@ fi
 # Ensure device nodes have been created
 udevadm settle
 
+# Enable the service for reporting failures to console
+echo "systemctl enable coreos-installer-emergency.service"
```
zonggen, line 96: This does not seem right, either this or enabling the service as a `postprocess` in `manifest.yaml` under FCOS config directory would work, though it will be much better if the service is enabled by default..

zonggen:
Concerns: 
 - [ ] better to enable service by default instead of enabling it as a postprocess script
 - [ ] user has to hit [Enter] twice in order to get to interactive shell (one for entering emergency mode from the `coreos-installer-emergency.service`, one for entering interactive shell in emergency mode)
 -  [ ] after entering emergency mode, hitting [Control - D] instead of [Enter] will run the coreos installer emergency service again but this time the user cannot type [Enter] (i.e. the stdin is not correct). Although the user could stop the service by hitting [Control - C] and enter emergency mode again, it's better to keep the behavior consistent.

cgwalters:
FWIW I tried generalizing the ignition-dracut emergency console stuff over here: https://github.com/coreos/ignition-dracut/pull/168

That only runs in the initramfs, but...offhand I don't see a reason why we shouldn't have the same behavior for `emergency.target` in the initramfs and the real root?

zonggen:
If you meant overriding default `emergency.service` and run the reporting script during the service, I was trying to avoid pollute the default emergency service and try to separate out the reporting service and added a standalone service _before_ `emergency.service`

zonggen:
For generalizing the script to all service failures, I'm not completely sure if catching any service failure and hanging for user input is desirable UX but I could be wrong...

systemd/coreos-installer-service
bgilbert: We should do this from the generator.

systemd/coreos-installer-emergency.service
```diff
@@ -0,0 +1,22 @@
...
+StandardInput=tty-force
+StandardOutput=inherit
+StandardError=inherit
+KillMode=process
+IgnoreSIGPIPE=no
+SendSIGHUP=yes
+TimeoutSec=330
+
+[Install]
+WantedBy=emergency.target
```
bgilbert, line 22: If the generator creates the symlink directly, this is not needed.

systemd/coreos-installer-emergency.service
```diff
@@ -0,0 +1,22 @@
...
+[Unit]
+Description=Report CoreOS Installer Failures to Console
+DefaultDependencies=no
+Conflicts=shutdown.target
+Conflicts=rescue.service
+Before=emergency.target
+Before=emergency.service
+
+[Service]
+Type=forking
```
bgilbert, line 11: Shouldn't this be `oneshot`?

systemd/coreos-installer-emergency.service
```diff
@@ -0,0 +1,22 @@
...
+ExecStart=/usr/lib/systemd/systemd-coreos-installer-emergency
```
bgilbert, line 12: For consistency, this should be `/usr/libexec/coreos-installer-emergency`.

systemd/coreos-installer-emergency.service
```diff
@@ -0,0 +1,22 @@
...
+StandardInput=tty-force
```
bgilbert, line 13: Without the timeout prompt, we don't need this.

systemd/coreos-installer-emergency.service
```diff
@@ -0,0 +1,22 @@
...
+StandardOutput=inherit
```
bgilbert, line 14: ...which means this should be `tty-force`.

systemd/coreos-installer-emergency.service
```diff
@@ -0,0 +1,22 @@
...
+StandardError=inherit
+KillMode=process
+IgnoreSIGPIPE=no
+SendSIGHUP=yes
+TimeoutSec=330
```
bgilbert, line 19: I don't think any of these are needed?

systemd/systemd-coreos-installer-emergency
```diff
@@ -0,0 +1,80 @@
...
+    # This is not very nice, but since reboot.target likely conflicts with
+    # the existing goal target wrt the desired state of shutdown.target,
+    # there doesn't seem to be a better option.
+    systemctl reboot --force
+    exit 0
+}
+
+# If we're invoked from a dracut breakpoint rather than
+# dracut-emergency.service, we won't have a controlling terminal and stdio
+# won't be connected to it. Explicitly read/write /dev/console.
```
bgilbert, line 79: Shouldn't be relevant here.

bgilbert (commented):
I think it could make sense to generalize this to report all failed services, as @cgwalters suggests, in which case this should go in fedora-coreos-config.

bgilbert:
Closing in favor of doing this distro-wide, which is https://github.com/coreos/fedora-coreos-tracker/issues/924.

PR #185, opened 2020-03-11
[WIP] multipath support

tuan-hoang1:

PR #211, opened 2020-04-06
blockdev: use 'kpartx' for reading/clearing partition table of dm-devices

nikita-dubrovskii:
When installing CoreOS on DM device, such as '/dev/mapper/mpatha' or '/dev/dm-0',
regular 'ioctl(fd, BLKRRPART)' call always fails with EINVAL. This also means,
that our check for device being mounted, which on regular devices fails with EBUSY,
is not valid anymore, that's why now /proc/mounts is used.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use error_chain::bail;
+use error_chain::{bail, ensure};
 use nix::sys::stat::{major, minor};
 use nix::{errno::Errno, mount};
 use regex::Regex;
@@ -233,7 +233,34 @@ impl Drop for Mount {
     }
 }
 
-pub fn reread_partition_table(file: &mut File) -> Result<()> {
+/// Check if device is DM-device
+pub fn is_dm_device<T: AsRef<str>>(device: T) -> bool {
+    let device = device.as_ref();
+    device.starts_with("/dev/mapper/") || device.starts_with("/dev/dm-")
+}
+
+/// Check whether device mounted or not
+pub fn is_mounted<T: AsRef<str>>(device: T) -> Result<bool> {
+    let device = device.as_ref();
+    // looking at /proc/mounts is the best way to be 100% sure of seeing what is mounted on system,
+    // as the kernel is providing this information
+    let mounts = read_to_string("/proc/mounts").chain_err(|| "reading /proc/mounts")?;
+    if let Some(mp) = mounts.lines().find(|v| v.starts_with(device)) {
+        eprintln!("{} is mounted: {}", device, mp);
+        return Ok(true);
+    }
+    Ok(false)
+}
+
+pub fn reread_partition_table<T: AsRef<str>>(device: T, file: &mut File) -> Result<()> {
+    let device = device.as_ref();
+    if is_dm_device(device) {
+        return kpartx_reread_partitions(device);
+    }
+    ioctl_reread_partitions(file)
+}
+
+fn ioctl_reread_partitions(file: &mut File) -> Result<()> {
     let fd = file.as_raw_fd();
     // Reread sometimes fails inexplicably.  Retry several times before
     // giving up.
@@ -259,6 +286,31 @@ pub fn reread_partition_table(file: &mut File) -> Result<()> {
             }
         }
     }
+    Ok(())
+}
+
+/// Read partition table of DM-devices
+fn kpartx_reread_partitions<T: AsRef<str>>(device: T) -> Result<()> {
+    let device = device.as_ref();
+    let status = std::process::Command::new("kpartx")
+        .arg("-u")
+        .arg(device)
+        .stderr(std::process::Stdio::null())
+        .status()
+        .chain_err(|| "running kpartx -u")?;
+    ensure!(status.success(), "kpartx -d {} failed: {}", device, status);
+    Ok(())
+}
+
+/// Clean partition table of DM-devices
+pub fn kpartx_delete_partitions<T: AsRef<str>>(device: T) -> Result<()> {
+    let device = device.as_ref();
+    let status = std::process::Command::new("kpartx")
+        .arg("-d")
+        .arg(device)
+        .status()
+        .chain_err(|| "running kpartx -d")?;
+    ensure!(status.success(), "kpartx -u {} failed: {}", device, status);
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


bgilbert:
Related: #91

We generally try to minimize dependencies on external programs, so coreos-installer intentionally doesn't do this.  The current situation could use improvement though.  I guess we could avoid the external dependency by parsing the partition table and issuing the ioctls ourselves, but that seems a bit painful.

src/blockdev.rs
```diff
@@ -202,6 +202,15 @@ pub fn reread_partition_table(file: &mut File) -> Result<()> {
                         return Err(err).chain_err(|| "couldn't reread partition table");
                     }
                 } else {
+                    let status = std::process::Command::new("kpartx")
```
bgilbert, line 205: This should run after all retries, not after every retry.

nikita-dubrovskii: Done

src/blockdev.rs
```diff
@@ -191,6 +191,14 @@ pub fn reread_partition_table(file: &mut File) -> Result<()> {
             Ok(_) => break,
             Err(err) => {
                 if retries == 0 {
+                    let status = std::process::Command::new("kpartx")
```
jlebon, line 194: Could we fallback to this only if we get `Errno::EINVAL`?

bgilbert:
I realized I was thinking of `partprobe`, not `kpartx`.  So, a few things:

1. If the device is actually a `/dev/mapper` device (e.g. created with `echo "0 16777216 linear /dev/loop0 0" | dmsetup create loopdev` then this patch works fine.  After `kpartx -u` runs, `lsblk` correctly returns e.g. `loopdev1` as the name of the first partition.

    If the device is not a DM device, this doesn't work, because `kpartx` doesn't actually create partitions; it creates DM devices which map each partition read from the partition table.  I tried using this patch to install onto a non-partitionable loop device and got:

```
$ sudo target/debug/coreos-installer install -f ~/tmp/web/fedora-coreos-30.20191014.0-metal.x86_64.raw.xz /dev/loop0 -i ~/fcos/live-var.ign
Copying image from /home/bgilbert/tmp/web/fedora-coreos-30.20191014.0-metal.x86_64.raw.xz
Reading signature from /home/bgilbert/tmp/web/fedora-coreos-30.20191014.0-metal.x86_64.raw.xz.sig
GPT:Primary header thinks Alt. header is not at the end of the disk.
GPT:Alternate GPT header not at the end of the disk.
GPT: Use GNU Parted to correct GPT errors.
gpg: Signature made Mon 14 Oct 2019 11:59:16 PM EDT
gpg:                using RSA key EF3C111FCFC659B9
gpg: Good signature from "Fedora (30) <fedora-30-primary@fedoraproject.org>" [ultimate]
> Read disk 416.2 MiB/416.2 MiB (100%)   
GPT:Primary header thinks Alt. header is not at the end of the disk.
GPT:Alternate GPT header not at the end of the disk.
GPT: Use GNU Parted to correct GPT errors.
Error: mounting device /dev/loop0p1 on /tmp/coreos-installer-dIA7ra
Caused by: ENOENT: No such file or directory
Clearing partition table
Error: install failed
```

Which makes sense, because we're still looking for the partitions in the usual place:

```
$ sudo lsblk --pairs --output NAME,LABEL,FSTYPE /dev/loop0
NAME="loop0" LABEL="" FSTYPE=""
NAME="loop0p1" LABEL="boot" FSTYPE="ext4"
NAME="loop0p2" LABEL="EFI-SYSTEM" FSTYPE="vfat"
NAME="loop0p3" LABEL="" FSTYPE=""
NAME="loop0p4" LABEL="root" FSTYPE="xfs"
```

So we should probably only run the `kpartx` path on DM devices.  Or we could go the other direction and use `kpartx` in all cases, and update our expectation of where to find device nodes.

2. `kpartx -u` creates the partition mappings, but there's nothing in the code that deletes them after we're done.

3. [This code](https://github.com/coreos/coreos-installer/blob/89fa9cd14ed090a6a8b37b9613a84d6845f236f2/src/install.rs#L70-L71) uses `BLKRRPART` as a safety check to make sure no filesystems on the disk are mounted before overwriting them.  `kpartx -u` will succeed even if an existing partition is mounted, so we'd need another way to perform that check.

3. The "invalid GPT" errors are benign and we might want to filter them out to avoid confusing the user.

4. Minor: the Dockerfile currently does not install `kpartx` in the container.

src/blockdev.rs
```diff
@@ -191,6 +191,14 @@ pub fn reread_partition_table(file: &mut File) -> Result<()> {
...
+                        .arg("-u")
```
bgilbert, line 195: Do we also need to pass `-s`?

src/blockdev.rs
nikita-dubrovskii: Done

src/blockdev.rs
nikita-dubrovskii: i'm not sure. my expectation is that after successful copying of metal-image to disk we don't have to wait until partitions are created,because by that time disk should be ready, and we only have to re-read partition table.

bgilbert: Sure, but we need to wait for the device nodes to appear before using them.  `udev_settle()` tries to do that, but there's currently still a possible race condition.  It should be possible to address that generically with #155 or something like it, so I'm okay omitting `-s` if we want to exercise `udev_settle()` uniformly in all paths.

nikita-dubrovskii: during testing i had extra calls to `udevadm trigger -c add` for `subsystems and devices` within `udev_settle()`, but it looked like an overhead, so it was removed.
What do you think about using it?

bgilbert: As I understand it, `udevadm trigger` shouldn't generally be used in production.  udev events will already be generated by the new device nodes; we just need to make sure we're waiting for them appropriately.

nikita-dubrovskii: Agree. Is there smth else i have to modify/improve ?

bgilbert: I think we're okay on the udev part.  Items 1-4 of https://github.com/coreos/coreos-installer/pull/211#issuecomment-610705093 should still be addressed.

nikita-dubrovskii:
Btw, i guess this PR could help closing https://github.com/coreos/coreos-installer/issues/91

tuan-hoang1:
Probably having the same thing : https://github.com/coreos/coreos-installer/pull/185/

bgilbert:
Alternative implementation in #285.

Issue #205, opened 2020-04-03
install: UI for forwarding network information into next boot 


dustymabe:
Right now I'm working on an interactive experience for users to configure networking of the system in the booted Live ISO right before they run `coreos-install install`. As a user I'm currently:

- getting the system booted up (no DHCP) to boot prompt
- running `nmtui` to configure networking and gain internet access
- run curl to grab ignition config
- run coreos-installer to install to disk
- reboot

I need to be able to take the networking information from the current system (for FCOS it would be `/etc/NetworkManager/system-connections`) into the initramfs of the firstboot. My current thinking is that we just copy the files from `/etc/NetworkManager/system-connections` into a directory in the boot partition like we do for other things and then have a systemd unit that runs in the initramfs to pick that information up and carry it forward.

Assuming that all sounds good what do we want the UI for this to look like? One option:

- `coreos-installer install --forward-networking[=/etc/NetworkManager/system-connections]`
    - defaults to `/etc/NetworkManager/system-connections` but directory can be overridden
    - the option could be `--networking-files=` instead maybe?

Another option is that we make this more generic and make `--initramfsfiles=/path/to/initramfs/rootfs` and make a generic way to copy files into the initramfs environment but I don't know how much we should support something like this.

Thoughts?

dustymabe:
Talking with @bgilbert on this topic I think we settled on something like

- short option `-n`
- long option `--forward-networking-files`
- the options optionally accept a value that is a path to a directory where the networking files are located
    - defaults to `/etc/NetworkManager/system-connections`

bgilbert:
I'm so-so on `--forward-networking-files`.  "files" are an implementation detali.  `--copy-network-config`?

dustymabe:
I could go either way on those two. As long as you're still OK with the optional accepting a `/path/to/networking/config/files/` to override the default.

jlebon:
Could also be a more generic `--copy-config /etc/subpath`.

bgilbert:
@jlebon, I think going that far would start encouraging users to bypass Ignition.

jlebon:
Yeah, true. Though on that note, maybe what we actually want is to have whatever wraps NetworkManager-tui to spit out an Ignition config that can be merged into the main config (could easily build that functionality on top of https://github.com/coreos/fcct/issues/83).

That way users have the canonical config to use for future automated re-installs (and also for other bare metal hosts that need installing if the network config isn't inherently node-specific).

[...]

PR #212, opened 2020-04-14
install: add --copy-network and --network-dir options

dustymabe:
This will allow us to copy in networking configuration from a specified
location into a predetermined directory inside of `/boot/` of the
installed system (which follows the path of other postprocess steps).
Having this mechanism will allow for a user to interactively configure
networking in the Live ISO environment via `nmtui` and then have those
settings propagate forward into the installed system.

After the files are copied into the directory within `/boot/` a systemd
unit from the initramfs will pick them up on firstboot and propagate
them into the appropriate locations.

Fixes: https://github.com/coreos/coreos-installer/issues/205

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -38,6 +38,7 @@ pub struct InstallConfig {
     pub firstboot_kargs: Option<String>,
     pub insecure: bool,
     pub preserve_on_error: bool,
+    pub network_config: Option<String>,
 }
 
 pub struct DownloadConfig {
@@ -135,6 +136,22 @@ pub fn parse_args() -> Result<Config> {
                         .value_name("args")
                         .help("Additional kernel args for the first boot")
                         .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("copy-network")
+                        .short("n")
+                        .long("copy-network")
+                        .help("Copy network config from install environment"),
+                )
+                .arg(
+                    Arg::with_name("network-dir")
+                        .long("network-dir")
+                        .value_name("path")
+                        .default_value("/etc/NetworkManager/system-connections/")
+                        .takes_value(true)
+                        .empty_values(false)
+                        .help("For use with -n.")
+                        .next_line_help(true), // so we can stay under 80 chars
                 )
                 // obscure options without short names
                 .arg(
@@ -400,6 +417,15 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
     // and report it to the user
     eprintln!("{}", location);
 
+    // If the user requested us to copy networking config by passing
+    // -n or --copy-network then copy networking config from the
+    // directory defined by --network-dir.
+    let network_config = if matches.is_present("copy-network") {
+        matches.value_of("network-dir").map(String::from)
+    } else {
+        None
+    };
+
     // build configuration
     Ok(Config::Install(InstallConfig {
         device,
@@ -409,6 +435,7 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         firstboot_kargs: matches.value_of("firstboot-kargs").map(String::from),
         insecure: matches.is_present("insecure"),
         preserve_on_error: matches.is_present("preserve-on-error"),
+        network_config,
     }))
 }
 
--- a/src/install.rs
+++ b/src/install.rs
@@ -14,7 +14,7 @@
 
 use error_chain::{bail, ChainedError};
 use nix::mount;
-use std::fs::{create_dir_all, read_dir, File, OpenOptions};
+use std::fs::{copy as fscopy, create_dir_all, read_dir, File, OpenOptions};
 use std::io::{copy, Read, Seek, SeekFrom, Write};
 use std::os::unix::fs::FileTypeExt;
 use std::path::Path;
@@ -111,7 +111,11 @@ fn write_disk(
     udev_settle()?;
 
     // postprocess
-    if ignition.is_some() || config.firstboot_kargs.is_some() || config.platform.is_some() {
+    if ignition.is_some()
+        || config.firstboot_kargs.is_some()
+        || config.platform.is_some()
+        || config.network_config.is_some()
+    {
         let mount = mount_partition_by_label(&config.device, "boot", mount::MsFlags::empty())?;
         if let Some(ignition) = ignition {
             write_ignition(mount.mountpoint(), ignition)?;
@@ -121,6 +125,9 @@ fn write_disk(
         }
         if let Some(platform) = config.platform.as_ref() {
             write_platform(mount.mountpoint(), platform)?;
+        }
+        if let Some(network_config) = config.network_config.as_ref() {
+            copy_network_config(mount.mountpoint(), network_config)?;
         }
     }
 
@@ -227,6 +234,37 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
             config
                 .write(new_contents.as_bytes())
                 .chain_err(|| format!("writing {}", path.display()))?;
+        }
+    }
+
+    Ok(())
+}
+
+/// Copy networking config if asked to do so
+fn copy_network_config(mountpoint: &Path, net_config_src: &str) -> Result<()> {
+    eprintln!("Copying networking configuration from {}", net_config_src);
+
+    // get the path to the destination directory
+    let net_config_dest = mountpoint.join("coreos-installer-network");
+
+    // make the directory if it doesn't exist
+    create_dir_all(&net_config_dest).chain_err(|| {
+        format!(
+            "creating destination networking config directory {}",
+            net_config_dest.display()
+        )
+    })?;
+
+    // copy files from source to destination directories
+    for entry in
+        read_dir(&net_config_src).chain_err(|| format!("reading directory {}", net_config_src))?
+    {
+        let entry = entry.chain_err(|| format!("reading directory {}", net_config_src))?;
+        let srcpath = entry.path();
+        let destpath = net_config_dest.join(entry.file_name());
+        if srcpath.is_file() {
+            eprintln!("Copying {} to installed system", srcpath.display());
+            fscopy(&srcpath, &destpath).chain_err(|| "Copying networking config")?;
         }
     }
 


dustymabe:
This is my first ever rust PR. It took me waaaay too long to get this code working 😜. Please be gentle in code review.

marking as WIP for now as I test things out more

src/install.rs
```diff
@@ -233,6 +236,53 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
...
+        .chain_err(|| format!("reading directory {}", net_config_dest.display()))?
+    {
+        let entry = entry
+            .chain_err(|| format!("reading directory {}", net_config_dest.display()))?;
+        if entry.path().is_file() {
+            eprintln!("Removing file {}", entry.path().display());
+            remove_file(entry.path()).
+                chain_err(|| format!("removing {}", entry.path().display()))?;
+        }
+    }
```
dustymabe, line 267: I'll delete this code as this should never happen because we just wrote out the fresh disk image to the block device.

src/cmdline.rs
```diff
@@ -136,6 +137,15 @@ pub fn parse_args() -> Result<Config> {
                         .help("Additional kernel args for the first boot")
                         .takes_value(true),
                 )
+                .arg(
+                    Arg::with_name("copy-network-config")
+                        .short("n")
+                        .long("copy-network-config")
+                        .value_name("path")
+                        .default_value("/etc/NetworkManager/system-connections/")
```
dustymabe, line 145: This default value will likely need to be different for RHCOS (for now). I'm thinking we can just carry that as a patch in the RPM in downstream. Does that sound OK?

src/install.rs
```diff
@@ -233,6 +236,53 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
     Ok(())
 }
 
+/// Copy networking config if asked to do so
+fn copy_network_config(mountpoint: &Path, net_config_src: &str) -> Result<()> {
+    eprintln!("Copying networking configuration from {}", net_config_src);
```
cgwalters, line 241: Since this isn't an error, let's use `println!`.

src/install.rs
```diff
@@ -233,6 +236,53 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
...
+
+    // get the path to the destination directory
+    let mut net_config_dest = mountpoint.to_path_buf();
+    net_config_dest.push("coreos-installer-network");
```
cgwalters, line 245: This can just be:
```
let net_config_dest = mountpoint.join("coreos-installer-network");
```

src/install.rs
```diff
@@ -233,6 +236,53 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
...
+
+    // make the directory if it doesn't exist
+    create_dir_all(&net_config_dest)
+        .chain_err(|| {
+            format!(
+                "creating destination networking config directory {}",
+                net_config_dest.display()
```
cgwalters, line 252: I don't think you need the explicit `.display()` here.

src/cmdline.rs
cgwalters: We also do want to copy that path for RHCOS too.  So it's more like disabling ifcfg copying on FCOS.  We could do something like look at the *target* OS install and conditionalize based on that...since we do want to allow people to use the upstream/Fedora `coreos-installer` binary to install RHCOS.

cgwalters: Or to rephrase, FCOS supports NM config files; RHCOS supports NM config files *and* ifcfg files.  It's not "RHCOS only supports ifcfg files".

Perhaps simplest is just to make this a boolean value, and copy files from both paths if they exist and the option is specified.

cgwalters:
> This will allow us to copy in networking configuration from a specified
location into a predetermined directory inside of /boot/ of the
installed system

Another approach would be to convert the files into an Ignition config that gets merged with any other config provided.  Something like having Ignition support a `/boot/config.ign.d` directory instead of just `/boot/config.ign`, and then we drop a `/boot/config.ign.d/coreos-installer.ign`?

dustymabe:
> > This will allow us to copy in networking configuration from a specified
> > location into a predetermined directory inside of /boot/ of the
> > installed system
> 
> Another approach would be to convert the files into an Ignition config that gets merged with any other config provided. Something like having Ignition support a `/boot/config.ign.d` directory instead of just `/boot/config.ign`, and then we drop a `/boot/config.ign.d/coreos-installer.ign`?

We discussed this a bit before I think: https://github.com/coreos/coreos-installer/issues/205#issuecomment-610576430. We need the networking to be brought up in the initramfs.

src/cmdline.rs
dustymabe: > We also do want to copy that path for RHCOS too. So it's more like disabling ifcfg copying on FCOS. We could do something like look at the _target_ OS install and conditionalize based on that...since we do want to allow people to use the upstream/Fedora `coreos-installer` binary to install RHCOS.

Using the FCOS installer to install RHCOS is probably a little risky long term, though for this case the user could still pass `-n /etc/sysconfig/network-scripts/` and get it to work.

> Or to rephrase, FCOS supports NM config files; RHCOS supports NM config files _and_ ifcfg files. It's not "RHCOS only supports ifcfg files".
> 
> Perhaps simplest is just to make this a boolean value, and copy files from both paths if they exist and the option is specified.

Since the use case here is to propagate forward the files generated by `nmtui` I think we should support the default format that is spit out by `nmtui` on each platform and not overcomplicate things.

src/install.rs
dustymabe: will look into this. Thanks for the suggestion.

src/install.rs
dustymabe: cool I'll investigate this and check

src/install.rs
dustymabe: ok, I'll check. thanks!

src/cmdline.rs
cgwalters: > Since the use case here is to propagate forward the files generated by nmtui I think we should support the default format that is spit out by nmtui on each platform and not overcomplicate things.

Hmm.  I suppose you're right that NM will save in ifcfg files by default on RHCOS.  I wonder how hard that'd be to change though...i.e. make ifcfg files a *read* source but not a write target.

dustymabe: > i.e. make ifcfg files a _read_ source but not a write target.

+1, that's what it does for nm keyfiles today IIUC. I'd have to dig into the plugin to see if it's possible. For now I'm going to stick with this and I'll come back and change it if I have enough time between now and this Friday.

dustymabe:
The systemd unit/script that will pick up these networking config files placed by coreos-installer is in https://github.com/coreos/fedora-coreos-config/pull/346

dustymabe:
I'll get to the code review comments/suggestions in this PR first thing tomorrow. I'll be doing a good dose of testing as well.

src/install.rs
bgilbert: Nope, all of our existing status logging goes to stderr.

src/cmdline.rs
```diff
@@ -136,6 +137,15 @@ pub fn parse_args() -> Result<Config> {
...
+                        .takes_value(true)
+                        .help("Optionally copy the network config from <path> in the installer environment"),
```
bgilbert, line 147: This breaks the 80-column limit on readable help text.  `Copy network config from <path>` almost fixes it, but the option description (`--copy-network-config <path>`) is so wide that it shifts _other_ options' help text to the right.

Which intersects with another issue: I don't think the single option is going to work.  `coreos-installer install -n /dev/loop0` and `coreos-installer install --copy-network-config /dev/loop0` both misinterpret `/dev/loop0` as the path to the network configs.  Without a more discerning parser, I think we should split this into two options: a boolean, and an option to override the default path.

Dropping `<path>` from the option description would fix the help text length without needing to rename the option, though it then moves the problem to choosing the name of the second option.

I know the character-counting is painful but we don't really get usable help text otherwise.  We could enable the `wrap_help` clap feature but it's... only slightly less bad:

```
Install Fedora CoreOS or RHEL CoreOS

USAGE:
    coreos-installer install [OPTIONS] <device>

OPTIONS:
    -s, --stream <name>                 Fedora CoreOS stream [default: stable]
    -u, --image-url <URL>               Manually specify the image URL
    -f, --image-file <path>             Manually specify a local image file
    -i, --ignition-file <path>          Embed an Ignition config from a file
    -p, --platform <name>               Override the Ignition platform ID
        --firstboot-args <args>
            Additional kernel args for the first boot

    -n, --copy-network-config <path>
            Optionally copy the network config from <path> in the installer
            environment [default: /etc/NetworkManager/system-connections/]
        --insecure                      Skip signature verification
        --stream-base-url <URL>
            Base URL for Fedora CoreOS stream metadata

        --architecture <name>
            Target CPU architecture [default: x86_64]

        --preserve-on-error             Don't clear partition table on error
    -h, --help                          Prints help information

ARGS:
    <device>    Destination device
```

src/install.rs
```diff
@@ -233,6 +236,53 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
...
+    // copy files from source to destination directories
+    for entry in read_dir(&net_config_src)
+        .chain_err(|| format!("reading directory {}", net_config_src))?
+    {
+        let entry = entry
+            .chain_err(|| format!("reading directory {}", net_config_src))?;
+        let srcpath = entry.path();
+        let destpath = net_config_dest.join(entry.file_name());
+        if srcpath.is_file() {
+            eprintln!("Copying {} to installed system", srcpath.display());
```
bgilbert, line 278: This seems too verbose.  We're already announcing that we're copying the parent directory.

src/install.rs
dustymabe: So I'll leave this as is in this PR, but it's probably worth us having a followup where we convert our non-error status logging to non stderr, right?

src/install.rs
dustymabe: I think it does more good than harm because a user could be claiming that they set up networking and it wasn't getting propagated properly. Well, how did they set up networking? Did they use nmtui? Did the copy files around themselves? Did they use `ip` directly?

This little bit of information let's us know at least what file(s) got copied over, if any.

src/install.rs
dustymabe: looks like we do need it:

```
error[E0277]: `std::path::PathBuf` doesn't implement `std::fmt::Display`
   --> src/install.rs:251:17
    |
251 |                 net_config_dest
    |                 ^^^^^^^^^^^^^^^ `std::path::PathBuf` cannot be formatted with the default formatter
    |
    = help: the trait `std::fmt::Display` is not implemented for `std::path::PathBuf`
```

src/cmdline.rs
dustymabe: > This breaks the 80-column limit on readable help text. `Copy network config from <path>` almost fixes it, but the option description (`--copy-network-config <path>`) is so wide that it shifts _other_ options' help text to the right.
> 
> Which intersects with another issue: I don't think the single option is going to work. `coreos-installer install -n /dev/loop0` and `coreos-installer install --copy-network-config /dev/loop0` both misinterpret `/dev/loop0` as the path to the network configs. Without a more discerning parser, I think we should split this into two options: a boolean, and an option to override the default path.

I'll work on adding the second option.

> Dropping `<path>` from the option description would fix the help text length without needing to rename the option, though it then moves the problem to choosing the name of the second option.

I'll try to get the text to under 80 characters.

> 
> I know the character-counting is painful but we don't really get usable help text otherwise. We could enable the `wrap_help` clap feature but it's... only slightly less bad:

On a console readability within 80 characters in important.

src/install.rs
dustymabe: 👍 this seems to work, thanks!

src/install.rs
dustymabe: In the vast majority of cases there will probably only be 1 interface so one extra line of logging output.

src/cmdline.rs
dustymabe: The closest I could get was changing the long form of `-n` to `--copy-network`, adding a `--network-dir` option and using `.next_line_help(true)`. Here's what we end up with, which is less than 80 characters and I think a good compromise:

```
Install Fedora CoreOS or RHEL CoreOS

USAGE:
    coreos-installer install [OPTIONS] <device>

OPTIONS:
    -s, --stream <name>            Fedora CoreOS stream [default: stable]
    -u, --image-url <URL>          Manually specify the image URL
    -f, --image-file <path>        Manually specify a local image file
    -i, --ignition-file <path>     Embed an Ignition config from a file
    -p, --platform <name>          Override the Ignition platform ID
        --firstboot-args <args>    Additional kernel args for the first boot
    -n, --copy-network             Copy network config from install environment
        --network-dir <path>
            For use with -n. [default: /etc/NetworkManager/system-connections/]

        --insecure                 Skip signature verification
        --stream-base-url <URL>    Base URL for Fedora CoreOS stream metadata
        --architecture <name>      Target CPU architecture [default: x86_64]
        --preserve-on-error        Don't clear partition table on error
    -h, --help                     Prints help information

ARGS:
    <device>    Destination device
```

dustymabe:
based on https://github.com/coreos/coreos-installer/pull/212#discussion_r408669758 I changed this to `install: add --copy-network and --network-dir options`.

src/install.rs
bgilbert: Why?  Status reports aren't the primary output of the program, so it makes sense to me that they would go to stderr.

cgwalters: I grant you this is somewhat debatable but I think precedent weighs most strongly towards `stderr` is for *errors*.

dustymabe: Yes I typically think that you only print to stderr if there is a problem. That's mostly what I've experienced. I don't know if there is a written down standard anywhere.

bgilbert: I don't agree.  If we ever add e.g. a JSON report describing what actions `install` has taken, that would go to stdout for piping to another process, and the status reports would need to not be commingled with it.

@lucab, thoughts?

src/cmdline.rs
bgilbert: We could avoid the second line with e.g. `Override default network config path` and `hide_default_value(true)`.

src/install.rs
bgilbert: Okay, makes sense.

src/cmdline.rs
dustymabe: I do think showing the user the default is important.

src/install.rs
dustymabe: > I don't agree. If we ever add e.g. a JSON report describing what actions `install` has taken, that would go to stdout for piping to another process, and the status reports would need to not be commingled with it.

Sounds like a good use case for a `--quiet`.

dustymabe: Either way maybe we should move this coversation into an issue since it no longer really reflects a change we are going to make in this PR.

dustymabe: follow up: https://github.com/coreos/coreos-installer/issues/214

cgwalters: > I don't agree. If we ever add e.g. a JSON report describing what actions install has taken, that would go to stdout for piping to another process, and the status reports would need to not be commingled with it.

What you're talking about is one of the most long-standing things in how Unix pipelines work - the fact they're just connected by bytestreams and there's no common APIs or structure.  A lot of successor projects try to address this, from [Windows Powershell](https://docs.microsoft.com/en-us/powershell/scripting/getting-started/getting-started-with-windows-powershell?view=powershell-7) to [nushell](https://github.com/nushell/nushell) to many others.  (In fact long ago I tried to write one myself...)

Practically speaking, I think it's better to do e.g. `--status-report-json /path/to/status.json`.  Or alternatively `--output=json` and that also suppresses internal printing.

src/cmdline.rs
dustymabe: opened this against NM upstream to see if there's any way to do what we want: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/415

PR #213, opened 2020-04-15
systemd: remove SYSTEMD_SULOGIN_FORCE override for emergency.service

dustymabe:
This is no longer needed since we changed all of Fedora CoreOS to
set SYSTEMD_SULOGIN_FORCE=1 for emergency.service in
https://github.com/coreos/fedora-coreos-config/commit/eb74f2ea3e9b453902315539e4f327481162c4f8


dustymabe:
rebased on top of latest master

PR #215, opened 2020-04-16
blockdev: get path directly from lsblk output

nikita-dubrovskii:

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -72,8 +72,10 @@ struct BlkDev {
 
 fn get_partitions(device: &str) -> Result<Vec<BlkDev>> {
     // have lsblk enumerate partitions on the device
+    // Older lsblk, e.g. in CentOS 7.6, doesn't support PATH, but -p option
     let result = Command::new("lsblk")
         .arg("--pairs")
+        .arg("--paths")
         .arg("--output")
         .arg("NAME,LABEL,FSTYPE")
         .arg(device)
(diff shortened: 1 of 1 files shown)


src/blockdev.rs
```diff
@@ -72,8 +72,10 @@ struct BlkDev {
 
 fn get_partitions(device: &str) -> Result<Vec<BlkDev>> {
     // have lsblk enumerate partitions on the device
+    // Older lsblk, e.g. in CentOS 7.6, doesn't support PATH, but -p option
     let result = Command::new("lsblk")
         .arg("--pairs")
+        .arg("-p")
```
jlebon, line 78: Can we use `--paths` here instead to be more clear?

src/blockdev.rs
```diff
@@ -91,27 +93,13 @@ fn get_partitions(device: &str) -> Result<Vec<BlkDev>> {
...
-            .parent()
-            .chain_err(|| format!("couldn't get parent directory of {}", device))?
-            .to_path_buf();
-        match fields.get("NAME") {
-            None => continue,
-            Some(name) => path.push(name),
-        }
-        // Skip the device itself
-        if path == Path::new(device) {
-            continue;
```
jlebon, line 107: I think we still need some logic like this though, but I guess comparing to `/dev/${device}` instead?

jlebon:
Could you add some information in the commit message about what issue this patch is resolving and why? Providing context is important both for reviewers and for posterity. And maybe a commit message title like: `blockdev: get path directly from lsblk output` or similar?

src/blockdev.rs
nikita-dubrovskii: I guess `device` always contains `/dev/` prefix.
But i don't think we really need to filter out device itself, the only reason to do it is to fail with `couldn't find any partitions on {device}`, i think it's better to fail with `couldn't find {} device for {label}`

src/blockdev.rs
nikita-dubrovskii: agree. done

jlebon:
Gonna merge this one. This also fixes an issue where if you pass a `/dev/disk/by-X/...` disk path, it builds a partition path like `/dev/disk/by-X/${NAME}` which is wrong in this case. Using `--paths` prints the right path.

PR #216, opened 2020-04-16
install: change /boot/ directory for --copy-network

dustymabe:
In another code review [1] we agreed to make the name of the directory
more generic.

[1] https://github.com/coreos/fedora-coreos-config/pull/346#discussion_r409640904

PR #217, opened 2020-04-17
ci: Compress images once, use explicit scenario list

cgwalters:
This is done in the RHCOS pipeline and will be done soon in FCOS
https://github.com/coreos/fedora-coreos-pipeline/pull/218

Also switch to explicit scenario list as another workaround
for https://github.com/coreos/coreos-assembler/pull/1365


dustymabe:
merging this to unblock #216

PR #220, opened 2020-04-22
 Install as coreos-installer.legacy

cgwalters:
For https://github.com/openshift/enhancements/pull/210 I think
we will need at least some transitional period where RHCOS
ships both.  Even ignoring customers, the way our CI works
strongly encourages "ratcheting" changes in a compatible way,
i.e.:

- Introduce new thing, ship it
- Port consumers to new thing
- Remove old thing

And honestly I think this is a good model, because it forces
us to experience some of the pain that customers might feel here.

The basic idea here is:

- Install as `/usr/libexec/coreos-installer.legacy` so there's
  no chance of conflict with the real one
- Test for the presence of the "stamp file' we drop in our live initramfs
  so we know to defer to the real root's coreos-installer

This would be even more robust if we changed coreos-assembler
to inject an explicit stamp file into the legacy initramfs
(and ideally, we don't ship legacy installer in the initramfs
 but that's a bit trickier to do cleanly)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,5 +2,5 @@ all:
 	@echo "(Nothing to do)"
 
 install:
-	install -D -t $(DESTDIR)/usr/libexec -m 0755 coreos-installer
+	install -D -m 0755 coreos-installer $(DESTDIR)/usr/libexec/coreos-installer.legacy
 	for x in dracut/*; do install -D -t $(DESTDIR)/usr/lib/dracut/modules.d/$$(basename $$x) $$x/*; done
--- a/dracut/30coreos-installer/coreos-installer-generator
+++ b/dracut/30coreos-installer/coreos-installer-generator
@@ -4,6 +4,12 @@
 
 set -e
 
+# Verify we are in the legacy installer initramfs.
+# Requires https://github.com/coreos/coreos-assembler/pull/1389
+if ! [ -f /etc/coreos-legacy-installer-initramfs ] ; then
+  exit 0
+fi
+
 UNIT_DIR="${1:-/tmp}"
 
 cmdline=( $(</proc/cmdline) )
--- a/dracut/30coreos-installer/coreos-installer.service
+++ b/dracut/30coreos-installer/coreos-installer.service
@@ -6,7 +6,7 @@ OnFailureJobMode=isolate
 
 [Service]
 Type=simple
-ExecStart=/usr/libexec/coreos-installer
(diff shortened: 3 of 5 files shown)


cgwalters:
```
Successfully tested scenario:legacy-install for 45.81.202004222120-0 on bios (metal)
Successfully tested scenario:pxe-install for 45.81.202004222120-0 on bios (metal)
Successfully tested scenario:iso-install for 45.81.202004222120-0 on bios (metal)
````

All that testing work paid off!

**Comment by cgwalters** [comment]
Aaaah and of course *this* PR only worked with the code before
https://github.com/coreos/coreos-assembler/pull/1389#issuecomment-618506395

```
+if ! [ -f /etc/coreos-legacy-installer-initramfs ]; then
```

because I forgot to remove the `!` there so the two bugs cancelled out...

**Comment by cgwalters** [comment]
I'm still testing this.  I'd like to have https://github.com/coreos/coreos-assembler/pull/1389 merged and be able to rely on it.

**Comment by cgwalters** [comment]
OK so what I've been banging my head against is `kola testiso --scenarios pxe-install` doesn't work with RHCOS and it seems to be for some reason the `e1000` driver isn't being loaded in the initramfs.

**Comment by cgwalters** [comment]
>  and it seems to be for some reason the e1000 driver isn't being loaded in the initramfs.

No that was just an artifact of me using `rd.break=cmdline` to debug.  Oh god debugging in the initramfs sucks.

**Comment by cgwalters** [comment]
Aaargh somehow the PXE initrd is missing `/etc/initrd-release`?  Computers are cursed and we should just throw them all in the sea

**Comment by cgwalters** [comment]
OK nevermind:

```
switch_root:/# grep -rE initrd-release /usr
...
Binary file /usr/lib/systemd/libsystemd-shared-239.so matches
/usr/lib/dracut-lib.sh:    [ -f /usr/lib/initrd-release ] || return
/usr/lib/dracut/hooks/cmdline/90-parse-coreos.sh:    rm -f /etc/initrd-release
```

One of these things is...not right.

cgwalters:
OK, so we just need to skip `90-parse-coreos.sh` in the non-legacy case too.  With that, all scenarios work!

cgwalters:
> Though I think it's still worth renaming the generator and systemd name to -legacy so make it more obvious which is which?

I think so yes but I also carefully tested all of this together so I'll look at this as a followup.

PR #221, opened 2020-04-23
legacy: resolve symlinks for DEST_DEV

nikita-dubrovskii:
User may specify `dev/by-path/name` as `coreos.inst.install_dev`,
but on s390x with DASD this breaks the installation.

### DIFF
--- a/coreos-installer
+++ b/coreos-installer
@@ -904,6 +904,11 @@ main() {
     # by a few of the functions below.
     DEST_DEV=$(cat /tmp/selected_dev)
     DEST_DEV=/dev/$DEST_DEV
+    if [ -L "${DEST_DEV}" ]; then
+        REAL_DEV=$(realpath $DEST_DEV)
+        echo "$DEST_DEV --> $REAL_DEV"
+        DEST_DEV=$REAL_DEV
+    fi
 
     #import_gpg_key
     get_img_url


cgwalters:
Hmm, why does it break?  Which tool wants the real device path?

coreos-installer
```diff
@@ -904,6 +904,11 @@ main() {
     # by a few of the functions below.
     DEST_DEV=$(cat /tmp/selected_dev)
     DEST_DEV=/dev/$DEST_DEV
+    REAL_DEV=`readlink -n -f $DEST_DEV || echo -n $DEST_DEV`
```
cgwalters, line 907: How about:
```
if [ -L "${DEST_DEV}" ]; then
  DEST_DEV=$(realpath $DEST_DEV)
fi
```
?

nikita-dubrovskii:
> Hmm, why does it break? Which tool wants the real device path?

This week i got an internal IBM bug report, and the root cause:

`if [[ "${DEST_DEV}" =~ "dasd" ]]; then
            write_image_to_dasd`

coreos-installer
nikita-dubrovskii: Thx, good idea. The reason why i did it a little bit different - to `echo` a log message, which could be useful for user.

coreos-installer
```diff
@@ -904,6 +904,11 @@ main() {
     # by a few of the functions below.
     DEST_DEV=$(cat /tmp/selected_dev)
     DEST_DEV=/dev/$DEST_DEV
+    if [ -L "${DEST_DEV}" ]; then
+	REAL_DEV=$(realpath $DEST_DEV)
+	echo "$DEST_DEV --> $REAL_DEV"
+	DEST_DEV=$REAL_DEV
```
jlebon, line 910: Minor: looks like tabs here instead of spaces? Madness!

nikita-dubrovskii: let me check, sounds crazy )

nikita-dubrovskii: emacs on test-machine had no config. fixed

PR #222, opened 2020-04-23
ci: fix building metal image

jlebon:
We were doing `cosa build metal` instead of `cosa buildextend-metal`.
And since `fcosBuild` itself already builds a qemu image, we were
essentially doing `create_disk.sh` twice.

Instead, just make the initial `fcosBuild` only build the metal image
since we don't need the QEMU image for our tests. And also skip the
regular kola tests since those don't really test coreos-installer
anyway (and they'd need the QEMU image).

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -3,10 +3,11 @@
 cosaPod(buildroot: true) {
     checkout scm
 
-    fcosBuild(make: true)
+    // we don't need the qemu image to test coreos-installer; just the OSTree
+    fcosBuild(make: true, skipKola: true, extraArgs: 'ostree')
 
     stage("Build metal+live") {
-        shwrap("cd /srv/fcos && cosa build metal")
+        shwrap("cd /srv/fcos && cosa buildextend-metal")
         shwrap("cd /srv/fcos && cosa buildextend-metal4k")
         // Compress once
         shwrap("cd /srv/fcos && cosa buildextend-live")


jlebon:
Requires: https://github.com/coreos/coreos-ci-lib/pull/25

jlebon:
> And since `fcosBuild` itself already builds a qemu image, we were
essentially doing `create_disk.sh` twice.

This is actually even worse right now because `fcosBuild` does `--strict`, but we didn't do it here, so on the second run from `cosa build metal`, it would try to pull in libbrotli-1.0.7-10.fc32 from the pool which would have caused a full other rpm-ostree compose too. (Noticed this in the CI results for https://github.com/coreos/coreos-installer/pull/187).

Aside: this is what https://github.com/coreos/rpm-ostree/pull/2059 fixes.

dustymabe:
hmm this looks more confusing too me.

why not leave the `buildextend-metal`, but just make the `build` just build the ostree (no disks).

jlebon:
> why not leave the buildextend-metal, but just make the build just build the ostree (no disks).

Hmm, we could do that; though I don't follow how one is less confusing than the other. Can you expand? Note also by using `fcosBuild` we get `--strict` automatically for example.

dustymabe:
I do propose to still use `fcosBuild` but make a separate call for `buildextend-metal` just like we have for `metal4k`. For example:

```diff
diff --git a/.cci.jenkinsfile b/.cci.jenkinsfile
index 7f27f5c..a0a7c1e 100644
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -3,10 +3,11 @@
 cosaPod(buildroot: true) {
     checkout scm
 
-    fcosBuild(make: true)
+    # just `cosa build` the ostree
+    fcosBuild(make: true, skipKola: true, extraArgs: 'ostree')
 
     stage("Build metal+live") {
-        shwrap("cd /srv/fcos && cosa build metal")
+        shwrap("cd /srv/fcos && cosa buildextend-metal")
         shwrap("cd /srv/fcos && cosa buildextend-metal4k")
         // Compress once
         shwrap("cd /srv/fcos && cosa buildextend-live")
```

This just makes more sense to me when I read it, but either one works so I'll let you decide.

cgwalters:
I like Dusty's suggestion too but obviously either works.

jlebon:
OK, done! :arrow_up:

Issue #227, opened 2020-04-30
network karg parsing doesn't work for args without values


dustymabe:
For example I provide this set of kargs:

```
coreos.no_persist_ip
ipv6.disable
ip=dhcp
coreos.inst.install_dev=/dev/sda
coreos.inst.ignition_url=http://192.168.122.1:8000/remote-file.ign
coreos.inst.insecure
coreos.inst.stream=stable
coreos.inst.skip_reboot
console=ttyS0
```

I end up with this in the output:

```
         Starting CoreOS Installer...                                                                                 
######################################################################## 100.0%
[   12.374953] coreos-installer-service[885]: coreos-installer install /dev/sda --ignition /tmp/coreos-installer-4ohi4Y --firstboot-args rd.neednet=1 ipv6.disable=ipv6.disable ip=dhcp coreos.no_persist_ip=coreos.no_persist_ip  --stream stable --insecure
[   12.406566] coreos-installer-service[885]: Downloading stable image (raw.xz) and signature
[   14.254554] coreos-installer-service[885]: Read disk 8.1 MiB/461.6 MiB (1%)
```

and the resulting install has:

```
[root@localhost ~]# mount /dev/sda1 /mnt/
[root@localhost ~]# cat /mnt/ignition 
cat: /mnt/ignition: Is a directory
[root@localhost ~]# cat /mnt/ignition
ignition/           ignition.firstboot  
[root@localhost ~]# cat /mnt/ignition.firstboot 
set ignition_network_kcmdline="rd.neednet=1 ipv6.disable=ipv6.disable ip=dhcp coreos.no_persist_ip=coreos.no_persist_ip "
```

So we end up with `ipv6.disable=ipv6.disable` and `coreos.no_persist_ip=coreos.no_persist_ip` getting passed.

```
[root@localhost ~]# rpm -q coreos-installer
coreos-installer-0.1.3-1.fc31.x86_64
```

dustymabe:
working on a fix for his now.

Issue #156, opened 2020-02-12
coreos-installer-service doesn't support multiple ip kargs for firstboot


sebastienGuavus:
# Bug #

## Host Operating System Version ##
fedora-coreos-31

## Target Operating System Version ##
fedora-coreos-31

## coreos-installer Version ##
coreos-installer 0.1.2

## Expected Behavior ##
If I provide multiple ip kargs all are save for firstboot

## Actual Behavior ##
Only last ip kargs is saved

## Reproduction Steps ##

  1. ...
  2. ...

## Other Information ##

This apply to all others kargs in PERSIST_DRACUT_NET_PARAMS that could reasonably be specified more than once.

```bash
karg() {
    local name="$1" value=""
    for arg in "${cmdline[@]}"; do
        if [[ "${arg%%=*}" == "${name}" ]]; then
            if [[ -z ${value} ]]; then
                value="${arg#*=}"
            else
	        value="${value} ${arg#*=}"
            fi		
        fi
    done
    echo "${value}"
}
```

bgilbert:
Thanks for the report.  This is a regression from the legacy branch.

sebastienGuavus:
The snippet above doesn't work. Should be more like
```
    else
       value="${value} {$name}=${arg#*=}"
    fi
```
But then it could cause issue for karg_bool or if any of the coreos.inst.* karg are pass more than once.
Maybe a kargs function ?

PR #228, opened 2020-04-30
systemd: coreos-installer-service: fix network karg parsing

dustymabe:
This commit fixes two issues with network karg parsing. First,
there are some arguments that can be provided more than once (like
ip=). Second, there are arguments that are boolean args that don't
require a value, which aren't currently getting parsed correctly.

Fixes: https://github.com/coreos/coreos-installer/issues/156
Fixes: https://github.com/coreos/coreos-installer/issues/227

### DIFF
--- a/systemd/coreos-installer-service
+++ b/systemd/coreos-installer-service
@@ -63,14 +63,15 @@ if [ -n "${ignition_url}" -a "${ignition_url}" != "skip" ]; then
     args+=("--ignition" "${ignition_file}")
 fi
 
-# First-boot kernel arguments
+# First-boot kernel arguments. Filter out any args that aren't in our
+# list of networking arguments to forward and store them in $firstboot_args
 firstboot_args=""
-for param in "${PERSIST_KERNEL_NET_PARAMS[@]}" "${PERSIST_DRACUT_NET_PARAMS[@]}"
-do
-    value=$(karg "${param}")
-    if [ -n "${value}" ]; then
-        firstboot_args+="${param}=${value} "
-    fi
+for item in "${cmdline[@]}"; do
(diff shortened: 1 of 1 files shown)


cgwalters:
Only skimmed and seems OK but...we should get a shared shell script file for karg parsing that we can at least copy/paste, ideally include.  (Maybe just use dracut's)
(or of course, don't write this stuff in shell...)

dustymabe:
> Only skimmed and seems OK but...we should get a shared shell script file for karg parsing that we can at least copy/paste, ideally include. (Maybe just use dracut's)
> (or of course, don't write this stuff in shell...)

I agree - we do have arg parsing sprinkled in a lot of places. I think in general we do use dracut's when we are running in the initrd, but I haven't seen anyone using it in the real root.

dustymabe:
ok tested this. With

```
coreos.no_persist_ip
ipv6.disable
ip=eth0:dhcp
ip=eth1:dhcp
coreos.inst.install_dev=/dev/sda
coreos.inst.ignition_url=http://192.168.122.1:8000/remote-file.ign
coreos.inst.insecure
console=ttyS0
```

I see
```
         Starting CoreOS Installer...
######################################################################## 100.0%
[   12.031573] coreos-installer-service[887]: coreos-installer install /dev/sda --ignition /tmp/coreos-installer-PgwRwl --firstboot-args rd.neednet=1 coreos.no_persist_ip ipv6.disable ip=eth0:dhcp ip=eth1:dhcp  --insecure
[   12.066714] coreos-installer-service[887]: Installing Fedora CoreOS 31.20200430.dev.1 x86_64 (512-byte sectors)
```

and can confirm `coreos.no_persist_ip ipv6.disable ip=eth0:dhcp ip=eth1:dhcp` all make it to the next boot.

PR #229, opened 2020-04-30
systemd: coreos-installer-service: persist net.ifnames net.naming-scheme kargs

dustymabe:
More networking kargs we should probably persist. This came from
some user feedback about RHCOS missing this ability.

### DIFF
--- a/systemd/coreos-installer-service
+++ b/systemd/coreos-installer-service
@@ -3,7 +3,7 @@
 set -e
 
 # Kernel networking params to persist
-PERSIST_KERNEL_NET_PARAMS=("ipv6.disable")
+PERSIST_KERNEL_NET_PARAMS=("ipv6.disable" "net.ifnames" "net.naming-scheme")
 
 # Dracut networking params to persist
 # Everything other than rd.neednet.


dustymabe:
> This came from some user feedback about RHCOS missing this ability.

https://twitter.com/Arrfab/status/1255081239117205504

cgwalters:
> https://twitter.com/Arrfab/status/1255081239117205504

I see you're naming names here.

arrfab:
The corner case was for OCP4 deployment behind ci.centos.org (to replace and old openshift 3.6 setup). We were used to deploy all our CentOS nodes with net.ifnames=0 and so use the "traditional" eth0/1 etc naming convention (as we know which interface is connected to which chassis backplane/switch.
That's how I realized that through pxe kernel parament net.ifnames=0 and with then ip= with eth0 was working during install, when rhcos was rebooting, node didn't come back online. 
Removing net.ifnames=0 and specifying the "new style" name (one has to find it first, and it depends on hardware/chassis/etc) node was able to get network correctly configured at provisioning/pxe boot time and also when rebooting on rhcos.

So we're not blocked as we implemented in our ansible jinja2 template for pxe config the correct option. But maybe good to still support kernel parameter like a RHEL would ?

jlebon:
> The corner case was for OCP4 deployment behind ci.centos.org (to replace and old openshift 3.6 setup).

Out of curiosity, are you referring to https://console.apps.ci.centos.org:8443/? Because that is highly relevant to us :)  (See e.g. https://github.com/coreos/fedora-coreos-pipeline and https://github.com/coreos/coreos-ci).

PR #231, opened 2020-04-30
Release 0.2.0

bgilbert:
Major changes:

- By default, install from an offline image shipped with the running system, if available
- When installing from a Fedora CoreOS stream, automatically select 4Kn image if needed
- Add `--copy-network` and `--network-dir` to copy network configs from the running system
- Add `--ignition-hash` to verify the hash of the specified Ignition config

Minor changes:

- Fix `coreos.inst` forwarding of repeated network kargs
- Fix `coreos.inst` forwarding of network kargs without a value
- Get block device path from `lsblk` rather than constructing it
- Stop overriding locked root account when launching emergency shell
- Redirect systemd generator's stdout to kmsg

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,7 +159,7 @@ checksum = "b3a71ab494c0b5b860bdc8407ae08978052417070c2ced38573a9157ad75b8ac"
 
 [[package]]
 name = "coreos-installer"
-version = "0.2.0"
+version = "0.2.1-alpha.0"
 dependencies = [
  "bincode",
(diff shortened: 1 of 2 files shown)


dustymabe:
the release checklist does say: ` make sure the resulting PR contains exactly two commits`. Should we update it?

bgilbert:
That instruction is correct if we're only bumping the Z version.  The extra commit is needed because we're bumping Y.  The instructions don't cover that case.

PR #234, opened 2020-05-06
fix coreos-installer dracut module

dustymabe:
dracut hooks are sourced which means if we exit 0 here then we'll
stop parsing the rest of the hooks (after 90).


cgwalters:
Shell script is the worst

dustymabe:
ok I confirmed this fixes the problem I was seeing where networking wasn't coming up in the initrd for a RHCOS image because the cmdline hooks was exiting early because of this.

@cgwalters - I won't be around tomorrow. If possible could you test the "legacy installer" path (maybe CI does that for us now?) to make sure this doesn't break anything there and then merge this?

PR #236, opened 2020-05-08
ci: stop passing --osmet switch to buildextend-live

jlebon:
That's so 2 days ago!

https://github.com/coreos/coreos-assembler/pull/1418


cgwalters:
If we'd deprecated it, how long would it have taken us to notice that this CI was still passing it unnecessarily?  Hard to say...

PR #243, opened 2020-05-20
signing-keys: add Fedora 32 key

bgilbert:


dustymabe:
> Might as well add f33 too while we're at it?

I think let's hold on that. I think it would be prudent to keep the f33 key out and make sure that our procedural docuements for rebasing `next` on the next Fedora major are correct. If we add the f33 key now we won't know our procedures are wrong until f34

dustymabe:
LGTM - should we drop the f30 key?

bgilbert:
Dropping the F30 key would prevent installing older `testing` releases.  Do we want to do that yet?

dustymabe:
> Dropping the F30 key would prevent installing older `testing` releases. Do we want to do that yet?

Ehh feels like a good way to make people move forward. `30.20191014.1` was the last testing release. They can still use old coreos-installer to install it. I'm not sure how strong the guarantee of using "2 major releases ago" coreos-installer to install today's FCOS should be. 

@jlebon, WDYT?

jlebon:
> I'm not sure how strong the guarantee of using "2 major releases ago" coreos-installer to install today's FCOS should be.

I think it's the reverse, no? Using today's coreos-installer to install a "2 major releases ago" FCOS.

But yeah, I agree we probably shouldn't support that. There are a bunch of switches that just won't work correctly without support on the installed system (like the `--copy-network` stuff).

Refining further: I think coreos-installer should only support installing matching or newer releases? (That said, osmet is a big carrot here to make sure people mostly stick with matching versions.)

bgilbert:
> Refining further: I think coreos-installer should only support installing matching or newer releases?

In practice, that'd just mean that people stuck on old releases would also be stuck on an old installer.  I don't think we need an explicit rule about supported OS versions, but if new features don't work with old releases, I don't think that's too surprising.

bgilbert:
Dropped the F30 key.

cgwalters:
It's kind of tempting to make the build process require `distribution-gpg-keys` or so.  But eh, that can come later.

bgilbert:
@cgwalters That would present problems for people doing `cargo install` on e.g. Debian.

dustymabe:
can we get this into a new release (or at least backported into a rpm build) so we can include it in the next round of releases we do next week?

PR #251, opened 2020-05-27
pack: Support --real-rootdev for RHCOS+LUKS

cgwalters:
Pairs with: coreos/coreos-assembler#1483

We want RHCOS to support offline installs too (it's key to the UX)
but the LUKS container complicates things.  We discussed changing
the RHCOS LUKS bits to save the header in e.g. `/boot` but it's
a big change to that code, and introduces yet another layout
when we really want to switch to Ignition+LUKS.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -78,11 +78,17 @@ pub struct OsmetFiemapConfig {
     pub file: String,
 }
 
+pub struct OsmetRootBlkDevReal {
+    pub underlying_device: String,
+    pub offset_sectors: u32,
+}
+
 pub struct OsmetPackConfig {
     pub output: String,
     pub device: String,
     pub checksum: String,
     pub description: String,
+    pub rootdev: Option<OsmetRootBlkDevReal>,
 }
 
 pub struct OsmetUnpackConfig {
@@ -416,6 +422,13 @@ pub fn parse_args() -> Result<Config> {
                                 .help("Description of OS")
                                 .takes_value(true),
                         )
+                        .arg(
+                            Arg::with_name("real-rootdev")
+                                .long("real-rootdev")
+                                .value_name("PATH,OFFSET")
+                                .help("Underlying device for e.g. RHCOS LUKS container; /dev/disk/by-label/root should be mountable")
+                                .takes_value(true),
+                        )
                         // positional args
                         .arg(
                             Arg::with_name("device")
@@ -669,6 +682,23 @@ fn parse_iso_remove(matches: &ArgMatches) -> Result<Config> {
     }))
 }
 
+fn parse_real_rootdev<T: AsRef<str>>(s: Option<T>) -> Result<Option<OsmetRootBlkDevReal>> {
+    if let Some(v) = s {
+        let v = v.as_ref();
+        let parts: Vec<_> = v.splitn(2, ',').collect();
+        if parts.len() < 2 {
+            bail!("Expected DEVICE,OFFSET-SECTORS but found {}", v);
+        }
+        let offset = parts[1].parse()?;
+        Ok(Some(OsmetRootBlkDevReal {
+            underlying_device: parts[0].to_string(),
+            offset_sectors: offset,
+        }))
+    } else {
+        Ok(None)
+    }
+}
+
 fn parse_osmet_pack(matches: &ArgMatches) -> Result<Config> {
     Ok(Config::OsmetPack(OsmetPackConfig {
         output: matches
@@ -687,6 +717,7 @@ fn parse_osmet_pack(matches: &ArgMatches) -> Result<Config> {
             .value_of("description")
             .map(String::from)
             .expect("description missing"),
+        rootdev: parse_real_rootdev(matches.value_of("real-rootdev"))?,
     }))
 }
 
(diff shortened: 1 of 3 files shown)


cgwalters:
Current status is that this crashes with a checksum mismatch.  Debugging this is fun! (any tips appreciated)

My notes:

- We need to copy the LUKS header, so I think having `OsmetPartition` be the full partition is right
- I *think* the fiemap offsets are all relative to the dm-linear device's start, but I need to verify this

jlebon:
> I think the fiemap offsets are all relative to the dm-linear device's start, but I need to verify this

Yes, this is correct. We need the data offset of the LUKS container (`cryptsetup luksDump` prints this, but there's probably a cleaner way to query it), and add that to the fiemap offsets to have it be relative to the partition start. See https://github.com/coreos/coreos-installer/pull/187#issuecomment-619208591.

src/cmdline.rs
```diff
@@ -416,6 +422,13 @@ pub fn parse_args() -> Result<Config> {
                                 .help("Description of OS")
                                 .takes_value(true),
                         )
+                        .arg(
+                            Arg::with_name("real-rootdev")
```
jlebon, line 426: Hmm, why not just have coreos-installer auto-detect this and open up the LUKS container itself?

jlebon: Ahh OK I think I follow now. You already have the data offset from `rhcos_luks_header_size=32768`. Neat.

Hmm, is that offset in sector size? I think it might need a `* 512` before adding to the fiemap offsets.

cgwalters: Dunno; setting up device-mapper feels like it's something that should live externally?  Dunno.

cgwalters: > Hmm, is that offset in sector size? I think it might need a * 512 before adding to the fiemap offsets.

Ooh yes!  Of course.  And that matches up with
```
Data segments:
  0: crypt
	offset: 16777216 [bytes]
```

Got a successful osmet checksum match, doing a full `cosa buildextend-live` now!

cgwalters:
```
Successfully tested scenario iso-offline-install for 46.82.202005261954-0 on bios (metal)
```

:tada:

src/cmdline.rs
```diff
@@ -687,7 +714,8 @@ fn parse_osmet_pack(matches: &ArgMatches) -> Result<Config> {
             .value_of("description")
             .map(String::from)
             .expect("description missing"),
-    }))
+        rootdev: parse_real_rootdev(matches.value_of("real-rootdev"))?,
+        }))
```
jlebon, line 718: Looks like unintended indentation here. Missing a final `cargo fmt` perhaps?

src/osmet/mod.rs
```diff
@@ -188,10 +188,19 @@ pub fn find_matching_osmet_in_dir(
 
 fn scan_root_partition(
     root: &Mount,
+    mapping: Option<&OsmetRootBlkDevReal>,
```
jlebon, line 191: Bikeshed: How about `real_dev`? "Mapping" already means something in this code.

src/osmet/mod.rs
```diff
@@ -188,10 +188,19 @@ pub fn find_matching_osmet_in_dir(
...
     mut boot_files: HashMap<u64, PathBuf>,
 ) -> Result<(OsmetPartition, HashMap<PathBuf, Sha256Digest>)> {
     // query the trivial stuff first
-    let (start_offset, end_offset) = root.get_partition_offsets()?;
+    let ((start_offset, end_offset), offset) = if let Some(mapping) = mapping {
+        let bdev = BlkDev { path: mapping.underlying_device.clone(), ..Default::default() };
+        let (start_offset, end_offset) = bdev.get_partition_offsets()?;
+        let offset = mapping.offset_bytes.checked_mul(512)
```
jlebon, line 198: Hmm, does this need to actually be `* $sector_size`? Or (like partition offsets), is it always in 512 blocks? We can punt on investigating this until we actually plumb through osmet for metal4k images.

src/osmet/mod.rs
```diff
@@ -225,7 +234,12 @@ fn scan_root_partition(
         let object = object_path_to_checksum(entry.path())
             .chain_err(|| format!("invalid object path {:?}", entry.path()))?;
 
-        for extent in extents {
+        for mut extent in extents {
+            if let Some(offset) = offset {
+            extent.physical = extent.physical.checked_add(offset.into())
```
jlebon, line 239: Missing indent here.

src/osmet/mod.rs
cgwalters: Good question; yeah, I think we should defer this until we've gone through the metal4k exercise.
I wouldn't be surprised if for 4k we also had to change the LUKS bits in RHCOS too.

src/cmdline.rs
cgwalters: Random aside, a full 16MiB for the LUKS header seems *enormous*...not sure what's going on there.  That was one of the things that made me think the value was in bytes.

cgwalters:
One argument to have more "heavy lifting" in the supermin code in coreos-assembler is that it's code that isn't shipped in the main `coreos-installer` binary people download.  An argument against that is that it's (currently) shell script which there's way, way, way too much of in my life...

jlebon:
> One argument to have the "preparation" in the supermin code in coreos-assembler is that it's code that isn't shipped in the main coreos-installer binary people download. An argument against that is that it's (currently) shell script which there's way, way, way too much of in my life...

I think it's fine as is. We'll rip this out anyway once we drop the baked LUKS container.

PR #257, opened 2020-05-30
release-checklist: some cleanups

bgilbert:
@lucab, thoughts on the second commit?


lucab:
@bgilbert I'm not to opposed to that, it's currently overly verbose.
The underlying idea was: "make sure you are cargo-publishing the right version", because it's easy to miss the tag-checkout and publish the alpha version instead.

Perhaps we can re-order that list to read like the following?
 * checkout the tag
 * verify version match in `grep '^version' Cargo.toml`
 * cargo publish
 * push git tag

bgilbert:
Updated!

PR #267, opened 2020-06-11
ci: check line lengths of help text

bgilbert:
Manually enumerate subcommands, since there doesn't seem to be a great way to do it programmatically.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -3,20 +3,54 @@
 cosaPod(buildroot: true) {
     checkout scm
 
-    // we don't need the qemu image to test coreos-installer; just the OSTree
-    fcosBuild(make: true, skipKola: true, extraArgs: 'ostree')
+    shwrap("make")
 
-    stage("Build metal+live") {
-        shwrap("cd /srv/fcos && cosa buildextend-metal")
-        shwrap("cd /srv/fcos && cosa buildextend-metal4k")
-        // Compress once
-        shwrap("cd /srv/fcos && cosa buildextend-live")
-        shwrap("cd /srv/fcos && cosa compress --artifact=metal --artifact=metal4k")
-    }
-    stage("Test ISO") {
-        // No need to run the iso-live-login scenario (in theory, and also right
-        // now it's buggy)
-        shwrap("cd /srv/fcos && kola testiso -S --scenarios pxe-install,iso-install,iso-offline-install")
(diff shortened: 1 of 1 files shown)


.cci.jenkinsfile
```diff
@@ -6,6 +6,26 @@ cosaPod(buildroot: true) {
     // we don't need the qemu image to test coreos-installer; just the OSTree
     fcosBuild(make: true, skipKola: true, extraArgs: 'ostree')
 
+    stage("Check help") {
```
jlebon, line 9: I'd put this before `fcosBuild`; it's silly to build the full OS if we fail right after because of line length. That would involve calling `make` upfront, and then you can remove `make: true` from `fcosBuild()`.

(Even better would be doing it in parallel so that we get feedback on both branches in a single run.)

.cci.jenkinsfile
```diff
@@ -3,20 +3,56 @@
...
+                }
+                checklen
+                checklen install
+                checklen download
+                checklen list-stream
+                checklen iso
+                checklen iso embed
+                checklen iso show
+                checklen iso remove
+                if [ "${fail}" = 1 ]; then
```
jlebon, line 31: Hmm, I think we need to use `${fail:-}` or otherwise initialize it to `fail=0` at the start.

jlebon: (Because otherwise I think `set -u` will trip on this in the happy case.)

bgilbert: Yup, I'm not used to `set -u`.

.cci.jenkinsfile
```diff
@@ -3,20 +3,57 @@
...
-        // now it's buggy)
-        shwrap("cd /srv/fcos && kola testiso -S --scenarios pxe-install,iso-install,iso-offline-install")
-        shwrap("cd /srv/fcos && kola testiso -S --scenarios iso-install --qemu-native-4k")
+    // so cargo doesn't use /.cargo, which it can't access
+    withEnv(["CARGO_HOME=${env.WORKSPACE}/.cargo"]) {
+        shwrap("make")
     }
+
+    parallel(
+        lint: stage("Lint") {
```
jlebon, line 12:
```suggestion
        lint: { stage("Lint")
```

cgwalters:
We certainly had to go to great lengths to test this!

PR #268, opened 2020-06-11
install: add --delete-karg and --append-karg

jlebon:
This allows users to delete and append kernel arguments from the default
set. The major use cases for this are (1) avoiding an additional reboot
which might be costly in some situations, and (2) making it easier to
troubleshoot the system on firstboot.

For more details, see
https://github.com/coreos/fedora-coreos-tracker/issues/533#issuecomment-642234015.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/533

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -152,6 +152,8 @@ fn write_disk(config: &InstallConfig, source: &mut ImageSource, dest: &mut File)
     // postprocess
     if config.ignition.is_some()
         || config.firstboot_kargs.is_some()
+        || config.append_kargs.is_some()
+        || config.delete_kargs.is_some()
         || config.platform.is_some()
         || config.network_config.is_some()
     {
@@ -163,6 +165,18 @@ fn write_disk(config: &InstallConfig, source: &mut ImageSource, dest: &mut File)
         if let Some(firstboot_kargs) = config.firstboot_kargs.as_ref() {
             write_firstboot_kargs(mount.mountpoint(), firstboot_kargs)
                 .chain_err(|| "writing firstboot kargs")?;
+        }
+        if config.append_kargs.is_some() || config.delete_kargs.is_some() {
+            eprintln!("Modifying kernel arguments");
+
+            edit_bls_entries(mount.mountpoint(), |orig_contents: &str| {
+                bls_entry_delete_and_append_kargs(
+                    orig_contents,
+                    config.delete_kargs.as_ref(),
+                    config.append_kargs.as_ref(),
+                )
+            })
+            .chain_err(|| "deleting and appending kargs")?;
         }
         if let Some(platform) = config.platform.as_ref() {
             write_platform(mount.mountpoint(), platform).chain_err(|| "writing platform ID")?;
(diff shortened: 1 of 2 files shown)


src/install.rs
```diff
@@ -254,6 +259,31 @@ fn write_firstboot_kargs(mountpoint: &Path, args: &str) -> Result<()> {
     Ok(())
 }
 
+/// Write persistent kernel arguments.
```
cgwalters, line 262: One thing we need to keep in mind is this needs to trigger a rerun of `zipl` on s390x.

Hmm...I am increasingly thinking what we want is like a special systemd unit for zipl that checks the mtime of the BLS config and runs `zipl` if it's changed or something like that.

brokenjacobs:
Is there also a way to remove persistent arguments as well? (console=ttyS0,1152008n1).

It's a good idea in my opinion to emulate the various arguments to rpm-ostree kargs. They are all there for a reason.

lucab:
I like this but also have a bad feeling about it; I fear this is going to trip many people.

In particular, it looks like the logic implemented here is really an `--append-persistent-kargs`. Also, it implicitly assumes that users already know what are the current kargs (is that a static set?) and that previous values can be overridden in a last-key-wins ways (I know that's not the case generally, see for example hugepages).

src/install.rs
nikita-dubrovskii: Indeed, after installation we do write some s390x specific args parsed from `/proc/cmdline`

src/install.rs
```diff
@@ -254,6 +259,31 @@ fn write_firstboot_kargs(mountpoint: &Path, args: &str) -> Result<()> {
...
+fn write_persistent_kargs(mountpoint: &Path, args: &str) -> Result<()> {
+    eprintln!("Writing persistent kernel arguments");
+
+    edit_bls_entries(mountpoint, |orig_contents: String| {
```
lucab, line 266: For this and the other similar closure below, what about moving it out of this scope instead? I'd be really happier to see both covered with a few positive and negative unit tests.

jlebon:
> In particular, it looks like the logic implemented here is really an `--append-persistent-kargs`.

Yeah, see https://github.com/coreos/fedora-coreos-tracker/issues/533#issuecomment-644319795 related to this. Will rename the switch.

> Also, it implicitly assumes that users already know what are the current kargs (is that a static set?)

Hmm, though that's no different from eventual kargs-via-Ignition support, right?

>  and that previous values can be overridden in a last-key-wins ways (I know that's not the case generally, see for example hugepages).

Yes, agreed (turns out `console` is another such example as mentioned in the linked comment above). Combined with `--delete-persistent-kargs`, I think it provides enough flexibility.

(See also https://github.com/ostreedev/ostree/issues/1859 -- we'll need to make sure that the implementation of kargs-via-Ignition is ordering aware).

jlebon:
OK, split out prep patches in https://github.com/coreos/coreos-installer/pull/271!

src/install.rs
jlebon: This is part of https://github.com/coreos/coreos-installer/pull/271.

src/install.rs
jlebon: Ahh yup, goot catch. Hmm, let's circle back on this once https://github.com/coreos/coreos-installer/pull/153 merges? It'll make it easier to run `zipl`. I can open an issue once this merges.

nikita-dubrovskii: Sorry, haven't seen this message begore. So, i've already started working on patch to remove all logic in #153 `zipl.rs` in favour of using `--append-karg`.

PR #271, opened 2020-06-16
Prep patches for modifying kargs

jlebon:
```
commit b7e7a01b6d308fe7b85751cefc1f31cd3793db17
Date:   Thu Jun 11 11:41:59 2020 -0400

    install: factor out function to walk BLS configs

    Prep for next patch.

commit 25c71d62f937d52970f85be2294b4f166983d908
Date:   Tue Jun 16 16:38:46 2020 -0400

    install: rename --firstboot-args to --append-firstboot-kargs

    Let's use a more precise name for this switch. First, specify that it
    *appends* to the list of firstboot arguments. Second, use `kargs`
    instead of just `args` to be more clear.

    This is prep for new switches which will use the same naming convention.

commit 70371c47ab76b78f522e4469906e05e1edd05c6d
Date:   Tue Jun 16 17:15:41 2020 -0400

    install: test platform ID restamping bits

    Factor out the closure and add some basic unit tests.
```

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -246,21 +246,24 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
 
     eprintln!("Setting platform to {}", platform);
     edit_bls_entries(mountpoint, |orig_contents: &str| {
-        // Rewrite the config.  Assume that we will only install
-        // from metal images and that their bootloader configs will
-        // always set ignition.platform.id.  Fail if those
-        // assumptions change.  This is deliberately simplistic.
-        let new_contents = orig_contents.replace(
-            "ignition.platform.id=metal",
-            &format!("ignition.platform.id={}", platform),
-        );
-        if orig_contents == new_contents {
-            bail!("Couldn't locate platform ID");
-        }
-        Ok(new_contents)
+        bls_entry_write_platform(orig_contents, platform)
     })?;
 
     Ok(())
+}
+
+/// Modifies the BLS config, only changing the `ignition.platform.id`. This assumes that we will
+/// only install from metal images and that the bootloader configs will always set
+/// ignition.platform.id.  Fail if those assumptions change.  This is deliberately simplistic.
+fn bls_entry_write_platform(orig_contents: &str, platform: &str) -> Result<String> {
+    let new_contents = orig_contents.replace(
+        "ignition.platform.id=metal",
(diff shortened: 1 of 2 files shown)


jlebon:
https://github.com/coreos/coreos-ci-lib/commit/853d5fdef3153b2b03c19218a00d516d2b59f853

src/install.rs
```diff
@@ -245,7 +245,30 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
...
 
     eprintln!("Setting platform to {}", platform);
+    edit_bls_entries(mountpoint, |orig_contents: String| {
+        write_platform_in_bls(&orig_contents, platform)
+    })?;
+
+    Ok(())
+}
+
+fn write_platform_in_bls(orig_contents: &str, platform: &str) -> Result<String> {
```
lucab, line 255: `bls_entry_replace_platform()` maybe? Also, I think you can move portions of the inner comment to a docstring here.

lucab:
Code looks ok, I can stamp this later on if @bgilbert is fine with the new CLI flag arrangement.

src/install.rs
jlebon: I went with `bls_entry_write_platform` to match the name of the function `write_platform`. Will update https://github.com/coreos/coreos-installer/pull/268 too to match that naming convention.

jlebon:
Ahhh the rename to `--append-firstboot-kargs` shifts all the help strings to the right and we now go over 80 chars. And there doesn't seem to be a way to tell clap to put the help string on the next line *and* still keep the help string for all other options unindented.

I guess we can keep it as `--firstboot-args`, though I really wanted some consistency with `--append-karg` and `--delete-karg`.

jlebon:
Actually on that subject, couldn't we just "de-emphasize" (if not deprecate) and hide `--firstboot-args` now? The primary use case for this was network configs, and those are now much better covered by `--copy-network`, right?

dustymabe:
> Actually on that subject, couldn't we just "de-emphasize" (if not deprecate) and hide `--firstboot-args` now? The primary use case for this was network configs, and those are now much better covered by `--copy-network`, right?

Maybe we could `de-emphasize`, probably not deprecate at least the underlying code path as `--copy-network` won't apply for the "automation via installer kargs" path.

jlebon:
OK, I updated this now to hide `--firstboot-args`!

PR #272, opened 2020-06-17
legacy: include additional kargs when running zipl for firstboot

Prashanth684:
When playing around with some additional arguments noticed that they were not being propagated
on first boot for s390x. Included that to the list of args when running zipl.


Prashanth684:
> This LGTM, but note all focus is on the new installer now. We're switching to that in 4.6. IOW, should do this fix on master too, but I think #153 covers this already. :)

thanks @jlebon. good to know..i was testing 4.6 rhcos and it was still using the legacy installer. I did take a cursory look at #153 and it should be fixed, but i will test it out.

bgilbert:

PR #274, opened 2020-06-19
systemd: coreos-installer-service: Add support for persisting coreos.autologin additional arg

Prashanth684:
Added support for an autologin cmdline arg which would be really useful to debug cases where
the ability to login to the machine through ssh is broken due to some networking issue. This is
useful for debugging certain cases like OCP UPI deploys as we don't have to tinker with
the ignition config. The command line arg can be added and the system rebooted to autologin
and diagnose the issue.

xref: https://github.com/coreos/ignition-dracut/pull/195


lucab:
We decided not to support autologin kargs in https://github.com/coreos/fedora-coreos-tracker/issues/112. Additionally, this introduces a general persistent security bypass for what is supposed to be a temporary debugging hack for a very specific case.

jlebon:
For completeness, do note from https://github.com/coreos/fedora-coreos-tracker/issues/112 that we do have support for `single`: https://docs.fedoraproject.org/en-US/fedora-coreos/access-recovery/.

Issue #261, opened 2020-06-02
/dev/sda device is in use in bare-metal server


majorinche:
try to install fcos by pxe in bare-metal server, but meet the error, like

_checking for exclusive access to /dev/sda
caused by: couldn't reread partition table: device is in use_

other information:
pxelinux.cfg is:
DEFAULT pxeboot
TIMEOUT 20
PROMPT 0
LABEL pxeboot
    KERNEL fedora-coreos-31.20200505.3.0-live-kernel-x86_64
    APPEND ip=dhcp rd.neednet=1 initrd=fedora-coreos-31.20200505.3.0-live-initramfs.x86_64.img coreos.inst=yes coreos.inst.ignition_url=http://10.30.0.109/example.ign coreos.inst.image_url=http://10.30.0.109/fedora-coreos-31.20200505.3.0-metal.x86_64.raw.xz coreos.inst.install_dev=/dev/sda coreos.inst.insecure=yes
IPAPPEND 2

2goobers:
Same issue with iso boot live image.
Both fedora-coreos-31.20200517.3.0-live.x86_64.iso &
fedora-coreos-32.20200601.3.0-live.x86_64.iso
using CLI to initiate, and kernel line args

jlebon:
Can one of you print the full output from `coreos-installer`? Specifically, does this error happen at the end of the install or at the start?

2goobers:
The full output I get is this:

Error: checking for exclusive access to /dev/sda
Caused by: couldn't reread partition table: device is in use
Caused by: EBUSY: Device or resource busy

This comes up a few seconds after:
curl -O http://X.X.X.X:8000/example.ign
sudo coreos-installer install /dev/sda -i example.ign

bgilbert:
Could you drop the `coreos.inst=yes` and `coreos.inst.image_url` arguments, boot to a prompt (easier with the ISO image), and check `/proc/mounts` and `/proc/swaps` for partitions on `/dev/sda`?  The output of `sgdisk -p /dev/sda` would also be helpful.

2goobers:
Nothing to `/dev/sda` on `/proc/mounts`
Nothing in `/proc/swaps`

Here is `sudo sgdisk -p /dev/sda`

> Disk /dev/sda: 234441648 sectors, 111.8 GiB
> Model: KINGSTON SV300S3
> Sector size (logical/physical): 512/512 bytes
[...]

PR #276, opened 2020-06-23
Report busy partitions; disable MD and LVM auto-activation

bgilbert:
Ensure leftover MD and LVM partitions on the target disk are not automatically activated if install kargs are used, so they don't prevent the installer from gaining exclusive access to the disk.

Also extend the installer to report a list of busy partitions as a debugging aid.  Example output:

```
Installing Fedora CoreOS 32.20200601.3.0 x86_64 (512-byte sectors)
Partitions in use on /dev/sda:
    /dev/sda1 mounted on /var/mnt
    /dev/sda4 is swap device
    /dev/sda6 in use by /dev/dm-0
    /dev/sda8 in use by /dev/md127
Error: checking for exclusive access to /dev/sda
Caused by: couldn't reread partition table: device is in use
Caused by: EBUSY: Device or resource busy
```

Fixes https://github.com/coreos/coreos-installer/issues/261.


cgwalters:
This will only work for users that install via the kernel arguments...which is probably OK but we also want to support (if not emphasize) running coreos-installer via custom systemd units instead (possibly wrapped with other units before/after etc.) right?

Clearly an admin could inject those config fragments via Ignition...so maybe they'd end up in our documentation?

Can LVM be suppressed by systemd units instead?  That's probably simpler if so.

I'd assume MD can be suppressed by a kernel argument too?

bgilbert:
You're right that this doesn't work for custom coreos-installer units.  Would it make sense to supply a custom target (`coreos-installer-pre.target`) that the custom unit could Require+After?

There's no karg to suppress MD assembly in the real root.  For LVM, it might make sense to mask or override `lvm2-pvscan@.service`, but I'm not sure that's cleaner.

cgwalters:
Probably the cleanest but also the hardest fix would be to have `coreos-installer install --auto-teardown` or so which would look for any LVM/MD/etc devices which have have the target disk as a member.

Wait actually, wouldn't it work to just `wipefs -a` the target disk and then wait for a `udevadm settle` cycle?  (I assume this would work for kernel MD, LVM less sure)

Makefile
```diff
@@ -23,4 +23,4 @@ install-bin: all
 install-systemd: all
 	install -D -m 644 -t $(DESTDIR)/usr/lib/systemd/system systemd/*.{service,target}
 	install -D -t $(DESTDIR)/usr/lib/systemd/system-generators systemd/coreos-installer-generator
-	install -D -t $(DESTDIR)/usr/libexec systemd/coreos-installer-service
+	install -D -t $(DESTDIR)/usr/libexec systemd/coreos-installer-devices systemd/coreos-installer-service
```
lucab, line 26: I have a feeling both of these shouldn't be under `systemd/`, as it can only grow messier from here on.

src/blockdev.rs
```diff
@@ -123,14 +157,28 @@ fn get_partitions(device: &str) -> Result<Vec<BlkDev>> {
...
         if let Some(name) = fields.get("NAME") {
-            // Skip the device itself
-            if device == name {
+            // Only return partitions.  Skip the whole-disk device, as well
+            // as holders like LVM or RAID devices using one of the partitions.
+            if fields.get("TYPE") != Some(&"part".to_string()) {
                 continue;
             }
+            let (mountpoint, swap) = match fields.get("MOUNTPOINT") {
+                Some(mp) => {
```
lucab, line 166: Not sure if it reads better, but I think here you could employ a [match guard](https://doc.rust-lang.org/stable/rust-by-example/flow_control/match/guard.html):
`Some(mp) if mp == "XYZ" => { ... }`

src/blockdev.rs
```diff
@@ -97,6 +110,27 @@ impl BlkDev {
...
+        let holders = Path::new("/sys/block")
+            .join(
+                Path::new(&self.parent)
+                    .file_name()
+                    .chain_err(|| format!("{} has no filename", self.parent))?,
+            )
+            .join(
+                Path::new(&self.path)
+                    .file_name()
+                    .chain_err(|| format!("{} has no filename", self.path))?,
```
lucab, line 124: Nit: it would be nicer to have two slightly different error strings.

jlebon:
Hmm, I'm wary of trying to make coreos-installer that smart. At best, I agree with https://github.com/coreos/coreos-installer/pull/276#issuecomment-649526346 that this should probably be a switch. And then we should be more targeted about it, e.g.: if the user provides the switch, then de-activate all the layers which are directly in the mount hierarchy for the target device. This is harder to do though.

I think the new "Partitions in use" error message is a great improvement and might even be enough to guide users on what they need to do.

bgilbert:
> Probably the cleanest but also the hardest fix would be to have `coreos-installer install --auto-teardown` or so which would look for any LVM/MD/etc devices which have have the target disk as a member.

That feels odd to me.  If run from the command line, I think it's the user's job to give us a disk that's not currently in use, and if it is, they're in the best position to correct that.  And if run automatically by the operating system, it's the OS' job to not automatically start using parts of the target disk.  That view might be controversial though.

> Wait actually, wouldn't it work to just `wipefs -a` the target disk and then wait for a `udevadm settle` cycle? (I assume this would work for kernel MD, LVM less sure)

We can't wipe any data we're not actually going to overwrite; see #172.

> I think the new "Partitions in use" error message is a great improvement and might even be enough to guide users on what they need to do.

Yeah, I'm in favor of making the incremental improvement and seeing where that gets us.

bgilbert:
Updated.  I've added `coreos-installer-pre.target`, and we can document that custom installer units should `Require`/`After` it.

systemd/coreos-installer-pre.target
```diff
@@ -0,0 +1,4 @@
+[Unit]
+Description=Prepare for CoreOS Installer Target
+
+Requires=coreos-installer-devices.service
```
jlebon, line 4: Bikeshed: `coreos-installer-disable-auto-activation.service`? For `oneshot` services, I like names that have a verb in them to make it really clear what it does.

blueskyjunkie:
The problem this fixes seems to still exist?

https://github.com/coreos/coreos-installer/issues/261#issuecomment-2564358140

PR #277, opened 2020-06-24
systemd: move OnFailure into coreos-installer.service

bgilbert:
On Fedora 32, an `OnFailure` directive in a target unit [no longer does anything](https://github.com/systemd/systemd/issues/14142#issuecomment-634946368).

Reported by @nikita-dubrovskii.

### DIFF
--- a/systemd/coreos-installer.service
+++ b/systemd/coreos-installer.service
@@ -4,6 +4,8 @@ Before=coreos-installer.target
 After=network-online.target
 Wants=network-online.target
 ConditionKernelCommandLine=coreos.inst.install_dev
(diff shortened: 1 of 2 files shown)


dustymabe:
Should we add them to https://github.com/coreos/coreos-installer/blob/master/systemd/coreos-installer-noreboot.service as well?

bgilbert:
I thought about that, but decided that if `sulogin` failed for some reason, it'd be better to reboot anyway.  Thoughts?

dustymabe:
Not sure. I just noticed that it was covered before and now isn't. If you gave it some thought then I trust your judgement.

PR #282, opened 2020-06-29
systemd: use --ignition-url in service wrapper

bgilbert:
We no longer need to run curl by hand.

cc @jlebon

### DIFF
--- a/scripts/coreos-installer-service
+++ b/scripts/coreos-installer-service
@@ -53,14 +53,12 @@ if [ "${device##*/}" = "${device}" ]; then
 fi
 args+=("${device}")
 
-# Fetch the Ignition URL, if any, and cache it locally.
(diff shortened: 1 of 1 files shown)


lucab:
Wrt. retries, we do have some silly `retry` [module](https://github.com/coreos/afterburn/blob/master/src/retry/mod.rs) in Afterburn, which isn't particularly good and I'd be happy to drop in favor of a better crate. Honk if you have on in sight which is good enough to consolidate on.

bgilbert:
Filed https://github.com/coreos/coreos-installer/issues/283 for HTTP retries.

bgilbert:
Rebased.

bgilbert:
CI noticed that we need to pass `--insecure-ignition`.  Updated.

PR #284, opened 2020-06-29
blockdev: some refactoring

bgilbert:
blockdev still has a number of rough edges, but hopefully this is incrementally better.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -74,7 +74,7 @@ pub fn get_busy_partitions(device: &str) -> Result<Vec<BlkDev>> {
     Ok(ret)
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct BlkDev {
     pub path: String,
     pub label: Option<String>,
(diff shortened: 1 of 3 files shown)


src/blockdev.rs
```diff
@@ -85,8 +145,21 @@ pub struct BlkDev {
...
+            fstype: None,
+
+            parent: "".to_string(),
+            mountpoint: None,
+            swap: false,
+        }
+    }
+
+    /// Return start and end offsets within the disk.
+    pub fn get_offsets(&self) -> Result<(u64, u64)> {
```
lucab, line 162: This is not using anything from `&self` other than `.path`. On the other hand (if I didn't miss other places), I think this is the main consumer of the `new(path: &str)` constructor above (see comment there).

I'd make this an associate method which directly takes a path (i.e. not `self`), and directly attach it to either `Disk` or `Partition`.
Alternatively, put the start+end properties as fields in `Partition` and turn this into a plain getter.

bgilbert:
:100:, thanks!  Updated.

PR #290, opened 2020-07-02
 *: use our own std::io::copy() implementation

bgilbert:
The `std` one uses an 8 KiB buffer.  Use a larger buffer to amortize syscall overhead.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -1,0 +1,69 @@
+// Copyright 2019 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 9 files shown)


src/io.rs
```diff
@@ -0,0 +1,75 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use error_chain::bail;
+use std::io::{ErrorKind, Read, Write};
+
+use crate::errors::*;
+
+/// This is like `std::io:copy()`, but uses a buffer larger than 8 KiB.
```
cgwalters, line 20: Let's say *why* we do that too.  (It's in the commit message yes, but that stuff can get harder to find as code moves around)

I also looked at this for ostree a while ago...see https://github.com/ostreedev/ostree/commit/16c31a9b5#diff-7833e28ecd7891cb43d466346f39acbdR828
And linked from there: https://github.com/coreutils/coreutils/blob/master/src/ioblksize.h

Of course both that and the coreutils one are pre-Meltdown/Spectre.   And also pre-io-uring (which is going to dramatically change things like this).

Anyways seems fine as is, but it's worth thinking about e.g. whether we should argue for `std` to do this by default (although that'd be easier if there was a version that knew at least a lower bound for how much data we were going to be copying, which would help avoid allocating large buffers for small copies)

bgilbert:
Updated.  There's already an upstream bug: https://github.com/rust-lang/rust/issues/49921

PR #296, opened 2020-07-08
DNM: test Travis on s390x with newer toolchains

bgilbert:


bgilbert:
Works on 1.44, fails on 1.41.

PR #303, opened 2020-07-20
Split out a `libcoreinst` library and add a new `rdcore` binary

jlebon:
I thought I'd gauge interest in this approach before going too deep into it. My immediate goal here is to re-implement the logic for https://github.com/coreos/fedora-coreos-config/pull/503 into `rdcore`.

```
commit a1592f9a84dd8fa29d5c8be62e739a6af3f5b144
Date:   Mon Jul 20 17:35:20 2020 -0400

    Move logic to `libcoreinst` library

    Create a new `libcoreinst` library which contains all the logic, and
    make `coreos-installer` simply consist of `main.rs` which calls into it.

    This is prep for adding another binary which will reuse code in
    `libcoreinst`. Doing it this way is a bit awkward, because e.g.
    `cmdline` is also in the library, but it allows minimal churn while
    still making it easy to add other binaries.

commit 14016f3bcd49dd84891211add0ed4b3d7e54da55
Date:   Mon Jul 20 17:35:21 2020 -0400

    rdcore: new binary for initrd CoreOS tasks

    By design, FCOS and RHCOS have a lot going on in the initramfs stage and
    the logic there is likely going to keep growing. As a result, we have
    many non-trivial and crucial shell scripts living in the source config
    (https://github.com/coreos/fedora-coreos-config) which (1) have poor
    error-handling, (2) aren't easy to manually test or unit test, and (3)
    have low visibility because they're deeply nested within some dracut
    module.

    I'd like to propose a new `rdcore` binary which will try to address
    these issues simply by virtue of being in a proper programming language.
    This is shamelessly intended to be a "kitchen sink" collection of
    whatever logic is needed in the initramfs.

    Down the line we can split this out into its own package. But for now,
    it's much easier to tag along an existing one for experimenting without
    having to incur too much paperwork. It's an awkward arrangement, but I
    think the pros outweigh the cons overall, and at least
    `coreos-installer` already isn't very distro-agnostic.

    As for why make this a separate binary instead of just extending the
    CLI, a few reasons:
    1. it'll be easier to eventually rip out and move to its own package
    2. thanks to LTO, we only pay for the bits of the library we use
    3. we don't inadvertently add back installer support into the initrd

### DIFF
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -1,0 +1,42 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::{crate_version, App, AppSettings};
+use error_chain::bail;
+
+use libcoreinst::errors::*;
+
+pub enum Config {
+}
+
+/// Parse command-line arguments.
+pub fn parse_args() -> Result<Config> {
+    // Args are listed in --help in the order declared here.  Please keep
+    // the entire help text to 80 columns.
+    let app_matches = App::new("rdcore")
+        .version(crate_version!())
+        .setting(AppSettings::SubcommandRequiredElseHelp)
+        .global_setting(AppSettings::ArgsNegateSubcommands)
+        .global_setting(AppSettings::DeriveDisplayOrder)
+        .global_setting(AppSettings::DisableHelpSubcommand)
+        .global_setting(AppSettings::UnifiedHelpMessage)
+        .global_setting(AppSettings::VersionlessSubcommands)
+        .get_matches();
+
+    // temporarily; until we actually add our first command
+    #[allow(clippy::match_single_binding)]
+    match app_matches.subcommand() {
+        _ => bail!("unrecognized subcommand"),
+    }
+}
--- a/src/bin/rdcore/main.rs
+++ b/src/bin/rdcore/main.rs
@@ -1,0 +1,29 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by cgwalters** [comment]
We previously talked about having rpm-ostree be this "convenient Rust thing we can stick in the initramfs" (because we might do os updates, layering there) but I'm increasingly thinking that because of problems like "need to configure `/etc/yum.repos.d`" we basically need to do that in the real root.

So coreos-installer indeed may make more sense.

**Comment by lucab** [comment]
Where do we plan to draw the boundaries between Afterburn / Ignition / rdcore in initramfs?

I'm thinking about stuff like the [coreos-platform-chrony generator](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony) (whose logic could run as a service unit in initramfs).

**Comment by cgwalters** [comment]
Overall :+1: from me.  It'd probably be useful though to enumerate things we are likely to use this for in the short term.

For example, would we rewrite [the chrony sync](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/20platform-chrony/usr/lib/systemd/system-generators/coreos-platform-chrony)?  

[Edit, heh, Luca commented the same thing]

**Comment by jlebon** [comment]
Good questions.

Short-term, this will be used to house code needed for rootfs reprovisioning (see #304) and likely for the stage 2 rootfs work (see #305).

The rough guideline I'd say is that we should only write **new** complex CoreOS support code here and not convert well-tested working scripts otherwise. We might move them down the line, though without trying to fix/enhance their functionality it's not worth the risk IMO.

> Where do we plan to draw the boundaries between Afterburn / Ignition / rdcore in initramfs?

I think the boundary with Ignition is a bit easier to define because it tries to be distro-agnostic and is focused on just processing Ignition configs. The boundary with Afterburn is harder. Maybe let's define some guidelines together and stick it somewhere? Here's what I wrote up:
- Afterburn:
    - enablement for cloud-provided functionality
    - contains platform-specific code
    - contains dracut glue needed to wire it into CoreOS
- rdcore:
    - platform-agnostic
    - "implementation details" code to support CoreOS features
    - doesn't contain dracut glue in favour of keeping it all in f-c-c

**Comment by jlebon** [comment]
(Hmm, by those guidelines I think the chrony code should probably eventually move to Afterburn?)

File: src/bin/rdcore/cmdline.rs [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+        .version(crate_version!())
+        .setting(AppSettings::SubcommandRequiredElseHelp)
+        .global_setting(AppSettings::ArgsNegateSubcommands)
+        .global_setting(AppSettings::DeriveDisplayOrder)
+        .global_setting(AppSettings::DisableHelpSubcommand)
+        .global_setting(AppSettings::UnifiedHelpMessage)
+        .global_setting(AppSettings::VersionlessSubcommands)
+        .get_matches();
+
+    match app_matches.subcommand() {
```
bgilbert, line 37: clippy doesn't like the empty match block.

src/bin/rdcore/main.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use error_chain::quick_main;
+use errors::{Result, ResultExt};
+
+quick_main!(run);
+
+fn run() -> Result<()> {
+    let config = cmdline::parse_args().chain_err(|| "parsing arguments")?;
+
+    match config {
```
bgilbert, line 27: ...but is fine with this one.  :man_shrugging:

PR #305, opened 2020-07-21
rdcore: add stream-hash subcommand

bgilbert:
This command copies a stream from stdin to stdout, verifying the data via SHA-256 as it goes.  It receives an input file in the following format:

```
stream-hash sha256 <block-size>
<hexdigest>
<hexdigest>
<hexdigest>
...
```

The idea is that we can read `block-size` bytes into RAM, check their digest, send them to the next program in a shell pipeline, and move on to the next hash.  This lets us safely unpack a streaming download while verifying it, and is intended for downloading the PXE rootfs image.

### DIFF
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -12,12 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use clap::{crate_version, App, AppSettings};
+use clap::{crate_version, App, AppSettings, Arg, ArgMatches, SubCommand};
 use error_chain::bail;
 
 use libcoreinst::errors::*;
 
 pub enum Config {
+    StreamHash(StreamHashConfig),
+}
+
+pub struct StreamHashConfig {
+    pub hash_file: String,
 }
 
 /// Parse command-line arguments.
@@ -32,11 +37,30 @@ pub fn parse_args() -> Result<Config> {
         .global_setting(AppSettings::DisableHelpSubcommand)
         .global_setting(AppSettings::UnifiedHelpMessage)
         .global_setting(AppSettings::VersionlessSubcommands)
+        .subcommand(
+            SubCommand::with_name("stream-hash")
+                .about("Copy data from stdin to stdout, checking piecewise hashes")
+                .arg(
+                    Arg::with_name("hash-file")
+                        .value_name("hash-file")
+                        .help("Path to the piecewise hash file")
+                        .required(true)
+                        .takes_value(true),
+                ),
+        )
         .get_matches();
 
-    // temporarily; until we actually add our first command
-    #[allow(clippy::match_single_binding)]
     match app_matches.subcommand() {
+        ("stream-hash", Some(matches)) => parse_stream_hash(&matches),
         _ => bail!("unrecognized subcommand"),
     }
+}
+
+fn parse_stream_hash(matches: &ArgMatches) -> Result<Config> {
+    Ok(Config::StreamHash(StreamHashConfig {
+        hash_file: matches
+            .value_of("hash-file")
+            .map(String::from)
+            .expect("hash file missing"),
+    }))
 }
--- a/src/bin/rdcore/stream_hash.rs
+++ b/src/bin/rdcore/stream_hash.rs
@@ -1,0 +1,274 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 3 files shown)


src/bin/rdcore/cmdline.rs
```diff
@@ -32,11 +37,30 @@ pub fn parse_args() -> Result<Config> {
         .global_setting(AppSettings::DisableHelpSubcommand)
         .global_setting(AppSettings::UnifiedHelpMessage)
         .global_setting(AppSettings::VersionlessSubcommands)
+        .subcommand(
+            SubCommand::with_name("stream-hash")
+                .about("Copy data from stdin to stdout, checking piecewise hashes")
+                .arg(
+                    Arg::with_name("hash-file")
+                        .value_name("hash-file")
```
jlebon, line 45: We've been a bit inconsistent about this, though WDYT about using capital letters as a convention for value names? So e.g. `HASH_FILE`.

Obviously not a blocker.

src/bin/rdcore/stream_hash.rs
```diff
@@ -0,0 +1,240 @@
...
+use error_chain::bail;
+use openssl::sha;
+use regex::Regex;
+use std::fs::OpenOptions;
+use std::io::{self, stdin, stdout, BufRead, BufReader, Read, Write};
+
+use crate::cmdline::*;
+use crate::errors::*;
+
+pub fn stream_hash(config: &StreamHashConfig) -> Result<()> {
```
jlebon, line 24: Would be helpful to document the file format somewhere around here.

src/bin/rdcore/stream_hash.rs
```diff
@@ -0,0 +1,240 @@
...
+    };
+    let bufsize = captures
+        .get(2)
+        .expect("bufsize not found")
+        .as_str()
+        .parse::<usize>()
+        .chain_err(|| "couldn't parse bufsize")?;
+    if bufsize == 0 {
+        bail!("buffer size cannot be zero");
+    }
```
jlebon, line 71: Hmm, should be sanity-check the bufsize here that it has some reasonable upper bound? E.g. 16M or something.

src/bin/rdcore/stream_hash.rs
```diff
@@ -0,0 +1,240 @@
...
+    if hash_file
+        .read_line(&mut line)
+        .chain_err(|| "reading hash file")?
+        == 0
+    {
+        bail!("hash file is empty");
+    }
+
+    // parse it
+    let captures = Regex::new(r"^stream-hash ([a-z0-9]+) ([0-9]+)\n$")
```
jlebon, line 51: Heh, same feeling about regex here as [there](https://github.com/coreos/coreos-installer/pull/153#discussion_r451780476). Feels like a `splitn` would be cleaner and possibly less lines, but good with this as well!

src/bin/rdcore/stream_hash.rs
```diff
@@ -0,0 +1,240 @@
...
+        loop {
+            count += match input.read(&mut buf[count..]) {
+                Ok(0) => break,
+                Ok(n) => n,
+                Err(ref e) if e.kind() == io::ErrorKind::Interrupted => continue,
+                Err(e) => return Err(e).chain_err(|| "reading input"),
+            };
+        }
+        if count == 0 {
+            bail!("premature EOF at {} reading input", offset);
```
jlebon, line 93:
```suggestion
            bail!("premature data input EOF at offset {}", offset);
```

?

src/bin/rdcore/stream_hash.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        // hash and compare
+        let data = &buf[..count];
+        let found_hash = hash_func(data);
+        if expected_hash != found_hash {
+            bail!("hash mismatch at {}", offset);
```
jlebon, line 100:
```suggestion
            bail!("hash mismatch at {}; expected {}, got {}", offset, expected_hash, found_hash);
```

src/bin/rdcore/stream_hash.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        // write out buffer
+        output.write_all(data).chain_err(|| "writing output")?;
+        offset += data.len() as u64;
+    }
+
+    // ran out of hashes; make sure we ran out of data
+    if input.read(&mut buf[..1]).chain_err(|| "draining input")? != 0 {
+        bail!("found extra data at {} reading input", offset);
```
jlebon, line 110:
```suggestion
        bail!("found extra data input at offset {}", offset);
```

src/bin/rdcore/cmdline.rs
bgilbert: Grepping for `value_name`, I'd say I've been consistent and you've been consistent.  :grin:  Personally I find the upper-case less readable.

PR #307, opened 2020-07-22
osmet: add pack --fast option for worse compression

bgilbert:
On my machine, with coreos-installer built in release mode, this reduces a cosa `osmet-pack` invocation from 71s to 42s (and there are two of them in a `buildextend-live`), while increasing osmet file size from 5 MiB to 7.4 MiB.

We could presumably do better by swapping out the compression algorithm in this case, but that would require a file format change and some additional code complexity.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -93,6 +93,7 @@ pub struct OsmetPackConfig {
     pub checksum: String,
     pub description: String,
     pub rootdev: Option<OsmetRootBlkDevReal>,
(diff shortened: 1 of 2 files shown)


bgilbert:
gzip level 1 takes 38s on my machine and produces a 16 MiB file, which is not a meaningful improvement.  Using a speed-optimized algorithm (LZ4, etc.) would introduce another dependency.  So it seems reasonable to stick with xz.

PR #313, opened 2020-07-25
download: refactor progress reporting

bgilbert:
Implement our own Reader and drop the `progress_streams` dependency.

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -16,12 +16,12 @@ use byte_unit::Byte;
 use error_chain::bail;
 use flate2::read::GzDecoder;
 use nix::unistd::isatty;
-use progress_streams::ProgressReader;
 use std::fs::{remove_file, File, OpenOptions};
-use std::io::{copy, stderr, BufRead, BufReader, BufWriter, Read, Seek, SeekFrom, Write};
-use std::num::NonZeroU32;
+use std::io::{self, copy, stderr, BufRead, BufReader, BufWriter, Read, Seek, SeekFrom, Write};
+use std::num::{NonZeroU32, NonZeroU64};
 use std::os::unix::io::AsRawFd;
 use std::path::{Path, PathBuf};
+use std::result;
 use std::time::{Duration, Instant};
 use xz2::read::XzDecoder;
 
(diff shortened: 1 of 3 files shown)


src/download.rs
```diff
@@ -349,9 +303,90 @@ pub fn download_to_tempfile(url: &str) -> Result<File> {
...
+    last_report: Instant,
+
+    tty: bool,
+    prologue: &'static str,
+    epilogue: &'static str,
+}
+
+impl<'a, R: Read> ProgressReader<'a, R> {
+    fn new(source: R, length: Option<u64>, artifact_type: &'a str) -> Result<Self> {
+        let tty = isatty(stderr().as_raw_fd()).chain_err(|| "checking if stderr is a TTY")?;
```
lucab, line 321: Nit: I think it is more useful to log a warning and assume `false` if this `isatty` check fails.
That is, assuming this logic is mostly cosmetic, I think we are more interested into having the download logic still working (just with less bells and whistles) on those corner cases.

src/download.rs
```diff
@@ -349,9 +303,90 @@ pub fn download_to_tempfile(url: &str) -> Result<File> {
...
 }
 
-/// Format a size in bytes.
-fn format_bytes(count: u64) -> String {
-    Byte::from_bytes(count.into())
-        .get_appropriate_unit(true)
-        .format(1)
+struct ProgressReader<'a, R: Read> {
+    source: R,
+    length: Option<u64>,
```
lucab, line 308: Coming back from the rest of the code, you may consider making `length` an `Option<NonZeroU64>` (and storing the formatted label too).

src/download.rs
```diff
@@ -349,9 +303,90 @@ pub fn download_to_tempfile(url: &str) -> Result<File> {
...
+            || self.length == Some(self.position)
+        {
+            self.last_report = Instant::now();
+            match self.length {
+                Some(length) if length > 0 => eprint!(
+                    "{}Read {} {}/{} ({}%){}",
+                    self.prologue,
+                    self.artifact_type,
+                    Self::format_bytes(self.position),
+                    Self::format_bytes(length),
```
lucab, line 367: I think you can compute this cosmetic string when building the reader and directly cache it there, to avoid recomputing on each iteration.

lucab:
Code looks fine, I just left a bunch of minor suggestions.

However, I think I'm missing why we are doing this refactor. Is it only for code maintenance, or are there some performance implications behind this? Can you maybe expand the commit message a bit with the rationale?

bgilbert:
It's just a cleanup.  The `progress_streams` dependency is actually quite trivial, and I realized the code would be cleaner if progress handling were not implemented via a closure.

bgilbert:
Updated.

src/download.rs
```diff
@@ -349,9 +303,97 @@ pub fn download_to_tempfile(url: &str) -> Result<File> {
...
+            || self.length.map(|l| l.get()) == Some(self.position)
+        {
+            self.last_report = Instant::now();
+            match self.length {
+                Some(length) => eprint!(
+                    "{}Read {} {}/{} ({}%){}",
+                    self.prologue,
+                    self.artifact_type,
+                    Self::format_bytes(self.position),
+                    self.length_str.as_ref().unwrap(),
```
lucab, line 374: Minor: you may consider stuff this too into a `length: Option<(NonZeroU64, String)>` so that you can get it here via matching and without unwrapping.

bgilbert:
Updated.

PR #314, opened 2020-07-26
Move rdcore behind a feature flag

bgilbert:
We shouldn't build rdcore by default.  osmet is a tougher call, especially since it's impractical to omit the code without putting conditionals everywhere, but it seems useful to omit that functionality on non-CoreOS systems which won't use it anyway.


src/io.rs
```diff
@@ -209,4 +209,82 @@ mod tests {
             assert!(hasher.validate(&mut rd).is_ok() == *valid);
         }
     }
+
+    #[test]
+    fn test_copy_n() {
```
lucab, line 214: If this PR  does not progress quick enough, this commit can be split to its own PR and merged separately.

lucab:
In my experience feature flags makes a project harder to test/format/lint/maintain.
I think it may make sense to have some quick flag in the Makefile to only perform `cargo build --bin coreos-installer`, but I wouldn't start introducing a matrix of Cargo features for this.

bgilbert:
I'm okay dropping the osmet part, but we [document](https://github.com/coreos/coreos-installer#install-with-cargo) `cargo install coreos-installer` and I don't think that should install rdcore.

bgilbert:
Okay, dropped the osmet flag.

.travis.yml
```diff
@@ -28,6 +28,10 @@ before_script:
 
 script:
   - cargo test
+  - cargo test --features rdcore
   - bash -c 'if [[ "$TRAVIS_RUST_VERSION" == "$CLIPPY_RUST_VERSION" ]]; then
       cargo clippy -- -D warnings;
     fi'
+  - bash -c 'if [[ "$TRAVIS_RUST_VERSION" == "$CLIPPY_RUST_VERSION" ]]; then
+      cargo clippy --features rdcore -- -D warnings;
```
jlebon, line 36: Minor: can't we fold this into the previous script line?

bgilbert: Sure, if you'd prefer.  Done.

Issue #184, opened 2020-03-11
downloads/installs slowly


cgwalters:
In `kola testiso` we spawn a local webserver and point qemu at it.  The legacy installer downloads and writes to disk in about one second.  This new one takes...about 4 minutes.  Something must be going terribly wrong :smile: 

I didn't bother to try to debug this much yet, but it seems like the host webserver is spending a lot of time idle.

jlebon:
I'm going to guess two things: not using `O_DIRECT` anymore (via `dd`'s `oflag=direct`). and possibly using a small block size (with `dd` we had `bs=1M`). `std::io::copy` does opportunistically use `copy_file_range`, but I'm not sure if this works for block devices. It may be falling back to a read & write loop in 8K blocks, which... will take a while. Probably worth using our own 1M read & write loop and possibly using `O_DIRECT`.

bgilbert:
That's `fs::copy`.  `io::copy` seems to be doing a [straightforward copy loop](https://github.com/rust-lang/rust/blob/c20d7eecbc0928b57da8fe30b2ef8528e2bdd5be/src/libstd/io/util.rs#L43).

jlebon:
Ahh yup, you're right. So we're definitely [falling back to 8k blocks](https://github.com/rust-lang/rust/blob/c20d7eecbc0928b57da8fe30b2ef8528e2bdd5be/src/libstd/sys_common/io.rs#L1) then.

bgilbert:
We could pretty easily reimplement `copy()` on top of `copy_n()` from #187.

cgwalters:
I did this but it fails to compile because...well, Rust correctly points out that it wouldn't be threadsafe :tada: 

```
diff --git a/Cargo.lock b/Cargo.lock
index 5f75a3a..594370f 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,6 +119,7 @@ dependencies = [
  "byte-unit 3.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "clap 2.33.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "cpio 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "crossbeam-utils 0.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.12.2 (registry+https://github.com/rust-lang/crates.io-index)",
[...]

PR #315, opened 2020-07-26
Drop copy() reimplementation; use buffered I/O with larger buffers

bgilbert:
The current image pipeline is:

- local file or HTTP response →
- [some Readers that don't affect I/O size] →
- `BufReader` with 8 KiB buffer →
- decompressor →
- `copy()` with 256 KiB buffer →
- disk.

`copy()`'s large buffer doesn't matter here.  The decompressor will read data in whatever chunk size it wants, which will be aggregated only up to 8 KiB.  The decompressor will also produce data in whatever chunk size it wants, depending on the compressed stream, and those chunks won't be aggregated at all.

Upgrade the `BufReader` to a 256 KiB buffer and add 256 KiB `BufWriter`s.  This doesn't seem to affect performance much in my local tests, but it's still the right thing to do.  Switch back to `std::io::copy`, since our reimplementation was misguided; add buffers in notable places instead.

### DIFF
--- a/src/download.rs
+++ b/src/download.rs
@@ -18,7 +18,7 @@ use flate2::read::GzDecoder;
 use nix::unistd::isatty;
 use progress_streams::ProgressReader;
 use std::fs::{remove_file, File, OpenOptions};
-use std::io::{stderr, BufRead, BufReader, BufWriter, Read, Seek, SeekFrom, Write};
+use std::io::{copy, stderr, BufRead, BufReader, BufWriter, Read, Seek, SeekFrom, Write};
 use std::num::NonZeroU32;
 use std::os::unix::io::AsRawFd;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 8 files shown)


cgwalters:
Sounds OK to me (didn't review code).

The way I'd measure things is comparing versus `curl | gunzip` basically.  

I do think we should investigate `O_DIRECT` and, maybe [using multiple threads](https://github.com/coreos/coreos-installer/issues/184#issuecomment-619240165) works better on top of this?

src/download.rs
```diff
@@ -299,9 +296,13 @@ where
 pub fn image_copy_default(
     first_mb: &[u8],
     source: &mut dyn Read,
-    dest: &mut File,
+    dest_file: &mut File,
     _dest_path: &Path,
 ) -> Result<()> {
+    // Amortize write overhead.  The decompressor will produce bytes in
+    // whatever chunk size it chooses.
+    let mut dest = BufWriter::with_capacity(BUFFER_SIZE, dest_file);
```
lucab, line 304: It would be better to flush this `BufWriter` before it gets dropped.

src/s390x/dasd.rs
```diff
@@ -66,6 +66,9 @@ pub fn image_copy_s390x(
     // there shouldn't be any partition data in the first MiB, so don't
     // worry about copying first_mb
     let mut cursor: u64 = 1024 * 1024;
+    // amortize write overhead; the decompressor will produce bytes in
+    // whatever chunk size it chooses
+    let mut dest = BufWriter::with_capacity(BUFFER_SIZE, dest_file);
```
lucab, line 71: It would be better to flush this `BufWriter` before it gets dropped.

bgilbert:
Good catches.  Updated.

PR #320, opened 2020-07-27
io: use std::fs::canonicalize instead of resolve_symlink

nikita-dubrovskii:
resolve_symlink doesn't return an absolute path, so we can get as result smth like `../../devname`. For s390x
with DASD it breaks the installation, becase we use `/dev/dasd` as pattern in `is_dasd()`

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -18,7 +18,10 @@ use nix::{errno::Errno, mount};
 use regex::Regex;
 use std::collections::HashMap;
 use std::convert::TryInto;
-use std::fs::{metadata, read_dir, read_to_string, remove_dir, File, OpenOptions};
+use std::fs::{
+    canonicalize, metadata, read_dir, read_to_string, remove_dir, symlink_metadata, File,
+    OpenOptions,
+};
 use std::num::{NonZeroU32, NonZeroU64};
 use std::os::linux::fs::MetadataExt;
 use std::os::raw::c_int;
@@ -30,7 +33,6 @@ use std::thread::sleep;
 use std::time::Duration;
 
 use crate::errors::*;
-use crate::io::resolve_link;
 
 #[derive(Debug)]
 pub struct Disk {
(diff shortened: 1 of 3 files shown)


cgwalters:
Isn't it simpler to just change all of the callers of this function to use `.canonicalize()` directly and remove it?  Something like
```
diff --git a/src/blockdev.rs b/src/blockdev.rs
index 30d6f91..4ba6560 100644
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -388,7 +388,9 @@ impl Partition {
         // Now assume a kpartx "partition", where the path is a symlink to
         // an unpartitioned DM device node.
         // /sys/block/dm-1
-        let (target, is_link) = resolve_link(&self.path)?;
+        let meta = self.path.symlink_metadata()?;
+        let is_link = meta.file_type().is_symlink();
+        let target = self.path.canonicalize()?;
         if is_link {
             let devdir = basedir.join(
                 target
```

nikita-dubrovskii:
> Isn't it simpler to just change all of the callers of this function to use `.canonicalize()` directly and remove it? 

I don't think so, don't want to copy-paste it

bgilbert:
Thanks for the fix.  I didn't know about `canonicalize()` when I wrote that function.  I agree with @cgwalters; there are two callers and only one of them needs to know whether the path is a symlink.

nikita-dubrovskii:
> I agree with @cgwalters; there are two callers and only one of them needs to know whether the path is a symlink.

done

src/install.rs
```diff
@@ -428,7 +428,10 @@ fn clear_partition_table(dest: &mut File, table: &mut dyn PartTable) -> Result<(
 }
 
 fn is_dasd(config: &InstallConfig) -> Result<bool> {
-    let (target, _) = resolve_link(&config.device)?;
+    let target = PathBuf::from(&config.device);
+    let target = target
```
bgilbert, line 432: Unnecessary reassignment.

src/blockdev.rs
```diff
@@ -388,8 +387,16 @@ impl Partition {
...
         // /sys/block/dm-1
-        let (target, is_link) = resolve_link(&self.path)?;
+        let target = PathBuf::from(&self.path);
+        let is_link = target
+            .symlink_metadata()
+            .chain_err(|| format!("reading metadata for {}", target.display()))?
+            .file_type()
+            .is_symlink();
         if is_link {
+            let target = target
```
bgilbert, line 397: The initial value of `target` is misleading, I think.  What about this?

```suggestion
        if symlink_metadata(&self.path)
            .chain_err(|| format!("reading metadata for {}", &self.path))?
            .file_type()
            .is_symlink() {
            let target = Path::new(&self.path)
```

Issue #190, opened 2020-03-19
Consider supporting a safety check to avoid overwriting data partitions


bgilbert:
# Feature Request #

## Desired Feature ##

Users might want to use coreos-installer to reprovision existing systems that have a data partition on the boot disk.  coreos-installer overwrites the partition table, but that's okay, since the Ignition config can recreate the data partition in the same place and its data will be intact.

However, there's no explicit guarantee that future install images won't be larger than current ones, which could cause coreos-installer to silently clobber the data partition.

Consider adding some mechanism to ensure important data isn't clobbered.  I don't think we can do this automatically; we'd have to understand the intentions of the existing partition table.  The simplest (and least user-friendly) approach is to add an argument specifying a byte offset limit, and error out if we try to cross it.

We can't generally perform this safety check without actually trying to install; the supported decompressors don't provide a way to obtain the uncompressed length of the payload.

PR #321, opened 2020-07-27
Add partition preservation

glennswest:
Add `--save-partlabel <glob>` and `--save-partindex <range>` options, which save partition table metadata for existing data partitions and then restore those partitions after writing the image.  `--save-partlabel` takes a glob pattern matching the partition label, and `--save-partindex` takes a partition number or range of partition numbers (possibly single-ended).  For example, to save all partitions with indexes greater than 4, specify `--save-partindex 5-`.  The specified partitions need not exist.  If no `--save-*` options are specified, no partitions are saved.

Multiples of each option can be specified, and/or multiple filters can be specified within a single option argument by separating them with commas.  There are corresponding kargs `coreos.inst.save_partlabel` and `coreos.inst.save_partindex`, which cannot be repeated but which accept comma-separated patterns.

Upon restore, try to reuse the original partition number.  If it's not available, renumber the partition to one more than the highest number used so far.  For simplicity, never backfill entries earlier in the partition table, even if the corresponding slot is unused.

If a saved partition overlaps with the image contents, fail.  This can't be detected in advance, so detect the overrun during fetch and stop before the saved partition is clobbered.  Also fail if the install image has a partition extending past the end of the image that overlaps with a saved partition.

On any failure, after clearing the partition table, restore any saved partitions.  In addition, augment partition-table clearing to clear the backup GPT, since some tools may otherwise hallucinate partitions that were overwritten during the install.

All of this assumes GPT partitioning both in the image and on disk, so if `--save-*` options are specified for a DASD target, fail.

Fixes https://github.com/coreos/coreos-installer/issues/190.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,13 @@ line.
   CoreOS image is being installed on.  Optional; defaults to `metal`.
   Normally this should be specified only if installing inside a virtual
   machine.
+* `coreos.inst.save_partlabel` - Comma-separated labels of partitions to
+  preserve during the install.  Glob-style wildcards are permitted.  The
+  specified partitions need not exist.  Optional.
+* `coreos.inst.save_partindex` - Comma-separated indexes of partitions to
+  preserve during the install.  Ranges (`m-n`) are permitted, and either `m`
+  or `n` can be omitted.  The specified partitions need not exist.
+  Optional.
 * `coreos.inst.insecure` - Permit the OS image to be unsigned.  Optional.
 * `coreos.inst.skip_reboot` - Don't reboot after installing.  Optional.
 
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use error_chain::bail;
+use gptman::{GPTPartitionEntry, GPT};
 use nix::sys::stat::{major, minor};
 use nix::{errno::Errno, mount};
 use regex::Regex;
@@ -31,7 +32,9 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::thread::sleep;
 use std::time::Duration;
+use uuid::Uuid;
 
+use crate::cmdline::PartitionFilter;
 use crate::errors::*;
 use crate::util::*;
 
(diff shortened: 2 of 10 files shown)


src/blockdev.rs
```diff
@@ -188,8 +201,124 @@ impl Disk {
...
+  pub fn add_extra_gptpartitions(disk:&str,extra_parts:Vec<GptPart>,dsneeded:bool) -> Result<()>  {
+    let mut f = std::fs::File::open(disk.to_string())
+      .expect("Cannot open disk");
+    let mut gpt = gptman::GPT::find_from(&mut f)
+      .expect("GPT Partitions not found");
+    let max_size : u64 =  gpt.get_maximum_partition_size().unwrap_or(0);
+    let mut format_needed: bool = false;
+    if extra_parts.len() == 0 && dsneeded { 
+       let gb: u64 = 1073741824; // 1 GB
+       //let gb: u64 = 1000000000; // 1 GB
```
ashcrow, line 244: Left over from debugging?

ashcrow:
Looks like Travis CI failures on Rust < 1.45

- https://travis-ci.org/github/coreos/coreos-installer/jobs/712341958#L581-L617
- (looks more like a toolchain issue) https://travis-ci.org/github/coreos/coreos-installer/jobs/712341964

src/blockdev.rs
```diff
@@ -188,8 +201,124 @@ impl Disk {
...
+                   name:       p.partition_name.to_string(),
+                   } );
+              }
+           uidx = uidx + 1;
+           }
+        }
+   return result;
+   }
+
+  pub fn add_extra_gptpartitions(disk:&str,extra_parts:Vec<GptPart>,dsneeded:bool) -> Result<()>  {
```
ashcrow, line 235: nit: argument spacing :-)

src/cmdline.rs
```diff
@@ -262,9 +264,19 @@ pub fn parse_args() -> Result<Config> {
...
+                )
+                .arg(
+                    Arg::with_name("dsneeded")
+                        .long("dsneeded")
+                        .help("During install, create datastore if not present"),
+                )
+                .arg(
+                    Arg::with_name("wipedisk")
+                        .long("wipedisk")
+                        .help("Always delete all partitions before install"),
```
ashcrow, line 279:
```suggestion
                        .help("Before install, permanently delete all partitions"),
```

src/cmdline.rs
```diff
@@ -262,9 +264,19 @@ pub fn parse_args() -> Result<Config> {
                         .takes_value(true),
                 )
                 .arg(
-                    Arg::with_name("preserve-on-error")
-                        .long("preserve-on-error")
-                        .help("Don't clear partition table on error"),
+                    Arg::with_name("clear-on-error")
```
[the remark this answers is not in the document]
glennswest: The default is now "preserve-on-error", thus the reversal of the flag.

src/install.rs
```diff
@@ -67,32 +67,51 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
         }
 
         // return a generic error so our exit status is right
         bail!("install failed");
     }
 
+    // Make sure end_lba and partition table is consistent 
+    Disk::update_gpt_headers(&config.device);
+
+    Disk::add_extra_gptpartitions(&config.device,extrapart,config.dsneeded)
```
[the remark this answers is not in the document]
glennswest: This update headers is because there is a mismatch in the headers that can cause gptman issues, and most partitioning software. This was added to force it to be clean before anything else. Gptman may fail to do changes otherwise.

src/cmdline.rs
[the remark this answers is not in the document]
glennswest: If the default is to "wipe", then the liability gets to be a issue. Also the "partitions" to preserve may be "different" on different hosts in cluster. So this quickly explodes to complexity.

src/install.rs
```diff
@@ -67,32 +67,51 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+    println!("Extra Partitions = {}\n",extrapart.len());
+
     // get reference to partition table
     // For kpartx partitioning, this will conditionally call kpartx -d
     // when dropped
     let mut table = Disk::new(&config.device)
         .get_partition_table()
         .chain_err(|| format!("getting partition table for {}", &config.device))?;
 
+    if config.wipedisk {
```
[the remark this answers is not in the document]
glennswest: Yes, while I agree in part, my concern, based on experience is you have something that is "incorrect" in the existing disk, and you need to re-establish a sensible config. As implemented I believe it to be safer when cleanup is needed.

src/cmdline.rs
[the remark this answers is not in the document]
glennswest: default now is, if error, high partitions are saved, all others wiped. If clear-on-error, will wipe high partitions.

bgilbert:
Ready for review!

src/blockdev.rs
```diff
@@ -461,6 +464,205 @@ impl Mount {
     }
 }
 
+#[derive(Debug)]
+pub struct SavedPartitions {
+    sector_size: Option<u64>,
```
lucab, line 469: Nit: I think an `Option<NonZeroU16>` could be enough here. But most of the logic below really expects a `None` to never happen here; it would be good to rework the `new` flow in order to be able to drop the `Option` here.

src/blockdev.rs
```diff
@@ -461,6 +464,205 @@ impl Mount {
...
+            .open(disk)
+            .chain_err(|| format!("opening {} for reading", disk.display()))?;
+        // true except in unit tests
+        if f.metadata()
+            .chain_err(|| format!("getting metadata for {}", disk.display()))?
+            .file_type()
+            .is_block_device()
+        {
+            let disk_sector_size = get_sector_size(&f)?.get() as u64;
+            if disk_sector_size != self.sector_size.expect("sector size unset") {
```
lucab, line 557: Nit: I'd be happier if we could avoid this `expect` (and the matching `unwrap` below).

src/blockdev.rs
```diff
@@ -461,6 +464,205 @@ impl Mount {
...
+                }
+                GPT::new_from(
+                    &mut f,
+                    self.sector_size.expect("sector size unset"),
+                    *Uuid::new_v4().as_bytes(),
+                )
+                .chain_err(|| format!("creating new partition table for {}", disk.display()))?
+            }
+        };
+        if gpt.sector_size != self.sector_size.expect("sector size unset") {
```
lucab, line 587: Same feedback as the `expect` above.

src/cmdline.rs
```diff
@@ -230,6 +238,32 @@ pub fn parse_args() -> Result<Config> {
                         .help("For use with -n.")
                         .next_line_help(true), // so we can stay under 80 chars
                 )
+                .arg(
+                    Arg::with_name("save-label")
+                        .long("save-label")
+                        .value_name("glob")
+                        .help("Save partition with this label glob")
```
lucab, line 245: Plural "partitions" according to the code?

src/install.rs
```diff
@@ -418,7 +465,12 @@ fn clear_partition_table(dest: &mut File, table: &mut dyn PartTable) -> Result<(
         .chain_err(|| "seeking to start of disk")?;
     let zeroes: [u8; 1024 * 1024] = [0; 1024 * 1024];
     dest.write_all(&zeroes)
-        .chain_err(|| "clearing partition table")?;
+        .chain_err(|| "clearing primary partition table")?;
+    // and now the backup GPT
+    dest.seek(SeekFrom::End(-(zeroes.len() as i64)))
```
lucab, line 470: I'm concerned about this one:
 * the size seems way larger than the GPT (and in general it depends on the sector-size IIRC)
 * the offset seems off, and we may end up destroying legit data there

bgilbert: Good catch!  I picked 1 MiB because that's what we use at the start of the disk, but we have a right to clobber that data and not the data at the end of the disk.  I'll reduce it to one sector.

src/blockdev.rs
```diff
@@ -461,6 +464,205 @@ impl Mount {
...
+
+        for (i, p) in gpt.iter() {
+            if Self::accept(i, p, filters) {
+                result.partitions.push((i, p.clone()));
+            }
+        }
+        Ok(result)
+    }
+
+    fn accept(i: u32, p: &GPTPartitionEntry, filters: &[PartitionFilter]) -> bool {
```
jlebon, line 525: Optional/bikeshed: `matches_filters` ?

src/blockdev.rs
```diff
@@ -461,6 +464,205 @@ impl Mount {
...
+            return Ok(());
+        }
+
+        // get or create GPT
+        let disk = disk.as_ref();
+        let mut f = OpenOptions::new()
+            .read(true)
+            .open(disk)
+            .chain_err(|| format!("opening {} for reading", disk.display()))?;
+        // true except in unit tests
```
jlebon, line 553: Optional: could factor out into a e.g. `verify_sector_size_matches_gpt()` or something like that to reduce code duplication.

src/blockdev.rs
```diff
@@ -461,6 +464,205 @@ impl Mount {
...
+                    )
+                }
+            }
+        }
+
+        // merge saved partitions into partition table
+        // find partition number one larger than the largest used one
+        let mut next = gpt
+            .iter()
+            .fold(1, |prev, (i, e)| if e.is_used() { i + 1 } else { prev });
```
jlebon, line 623: Given how we handle `.is_used()`, should we error out if the user tries to save an unused partition?

src/blockdev.rs
bgilbert: I used `u64` because that's the type of `gpt.sector_size`, of which this is a copy.

`sector_size` is `None` in the case where we return early from `new()` (because we have no filters) and haven't looked at the GPT at all.  I think that's a useful check but I agree that the `expect`/`unwrap` are ugly; I'll unwrap `sector_size` early in `write()` and store it as a local.

src/blockdev.rs
```diff
@@ -616,14 +818,24 @@ pub fn udev_settle() -> Result<()> {
...
 /// sector size, if any can be determined.
-pub fn detect_formatted_sector_size(buf: &[u8]) -> Option<NonZeroU32> {
+pub fn detect_formatted_sector_size_start(buf: &[u8]) -> Option<NonZeroU32> {
+    detect_formatted_sector_size(buf, 512, 4096)
+}
+
+/// Inspect a buffer 4 KiB from the end of a disk image and return its
+/// formatted sector size, if any can be determined.
+pub fn detect_formatted_sector_size_end(buf: &[u8]) -> Option<NonZeroU32> {
+    detect_formatted_sector_size(buf, 3584, 0)
```
jlebon, line 827: Let's write as `4096 - 512` ?

src/blockdev.rs
```diff
@@ -717,13 +930,279 @@ mod tests {
...
+                    make_part(1, "boot", 1, 384),
+                    make_part(2, "EFI-SYSTEM", 384, 512),
+                    make_part(4, "root", 1024, 2200),
+                ],
+                vec![],
+            ),
+        ];
+
+        let base = make_disk(512, &base_parts);
+        for (testnum, (filter, expected_image, expected_blank)) in tests.iter().enumerate() {
```
jlebon, line 1071: Love these tests!

Minor/optional: wouldn't it make more sense to flip the last two tuple elements, e.g. `(filters, expected_blank, expected_image)`? So it'd be the inputs first, then the expected output.

src/blockdev.rs
```diff
@@ -461,6 +464,205 @@ impl Mount {
...
+
+impl SavedPartitions {
+    pub fn new<P: AsRef<Path>>(disk: P, filters: &[PartitionFilter]) -> Result<Self> {
+        let disk = disk.as_ref();
+        let mut result = Self {
+            sector_size: None,
+            partitions: Vec::new(),
+        };
+
+        // skip reading partition table if we don't have an accept filter
```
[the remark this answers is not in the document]
bgilbert: We'd still need `is_saved()`, because the existence of filters does not imply the existence of partitions matching the filters.

src/cmdline.rs
```diff
@@ -830,3 +922,64 @@ fn parse_osmet_fiemap(matches: &ArgMatches) -> Result<Config> {
...
+        let g = |v| Label(glob::Pattern::new(v).unwrap());
+        let i = |v| Some(NonZeroU32::new(v).unwrap());
+
+        assert_eq!(
+            parse_partition_filters(&["foo", "z*b?", ""], &["1", "7-7", "2-4", "-3", "4-"])
+                .unwrap(),
+            vec![
+                g("foo"),
+                g("z*b?"),
+                g(""),
```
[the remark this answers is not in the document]
bgilbert: They match empty partition names, which seems useful.  I'll add a test for that.

src/blockdev.rs
bgilbert: No, it's intentionally valid to specify a filter that doesn't match anything.  That lets the user specify all relevant partitions that might be on any given target system.

src/download.rs
```diff
@@ -409,3 +417,114 @@ impl<'a, R: Read> Drop for ProgressReader<'a, R> {
...
+        let allowed = self.remaining.min(buf.len() as u64);
+        if allowed == 0 {
+            // reached the limit; only error if we're not at EOF
+            return match self.source.read(&mut buf[..1]) {
+                Ok(0) => Ok(0),
+                Ok(_) => Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    format!("{} at offset {}", self.limit_cause, self.length),
+                )),
+                Err(e) => Err(e),
```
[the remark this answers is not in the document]
bgilbert: Do we need any?  We don't wrap the error, and that error path doesn't change any of our internal state, so the caller should be able to retry normally.

src/blockdev.rs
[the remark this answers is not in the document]
bgilbert: `new()` could return an `Option`, I suppose.  I see your point, though I tend to find the unconditional calls easier to read.  Will defer for now.

bgilbert:
Updated!

README.md
```diff
@@ -107,6 +107,13 @@ line.
   CoreOS image is being installed on.  Optional; defaults to `metal`.
   Normally this should be specified only if installing inside a virtual
   machine.
+* `coreos.inst.save_label` - Comma-separated labels of partitions to
```
jlebon, line 110: If it's too late to change this, that's OK, but I think these would be less ambiguous with `part` somewhere in their names (and the corresponding CLI switches). E.g. `coreos.inst.save_partlabel` and `coreos.inst.save_partindex`.

bgilbert: Not too late, but the problem with the command-line arguments is the usual one: 80-character limit.  And the kargs should probably match the command line.  Any ideas for shorter names?

bgilbert: Okay, updated to `{--,}save[-_]part{label,index}` with shorter parameter variables:

```
        --save-partlabel <lx>...    Save partitions with this label glob
        --save-partindex <id>...    Save partitions with this number or range
```

PR #323, opened 2020-07-28
man: Import initial set of manpages

travier:
The man pages are currently generated from the help text via help2man.

Hopefully this will be replaced by clap generated man pages in the future.


travier:
Unfortunately clap does not come with man page generation support yet (https://github.com/clap-rs/clap/issues/552). I also took a look at the following projects:
* https://github.com/rust-cli/clap-md: requires clap 3.0.0 (unreleased)
* https://github.com/rust-cli/man: requires some significant code change / addition and does not appear to support subcommands

Thus this clearly no great but best effort solution.

travier:
If we are OK with the approach here, I can work on updating the Makefile, etc.

bgilbert:
Man pages are good but I think the use case here is limited.  Fedora CoreOS and RHEL CoreOS don't ship `man(1)`, the man pages aren't useful inside the coreos-installer container, and `cargo install` [won't install them](https://github.com/rust-lang/cargo/issues/2729).  So AFAICT this is basically only useful for the Fedora package on non-CoreOS Editions.

Also, in general I'm not thrilled about checking in generated code.

travier:
Hum, indeed, this is not really useful in this form. I can convert them to HTML to make they viewable but I don't know where we could host them.

travier:
Or create a markdown help page with everything merged in one page, but we would have to keep it updated manually.

cgwalters:
> the man pages aren't useful inside the coreos-installer container

We could introduce a `coreos-tools` container image with e.g. `fcct` and `coreos-installer` and things like man pages for those tools.  And use that for our default `toolbox` image.

> Hum, indeed, this is not really useful in this form. I can convert them to HTML to make they viewable but I don't know where we could host them.

I could imagine an automated job to import to something like https://docs.fedoraproject.org/en-US/fedora-coreos/man ?

bgilbert:
We could add a release checklist step to PR fedora-coreos-docs with the output of a docs generator.  That doesn't seem too burdensome.

Another issue is that the `--help` output isn't really sufficient documentation.  Some of the `coreos-installer install` argument descriptions are fairly contorted to fit into 80 characters.

jlebon:
> Man pages are good but I think the use case here is limited. Fedora CoreOS and RHEL CoreOS don't ship `man(1)`,

FWIW, I personally still think we should ship man pages in FCOS at least. :) It's just a shame that it pulls in Perl, which we've managed to avoid so far.

I like @travier's idea to just keep it in a Markdown page here. If we do want manpages after, there's a bunch of Markdown -> manpage converters out there we could use (at least podman uses one).

travier:
Yes, I'm not a fan of help2man too. I also took a look at what podman is doing and I will try to make Markdown pages as they would be more interesting as a source format.

bgilbert:
Should this be closed in favor of #374?

travier:
Yes!

PR #304, opened 2020-07-20
rdcore: add rootmap command

jlebon:
The `rootmap` command will handle (1) discovering root block device
dependencies, (2) generating appropriate kernel arguments for each block
device, and (3) optionally injecting these kernel arguments into the BLS
config.

This is necessary in order to implement root on complex devices, such as
LUKS or RAID, as described in:

coreos/fedora-coreos-tracker#94

PR #324, opened 2020-07-28
Prep patches for `rdcore rootmap`

jlebon:
Split out of https://github.com/coreos/coreos-installer/pull/304.
See individual commit messages.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -33,6 +33,9 @@ use std::thread::sleep;
 use std::time::Duration;
 
 use crate::errors::*;
+use crate::util::*;
+
+use crate::{runcmd, runcmd_output};
 
 #[derive(Debug)]
 pub struct Disk {
@@ -264,17 +267,7 @@ impl PartTableKpartx {
         // but this blocks indefinitely inside a container.  See e.g.
         //   https://github.com/moby/moby/issues/22025
         // Use -n to skip blocking on udev, and then manually settle.
-        let result = Command::new("kpartx")
-            .arg(flag)
-            .arg("-n")
-            .arg(&self.path)
-            .output()
-            .chain_err(|| format!("running kpartx {} {}", flag, self.path))?;
-        if !result.status.success() {
-            // copy out its stderr
-            eprint!("{}", String::from_utf8_lossy(&*result.stderr));
-            bail!("kpartx {} {} failed: {}", flag, self.path, result.status);
-        }
+        runcmd_output!("kpartx", flag, "-n", &self.path)?;
         udev_settle()?;
         Ok(())
     }
@@ -453,21 +446,14 @@ fn read_sysfs_dev_block_value(maj: u64, min: u64, field: &str) -> Result<String>
 
 pub fn lsblk(dev: &Path) -> Result<Vec<HashMap<String, String>>> {
     // Older lsblk, e.g. in CentOS 7.6, doesn't support PATH, but -p option
-    let result = Command::new("lsblk")
-        .arg("--pairs")
+    let mut cmd = Command::new("lsblk");
+    cmd.arg("--pairs")
         .arg("--paths")
         .arg("--output")
         .arg("NAME,LABEL,FSTYPE,TYPE,MOUNTPOINT")
-        .arg(dev)
-        .output()
-        .chain_err(|| "running lsblk")?;
-    if !result.status.success() {
-        // copy out its stderr
-        eprint!("{}", String::from_utf8_lossy(&*result.stderr));
-        bail!("lsblk of {} failed", dev.display());
-    }
-    let output = String::from_utf8(result.stdout).chain_err(|| "decoding lsblk output")?;
+        .arg(dev);
 
+    let output = cmd_output(&mut cmd).chain_err(|| "running lsblk")?;
     let mut result: Vec<HashMap<String, String>> = Vec::new();
     for line in output.lines() {
         // parse key-value pairs
@@ -610,13 +596,7 @@ pub fn udev_settle() -> Result<()> {
     // our way out of this.
     sleep(Duration::from_millis(200));
 
-    let status = Command::new("udevadm")
-        .arg("settle")
-        .status()
-        .chain_err(|| "running udevadm settle")?;
-    if !status.success() {
-        bail!("udevadm settle failed");
-    }
+    runcmd!("udevadm", "settle")?;
     Ok(())
 }
 
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -90,7 +90,7 @@ impl Disk {
     fn get_partitions(&self, with_holders: bool) -> Result<Vec<Partition>> {
         // walk each device in the output
         let mut result: Vec<Partition> = Vec::new();
-        for devinfo in lsblk(Path::new(&self.path))? {
+        for devinfo in lsblk(Path::new(&self.path), true /* with_deps */)? {
             if let Some(name) = devinfo.get("NAME") {
                 match devinfo.get("TYPE") {
                     // If unknown type, skip.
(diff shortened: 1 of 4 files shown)


src/blockdev.rs
```diff
@@ -31,6 +31,9 @@ use std::time::Duration;
 
 use crate::errors::*;
 use crate::io::resolve_link;
+use crate::utils::*;
```
bgilbert, line 34: Nit: I generally see this spelled `util`.

src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::process::Command;
+
+use crate::errors::*;
+
+#[macro_export]
+macro_rules! runcmd {
+    ($cmd:expr, $($args:expr),*) => {{
+        let mut cmd = Command::new($cmd);
+        $( cmd.arg($args); )*
+        // print full cmdline tried here?
```
bgilbert, line 25: I think that would be better, yes.  `Command`'s `Debug` implementation does exactly this.

src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+        cmd_output(&mut cmd).chain_err(|| format!("running `{}`", $cmd))
+    }}
+}
+
+pub fn cmd_output(cmd: &mut Command) -> Result<String> {
+    let result = cmd.output().chain_err(|| "executing")?;
+    if !result.status.success() {
+        bail!("{}", String::from_utf8_lossy(&result.stderr));
```
bgilbert, line 43: The failure isn't that the program produced stderr, but that it exited nonzero.  We swallow stderr except on failure because some of the programs we call are noisy.

```suggestion
        eprint!("{}", String::from_utf8_lossy(&result.stderr));
        bail!("{:#?} failed: {}", cmd, result.status);
```

src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+    String::from_utf8(result.stdout).chain_err(|| "decoding as UTF-8")
```
bgilbert, line 45:
```suggestion
    String::from_utf8(result.stdout).chain_err(|| format!("decoding output of '{:#?}' as UTF-8", cmd))
```

src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+    ($cmd:expr, $($args:expr),*) => {{
+        let mut cmd = Command::new($cmd);
+        $( cmd.arg($args); )*
+        // print full cmdline tried here?
+        cmd_output(&mut cmd).chain_err(|| format!("running `{}`", $cmd))
+    }}
+}
+
+pub fn cmd_output(cmd: &mut Command) -> Result<String> {
+    let result = cmd.output().chain_err(|| "executing")?;
```
bgilbert, line 41:
```suggestion
    let result = cmd.output().chain_err(|| format!("executing '{:#?}'", cmd))?;
```

src/blockdev.rs
```diff
@@ -446,21 +439,14 @@ fn read_sysfs_dev_block_value(maj: u64, min: u64, field: &str) -> Result<String>
...
-        .chain_err(|| "running lsblk")?;
-    if !result.status.success() {
-        // copy out its stderr
-        eprint!("{}", String::from_utf8_lossy(&*result.stderr));
-        bail!("lsblk of {} failed", dev.display());
-    }
-    let output = String::from_utf8(result.stdout).chain_err(|| "decoding lsblk output")?;
+        .arg(dev);
 
+    let output = cmd_output(&mut cmd).chain_err(|| "running lsblk")?;
```
bgilbert, line 453: Not `runcmd_output!`?

src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use crate::errors::*;
+
+#[macro_export]
+macro_rules! runcmd {
+    ($cmd:expr, $($args:expr),*) => {{
+        let mut cmd = Command::new($cmd);
+        $( cmd.arg($args); )*
+        // print full cmdline tried here?
+        cmd.status().chain_err(|| format!("running `{}`", $cmd))
```
bgilbert, line 26: The callers all assume we fail if not `status().success()`, but we don't.

src/s390x/dasd.rs
```diff
@@ -174,19 +177,10 @@ fn is_formatted(dasd: &str) -> Result<bool> {
...
-        .chain_err(|| format!("executing fdasd -p {}", dasd))?;
-    if !cmd.status.success() {
-        bail!("fdasd -p {} failed", dasd);
-    }
-    Ok(std::str::from_utf8(&cmd.stdout)
-        .chain_err(|| "decoding fdasd output")?
-        .contains("disk label block is invalid"))
+    let mut cmd = Command::new("fdasd");
+    // we're looking for a hardcoded string in the output
+    cmd.env("LC_ALL", "C").arg("-p").arg(dasd);
```
bgilbert, line 182: Nit: why not continue to chain the builder calls directly from `Command::new()`?

src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use error_chain::bail;
+use std::process::Command;
+
+use crate::errors::*;
+
+#[macro_export]
+macro_rules! runcmd {
```
bgilbert, line 21: Each of these functions/macros should have a comment explaining its semantics.

src/blockdev.rs
```diff
@@ -437,15 +437,19 @@ fn read_sysfs_dev_block_value(maj: u64, min: u64, field: &str) -> Result<String>
...
+    // Older lsblk, e.g. in CentOS 7.6, doesn't support PATH, but --paths option
     let mut cmd = Command::new("lsblk");
     cmd.arg("--pairs")
         .arg("--paths")
         .arg("--output")
         .arg("NAME,LABEL,FSTYPE,TYPE,MOUNTPOINT")
         .arg(dev);
 
+    if !with_deps {
+        cmd.arg("--nodeps");
```
bgilbert, line 450: Can't the caller just filter on `TYPE` == `disk`?

bgilbert: Ah, I see you're running this on a partition.  The caller could filter on `NAME`, I guess, but that may not match the argument if a symlink is involved.  I'm okay leaving this as is.

src/utils.rs
jlebon: Oh nice, missed that.

src/utils.rs
jlebon: I just don't like a utility function printing to stderr like this vs just making it part of the error message. IOW, I think I'd prefer

```rust
bail!("{:#?} failed with {}; stderr: {}", cmd, result.status, String::from_utf8_lossy(&result.stderr));
```

Thoughts?

src/blockdev.rs
jlebon: Prep for #304 where one of the arguments will be conditional.

src/utils.rs
bgilbert: In general I agree.  In this case, I don't think of it as "printing to stderr" so much as "conditionally swallowing stderr".  We're just forwarding what the program originally sent there, which might be many lines.

src/s390x/dasd.rs
jlebon: We need to bind the `Command` object to a variable.

src/utils.rs
jlebon: Nice catch!

jlebon:
Updated!

src/s390x/dasd.rs
bgilbert: Whoops, yes, you're right.  `arg()` returns `&mut`.

src/utils.rs
jlebon: OK, adapted. Though in that spirit, I don't think we should decode as UTF-8 at all. We should just dump it as is to `stderr`. Will leave that as a follow-up.

src/util.rs
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use error_chain::bail;
+use std::process::Command;
+
+use crate::errors::*;
+
+/// Runs the provided command. The first macro argument is the executable, and each following
+/// argument are passed to the command. Returns a Result<()> describing whether the command failed.
```
bgilbert, line 21: "each following argument are"

(also below)

jlebon:
Trivial rebase!

PR #333, opened 2020-08-01
*: derive Eq, not just PartialEq

bgilbert:
The affected structs have no fields which can sometimes fail to compare equal to themselves.

Minor cleanup.


bgilbert:
Folded into #336.

PR #335, opened 2020-08-01
blockdev: drop stale comment

bgilbert:
We also reach this case in the install() --preserve-on-error error path.


bgilbert:
Folded into #336.

PR #339, opened 2020-08-02
iso: use copy_file_range to copy the ISO if possible

jlebon:
On filesystems and kernels that support it, use `copy_file_range` to
copy the ISO file. CoW semantics is a good fit here because the
transformations we do on the output file are minimal anyway, so there
isn't really a question of balancing against delayed copy costs.

This brings down the time to inject an Ignition config from 4s to 50ms
on my computer.

To do this, use `openat-ext`'s `FileExt` trait which supports doing
`copy_file_range` on `File` objects. (`std::fs::copy()` also knows
`copy_file_range`, but it only takes file paths, so we'd lose the
`O_EXCL | O_CREAT` semantics we want here).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -154,7 +154,8 @@ dependencies = [
  "hex",
  "libc",
  "maplit",
- "nix",
+ "nix 0.18.0",
+ "openat-ext",
  "openssl",
  "pipe",
  "regex",
@@ -373,7 +374,7 @@ dependencies = [
  "bincode",
  "crc",
  "err-derive",
- "nix",
+ "nix 0.18.0",
  "serde",
  "serde_derive",
 ]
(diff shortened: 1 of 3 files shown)


jlebon:
(And shout out to @cgwalters who wrote that crate!)

bgilbert:
Note that `openat-ext` is not packaged in Fedora.

jlebon:
CI for this needs https://github.com/coreos/coreos-installer/pull/342.

bgilbert:
Fedora package approved: https://bugzilla.redhat.com/show_bug.cgi?id=1877584

I think this is okay to proceed now.

jlebon:
Updated!

bgilbert (approved):

cgwalters:
> This brings down the time to inject an Ignition config from 4s to 50ms on my computer.

Random aside: It's not *just* about the initial timing - on Linux with buffered IO it's mostly measuring the cost of copying it to RAM.  The *real* cost is paid later when writeback kicks in and that 700MB or whatever needs to hit the disk.

jlebon:
> Random aside: It's not just about the initial timing - on Linux with buffered IO it's mostly measuring the cost of copying it to RAM. The real cost is paid later when writeback kicks in and that 700MB or whatever needs to hit the disk.

Didn't the old code pay that cost at `writer.flush()` time?

bgilbert:
> Didn't the old code pay that cost at writer.flush() time?

`flush()` just writes the process' buffers to the kernel; it doesn't `fsync()`.

cgwalters:
> Didn't the old code pay that cost at writer.flush() time?

Not unless `O_DIRECT` is involved, `flush()` is just about the buffering in the process memory (the `BufWriter` object - to tell the kernel to sync out you need e.g. https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_data )

jlebon:
Ahh right, makes sense.

PR #340, opened 2020-08-04
Continuously maintain saved partitions in GPT

bgilbert:
Suppose, by way of example, that Ignition injection fails during an install with saved partitions.  The primary GPT has the following contents over time:

1.  zeroed while performing the download,
2.  image partitions after the download,
3.  image partitions plus saved partitions before mounting Ignition configs,
4.  zeroed after Ignition injection fails,
5.  saved partitions.

Thus, if we fail during a phase where the saved partitions are not on disk, it's important to recover from the failure and rewrite the GPT.  We have good unit tests on the low-level pieces, but the high-level error paths in `install.rs` will rot and there have already been bugs.  Also, there are circumstances we don't handle, such as power loss (of course) and Unix signals.

This happened because partition saving was bolted on, rather than integrated into the core logic.  If we use `SavedPartitions` more broadly, we can ensure the GPT never omits the saved partitions.  In the above example, the GPT now has:

1.  saved partitions while performing the download,
2.  image partitions plus saved partitions after the download,
3.  saved partitions after the Ignition injection fails.

The resulting code is arguably cleaner, even though it has to plumb `SavedPartitions` into the download code.  DASDs use their own download callback, so the only case where we have to bypass `SavedPartitions` for DASDs is in the failure path.

If there are no saved partitions, we still go out of our way not to modify the image's GPT, so in that case we'll continue to leave a stale backup GPT until the OS fixes it on first boot.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -492,38 +492,65 @@ impl Drop for Mount {
 
 #[derive(Debug)]
 pub struct SavedPartitions {
-    sector_size: Option<u64>,
+    sector_size: u64,
     partitions: Vec<(u32, GPTPartitionEntry)>,
 }
 
 impl SavedPartitions {
     pub fn new(disk: &mut File, filters: &[PartitionFilter]) -> Result<Self> {
-        let mut result = Self {
-            sector_size: None,
-            partitions: Vec::new(),
-        };
-
-        // skip reading partition table if we don't have an accept filter
-        if filters.is_empty() {
-            return Ok(result);
-        }
-
-        // read GPT
-        let gpt = match GPT::find_from(disk) {
-            Ok(gpt) => gpt,
+        // try reading GPT
+        let maybe_gpt = match GPT::find_from(disk) {
+            Ok(gpt) => Some(gpt),
             // no GPT on this disk, so no partitions to save
-            Err(gptman::Error::InvalidSignature) => return Ok(result),
-            Err(e) => return Err(e).chain_err(|| "reading partition table")?,
+            Err(gptman::Error::InvalidSignature) => None,
+            Err(e) => return Err(e).chain_err(|| "reading partition table"),
         };
-        Self::verify_disk_sector_size_matches_gpt(disk, gpt.sector_size)?;
-        result.sector_size = Some(gpt.sector_size);
+
+        // determine sector size
+        let sector_size = if disk
+            .metadata()
+            .chain_err(|| "getting disk metadata")?
+            .file_type()
+            .is_block_device()
+        {
+            // We have a real disk, so treat it as the source of truth.
+            get_sector_size(&disk)?.get() as u64
+        } else if let Some(ref gpt) = maybe_gpt {
+            // No real disk, but an existing GPT.  Trust it.
+            gpt.sector_size
+        } else {
+            // We're not reading from a real disk, and there's no existing
+            // GPT.  This should only happen in unit tests.  Pick a default.
+            512
+        };
+
+        // early return if no GPT
+        let gpt = match maybe_gpt {
+            Some(gpt) => gpt,
+            None => {
+                return Ok(Self {
+                    sector_size,
+                    partitions: Vec::new(),
+                })
+            }
+        };
+
+        // cross-check disk and GPT sector sizes if we have both
+        Self::verify_disk_sector_size(disk, gpt.sector_size)?;
 
         // save partitions accepted by filters
-        for (i, p) in gpt.iter() {
-            if Self::matches_filters(i, p, filters) {
-                result.partitions.push((i, p.clone()));
+        let mut partitions = Vec::new();
+        if !filters.is_empty() {
+            for (i, p) in gpt.iter() {
+                if Self::matches_filters(i, p, filters) {
+                    partitions.push((i, p.clone()));
+                }
             }
         }
+        let result = Self {
+            sector_size,
+            partitions,
+        };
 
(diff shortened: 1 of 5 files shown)


src/blockdev.rs
```diff
@@ -492,38 +492,65 @@ impl Drop for Mount {
...
     pub fn new(disk: &mut File, filters: &[PartitionFilter]) -> Result<Self> {
-        let mut result = Self {
-            sector_size: None,
-            partitions: Vec::new(),
+        // try reading GPT
+        let maybe_gpt = match GPT::find_from(disk) {
+            Ok(gpt) => Some(gpt),
+            // no GPT on this disk, so no partitions to save
+            Err(gptman::Error::InvalidSignature) => None,
+            Err(e) => return Err(e).chain_err(|| "reading partition table")?,
```
lucab, line 506: Nit: I think you can have just one between `return` and `?`, plus you can avoid re-wrapping the error:
```
e => return e.chain_err(...),
```

(but I haven't tried this)

src/blockdev.rs
```diff
@@ -492,38 +492,65 @@ impl Drop for Mount {
...
+            .is_block_device()
+        {
+            // We have a real disk, so treat it as the source of truth.
+            get_sector_size(&disk)?.get() as u64
+        } else if let Some(ref gpt) = maybe_gpt {
+            // No real disk, but an existing GPT.  Trust it.
+            gpt.sector_size
+        } else {
+            // We're not reading from a real disk, and there's no existing
+            // GPT.  This should only happen in unit tests.  Pick a default.
```
lucab, line 523: It sounds like this could be inject from the outside, then. The real path would probe-or-fail, while the test cases would directly inject the relevant fake value.

src/blockdev.rs
```diff
@@ -492,38 +492,65 @@ impl Drop for Mount {
...
+        // cross-check disk and GPT sector sizes if we have both
+        Self::verify_disk_sector_size(disk, gpt.sector_size)?;
 
         // save partitions accepted by filters
-        for (i, p) in gpt.iter() {
-            if Self::matches_filters(i, p, filters) {
-                result.partitions.push((i, p.clone()));
+        let mut partitions = Vec::new();
+        if !filters.is_empty() {
+            for (i, p) in gpt.iter() {
```
lucab, line 544: If `gpt` is not used afterward, an `into_iter()` here may work to remove the need of a `clone` below.

src/download.rs
```diff
@@ -282,17 +278,21 @@ pub fn image_copy_default(
     // sparse-copy the image, falling back to non-sparse copy if hardware
     // acceleration is unavailable.  But BLKZEROOUT doesn't support
     // BLKDEV_ZERO_NOFALLBACK, so we'd risk gigabytes of redundant I/O.
-    copy(source, &mut dest).chain_err(|| "decoding and writing image")?;
+    //
+    // Amortize write overhead.  The decompressor will produce bytes in
+    // whatever chunk size it chooses.
+    let mut buf_dest = BufWriter::with_capacity(BUFFER_SIZE, dest);
+    copy(source, &mut buf_dest).chain_err(|| "decoding and writing image")?;
+    // we can't into_inner().chain_err() because of lifetime issues
```
lucab, line 302: If I understood this comment correctly, I think you should be doing something like `.into_inner().map_err(|_| ... )` instead (i.e. no point in bubbling up the bufwriter on error, a human message is enough).

src/blockdev.rs
```diff
@@ -876,6 +877,12 @@ pub fn get_block_device_size(file: &File) -> Result<NonZeroU64> {
     }
 }
 
+/// Get the size of the GPT metadata at the start of the disk.
+pub fn get_gpt_size(file: &mut (impl Read + Seek)) -> Result<u64> {
```
lucab, line 881: Unless I'm missing some case, I think you can directly take a `&GPT` to avoid scanning the disk twice everywhere.

lucab:
Overall it looks like a net gain in regard to safety :+1:.
I've left a few minor comments, but I didn't spot anything suspicious in the new logic flow (although reasoning about all the corner-cases is hard and I may have missed some).

src/download.rs
```diff
@@ -172,17 +173,19 @@ fn write_image_and_sig(
...
 
 /// Copy the image to disk and verify its signature.
+#[allow(clippy::too_many_arguments)]
 pub fn write_image<F>(
     source: &mut ImageSource,
     dest: &mut File,
     dest_path: &Path,
     image_copy: F,
     decompress: bool,
+    saved: Option<&SavedPartitions>,
```
jlebon, line 183: `saved_partitions` ?

bgilbert: `saved` is the variable name everywhere else in the code.

src/blockdev.rs
bgilbert: `GPT` doesn't implement `into_iter()`.

src/blockdev.rs
bgilbert: The type of the `Ok()` part is different, so `e => return e...` doesn't work.

src/blockdev.rs
bgilbert: The callers that aren't in `blockdev.rs` (including the sole caller that isn't a test case) don't have a `&GPT`, and I think it'd be better if they don't have to know about `gptman`.

bgilbert:
Updated.

src/blockdev.rs
bgilbert: Good suggestion, thank you!

PR #341, opened 2020-08-04
iso: add `kargs` commands to manage kernel args

jlebon:
The new commands are `coreos-installer iso kargs modify/reset/show`.
These allow modifying the baked in kernel arguments in a CoreOS live
ISO.

My primary motivation for this is for embedding the right serial console
argument into the ISO before booting it:

```
$ coreos-installer iso kargs modify --append console=ttyS0 my.iso
$ coreos-installer iso kargs show my.iso
mitigations=auto,nosmt ... console=ttyS0
$ cosa run --qemu-iso my.iso -c
...
[    0.000000] Command line: BOOT_IMAGE=/images/vmlinuz ...  console=ttyS0
...
```

This example specifically demonstrates what cosa will do automatically
itself once it can rely on this functionality. This will for example
allow `testiso` tests to be much more useful because we'll actually have
console logs.

But it's of course useful generically. Another way to look at it is that
it mirrors/brings to parity the provisioning capabilities of the live
PXE path.

Co-authored-by: Nikita Dubrovskii <nikita@linux.ibm.com>


jlebon:
The command requires https://github.com/coreos/coreos-assembler/pull/1637 to work so that ISOs have the necessary info in the header, but the PR itself can of course be merged independently.

lucab:
UX concern: it is my understanding that `iso show-kargs` only shows custom-embedded kargs, omitting the baked ones. That may be a surprising semantics for casual users, so it may be good to make it more explicit in the choice of sub-command names.

jlebon:
> UX concern: it is my understanding that iso show-kargs only shows custom-embedded kargs, omitting the baked ones. That may be a surprising semantics for casual users, so it may be good to make it more explicit in the choice of sub-command names.

Yeah, good point. Open to bikeshed on names. :paintbrush: 
@bgilbert suggested splitting out two separate commands:

```
$ coreos-installer iso ignition {embed|show|remove}
$ coreos-installer iso kargs {embed|show|remove}
```

(And deprecate/forward `coreos-installer iso {embed|show|remove}` to the first one.)

Maybe we could make that second command be `coreos-installer iso appended-kargs` or something?

bgilbert:
> Maybe we could make that second command be `coreos-installer iso appended-kargs` or something?

`coreos-installer iso kargs appended` or `show-appended`?

dustymabe:
> > Maybe we could make that second command be `coreos-installer iso appended-kargs` or something?
> 
> `coreos-installer iso kargs appended` or `show-appended`?

I don't mind `show` so much. We could output a message to the user explaining the limitation. Is there any way to show the baked ones even though we can't change them? If so, maybe something like:

```
coreos-installer iso kargs {append|show|show-all|remove}
```

bgilbert:
> We could output a message to the user explaining the limitation.

In help, presumably, not at runtime.

@jlebon Remember to update the subcommand list in `.cci.jenkinsfile`.

bgilbert:
> @bgilbert suggested splitting out two separate commands

https://github.com/coreos/coreos-installer/pull/343

lucab:
> coreos-installer iso kargs {append|show-appended|unappend}

This follows from @dustymabe suggestion. Some approach like this feels good to me because:
 * it is clear that kargs are appended (i.e. postfix-joined) to something else
 * it is clear that it can only show appended entries
 * it is clear that it can only remove previously appended entries

jlebon:
My argument for `coreos-installer iso appended-kargs {embed|show|remove}` is that the subcommands match exactly the ones for `coreos-installer iso ignition`, which I think is a nice property.

@lucab Does that meet the criteria list you have in https://github.com/coreos/coreos-installer/pull/341#issuecomment-669851309?

lucab:
@jlebon yes, not strongly attached to specific keywords as long as the intent is clear.

src/cmdline.rs
```diff
@@ -438,6 +457,40 @@ pub fn parse_args() -> Result<Config> {
                                 .takes_value(true),
                         ),
                 )
+                .subcommand(
+                    SubCommand::with_name("embed-kargs")
+                        .about("Embed additional kernel args in an ISO image")
```
cgwalters, line 462: Maybe `Add kernel arguments to the ISO image`?

One concern: we also have the same use case for the ISO of needing to *remove* the default `nosmt` in FCOS?

src/iso.rs
```diff
@@ -71,6 +72,123 @@ pub fn iso_embed(config: &IsoEmbedConfig) -> Result<()> {
     Ok(())
 }
 
+pub fn iso_embed_kargs(config: &IsoEmbedKargsConfig) -> Result<()> {
+    let mut holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
+
+    // make sure we include trailing newline in calculations below
+    let mut inject_string = config.kargs.clone();
```
cgwalters, line 79: Minor but we could go `mut-less` like:
```
let inject_string = if !inject_string.ends_with('\n') {
  let c = inject_string.clone();
  c.push('\n');
  Cow::Owned(c)
} else {
  Cow::Borrowed(inject_string.as_str())
}
```
or so.

(I know this is pointless here, I just try to avoid having `mut` bindings over long stretches of code unless necessary, and `Cow` :cow2: is cool)

src/iso.rs
```diff
@@ -91,6 +209,35 @@ pub fn iso_show(config: &IsoShowConfig) -> Result<()> {
...
+        .chain_err(|| "seeking to karg area offset")?;
+
+    let mut area = [0u8; COREOS_KARG_EMBED_AREA_MAX_SIZE];
+    file.read_exact(&mut area[..area_size])
+        .chain_err(|| "reading karg embed area")?;
+
+    let s = std::str::from_utf8(&area[..area_size]).chain_err(|| "invalid UTF-8 area")?;
+    let s = s.trim_end_matches('#');
+    if s.is_empty() {
+        bail!("No embedded additional kernel args.");
```
cgwalters, line 230: Is that an error?

src/cmdline.rs
jlebon: The wording here matches the wording used for the `ignition` subcommand. It's more or less `s/embedded Ignition config/embedded additional kernel args/`. Thought that was nice, but not against rewording entirely if folks prefer!

Re. `nosmt`, yeah right now this is limited to managing *additional* kernel args. I think we could redesign things so that it handles base kargs as well, though it's definitely easier technically (and I think somewhat conceptually too) to operate in an "overlay" pattern, again like the `ignition` command. For example, `appended-kargs remove` will change the file back to a binary match of the original file.

I was looking at the kargs we bake in today, and the two big ones that one might want to revert are the `smt` one and the cgroupsv2 one. Both of those can be done by just overriding with another karg (i.e. `mitigations=off systemd.unified_cgroup_hierarchy=1`).

jlebon: Actually hmm, for the `mitigations=off` test, my test might not be valid because I was testing in a VM with a single core & thread anyway. But I definitely observed `cat /sys/devices/system/cpu/vulnerabilities/*` change output. I'll do another sanity-check that SMT can be re-enabled this way.

src/iso.rs
jlebon: This matches the semantics for `iso ignition show`. :)

src/iso.rs
jlebon: Thanks, tweaked! I usually try to avoid `mut` bindings too.

jlebon:
OK, rebased this on top of #343 (that was quite the conflict-fest) and moved to `coreos-installer iso appended-kargs {embed|show|remove}`!

src/cmdline.rs
jlebon: > I'll do another sanity-check that SMT can be re-enabled this way.

OK, so this test was slightly wrong because the default kernel behaviour is `mitigations=auto`, not `off`. But anyway, I did verify using `qemu-kvm -smp threads=2` and `lscpu` that appending `mitigations=auto` is indeed equivalent to removing the `mitigations=auto,nosmt` flag.

bgilbert:
https://github.com/coreos/coreos-assembler/pull/1638 will need to be updated.

jlebon:
> coreos/coreos-assembler#1638 will need to be updated.

Yup, that's part of https://github.com/coreos/coreos-assembler/pull/1637!

If we merge the cosa PR first, then we should be able to see this in action in CI here by inspecting the `testiso` console logs.

jlebon:
How about renaming the command to just `kargs` since it's not just about appended kargs anymore?

nikita-dubrovskii:
> How about renaming the command to just `kargs` since it's not just about appended kargs anymore?

done

cgwalters:
I think given https://github.com/coreos/fedora-coreos-tracker/issues/567 we should probably also support explicitly reconfiguring the console, not just appending to the existing list.

nikita-dubrovskii:
> I think given [coreos/fedora-coreos-tracker#567](https://github.com/coreos/fedora-coreos-tracker/issues/567) we should probably also support explicitly reconfiguring the console, not just appending to the existing list.

didn't get your point. Now we can `karg remove -k "console=X" ISO ; karg add  -k "console=Y" ISO`.

cgwalters:
> didn't get your point. Now we can karg remove -k "console=X" ISO ; karg add -k "console=Y" ISO.

OK, that sounds good.  Is it possible to do in one commandline?  Otherwise people without reflinks are going to feel some pain in rewriting the ISO twice.

It looks like some basic tests could be added here for CI.

nikita-dubrovskii:
> OK, that sounds good. Is it possible to do in one commandline?

not yet

> It looks like some basic tests could be added here for CI.

exactly, but this depends on https://github.com/coreos/coreos-assembler/pull/1637 , so it's better to do a bit later in a separate PR

jlebon:
@nikita-dubrovskii Hope you don't mind, I picked this up again and pushed some refactors on top! I kept your co-ownership in the commit. Some things I did:
- renamed `modify` to `replace`
- used `lazy_static!` for the regex
- removed `embed` since it duplicates `add`
- changed `replace` so that it replaces the karg in-place since ordering matters for kargs (so we can't just remove the old karg and add the new one)
- made the order in which the various commands are handled in code always be `add`, `replace`, `remove`, `reset`, `show` to be consistent
- made `add`, `replace`, and `remove` take multiple kargs instead of just one (on that subject, I think I'd like to change the CLI further so that the commands are just `modify/show/reset` and we then have `modify` be able to take multiple `--add`/`--replace`/`--remove` switches; that way you can do all your modifications in one invocation -- WDYT?)
- made `replace` just support `KEY=OLD=NEW` without having to surround `OLD` and `NEW` with single quotes since it's more natural. I'm guessing you originally did it that way to avoid space escaping issues? Worth noting `rpm-ostree kargs` has been using that format for a while and that hasn't come up yet (and I guess in the worst case if there really is some space escaping issues, users can always fall back to `remove` and `add`)
- standardized all the operations to use the same set of four helper functions: `find_karg_embed_areas`, `get_current_kargs`, `modify_kargs`, and `iso_kargs_write`.
- split out the default file offset as an explicit separate return variable instead of as part of the other file offsets

There are some more tweaks I'd like to do, but the code works well as is!

Requires: https://github.com/coreos/coreos-assembler/pull/1985

nikita-dubrovskii:
> Hope you don't mind, I picked this up again and pushed some refactors on top! 

Np, thx for the updates

>     * removed `embed` since it duplicates `add`

`embed` could be better just to replace any existing `kargs`, but i don't mind 

>     * made `add`, `replace`, and `remove` take multiple kargs instead of just one (on that subject, I think I'd like to change the CLI further so that the commands are just `modify/show/reset` and we then have `modify` be able to take multiple `--add`/`--replace`/`--remove` switches; that way you can do all your modifications in one invocation -- WDYT?)

Hmm, i'm not a fan of many `--xyz` switches, but here it looks reasonable

>     * made `replace` just support `KEY=OLD=NEW` without having to surround `OLD` and `NEW` with single quotes since it's more natural. I'm guessing you originally did it that way to avoid space escaping issues? Worth noting `rpm-ostree kargs` has been using that format for a while and that hasn't come up yet (and I guess in the worst case if there really is some space escaping issues, users can always fall back to `remove` and `add`)

Yep, quotes were used to handle spaces, at least i was testing it. 

LGTM

jlebon:
OK, split out some prep patches into separate commits and simplified it down to having just `modify`, `reset`, and `show`!

jlebon:
To clarify, once we're agreed on the general direction here, my suggestion is to merge https://github.com/coreos/coreos-assembler/pull/1989 and then we should be able to verify in CI here that it's working by looking at the console log output for `kola testiso` tests. I can also add some more tests to exercise `reset` and `show` too.

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+
+fn get_kargs_at_offset(
+    holder: &mut CopiedFileHolder,
+    area_size: usize,
+    offset: u64,
+) -> Result<String> {
+    holder
+        .file
+        .seek(SeekFrom::Start(offset))
+        .chain_err(|| "seeking to karg area offset")?;
```
bgilbert, line 226: Should we report the offset?

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+    let area = {
+        let mut buf = vec![0u8; area_size];
+        holder
+            .file
+            .read_exact(&mut buf)
+            .chain_err(|| "reading karg area")?;
```
bgilbert, line 232: Should we report the offset?

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+        String::from_utf8(buf).chain_err(|| "invalid UTF-8 in karg area")?
+    };
+    Ok(area.trim_end_matches('#').trim().into())
+}
+
+fn write_iso_kargs(
```
bgilbert, line 238: Nit: none of the other new functions say `iso_`.

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+    }
+
+    let mut new_area = vec![b'#'; area_size];
+    new_area[..kargs.len()].copy_from_slice(kargs.as_bytes());
+
+    for &area_offset in area_offsets {
+        holder
+            .file
+            .seek(SeekFrom::Start(area_offset))
+            .chain_err(|| "seeking to karg area offset")?;
```
bgilbert, line 256: Report offset?

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+        holder
+            .file
+            .write_all(&new_area)
+            .chain_err(|| "writing karg embed area")?;
```
bgilbert, line 260: Maybe here too?

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+    }
+
+    holder.complete();
```
bgilbert, line 263: I think this should probably be left to the caller, which might want to mutate the file in other ways also.

As a consequence, maybe the first argument to this function should be a `&mut File`.

src/cmdline.rs
```diff
@@ -554,6 +575,92 @@ pub fn parse_args() -> Result<Config> {
...
             )
+            .subcommand(
+                SubCommand::with_name("kargs")
+                    .about("Modify kernel args in a CoreOS live ISO image")
+                    .subcommand(
+                        SubCommand::with_name("modify")
+                            .about("Modify kernel args in an ISO image")
+                            .arg(
+                                Arg::with_name("append")
+                                    .long("append")
```
bgilbert, line 586: Hmm, no short options for these?

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+            }
+        } else {
+            first_kargs = Some(kargs);
+        }
+    }
+    Ok(first_kargs.expect("at least one area offset"))
+}
+
+fn get_kargs_at_offset(
+    holder: &mut CopiedFileHolder,
```
bgilbert, line 219: Should this take a `&mut File`?

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+
+pub fn iso_kargs_reset(config: &IsoKargsResetConfig) -> Result<()> {
+    let mut holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
+    let (area_size, default_kargs_offset, area_offsets) = find_karg_embed_areas(&mut holder.file)?;
+    let default_kargs = get_kargs_at_offset(&mut holder, area_size, default_kargs_offset)?;
+    write_iso_kargs(&mut holder, area_size, &area_offsets, &default_kargs)
+}
+
+pub fn iso_kargs_show(config: &IsoKargsShowConfig) -> Result<()> {
+    let mut holder = CopiedFileHolder::new(&config.input, None)?;
```
bgilbert, line 186: If the helper functions take a file instead of a holder, we no longer need to create a holder here.

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+            COREOS_KARG_EMBED_AREA_MAX_SIZE
+        );
+    }
+    // default kargs
+    file.read_exact(&mut buf)
+        .chain_err(|| "reading karg embed header")?;
+    let default_kargs_offset: u64 = u64::from_le_bytes(buf);
+    // offsets
+    let mut offsets: Vec<u64> = Vec::new();
+    while offsets.len() < 6 {
```
bgilbert, line 304: Maybe `6` should be a constant?

bgilbert (commented):
LGTM generally!

src/live.rs
```diff
@@ -158,6 +162,163 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+
+    holder.complete();
+    Ok(())
+}
+
+fn find_karg_embed_areas(file: &mut File) -> Result<(usize, u64, Vec<u64>)> {
+    let mut buf: [u8; 8] = [0; 8];
+    // The ISO 9660 System Area is 32 KiB. Karg embed area information is located in the 72 bytes
+    // before the initrd embed area (see EmbedArea below):
+    // 8 bytes: magic string "corekarg"
```
bgilbert, line 271: https://github.com/coreos/coreos-assembler/pull/1996 changes this to `coreKarg`.

jlebon:
Updated and now with a test on the FCOS ISO! Once I moved the embed area-related values into a struct, I refactored things a bit more to just put the functions in the struct's `impl` which made things even nicer!

jlebon:
Cool! We can see https://github.com/coreos/coreos-assembler/pull/1989 in action in the `iso-install` console logs:

```
[    0.000000] Command line: BOOT_IMAGE=/images/pxeboot/vmlinuz initrd=/images/pxeboot/initrd.img,
/images/ignition.img mitigations=auto,nosmt systemd.unified_cgroup_hierarchy=0 coreos.liveiso=fedo
ra-coreos-33.20210108.dev.0 ignition.firstboot ignition.platform.id=metal console=ttyS0
...
```

tests/test-iso-kargs.sh
```diff
@@ -0,0 +1,62 @@
...
+coreos-installer iso kargs modify -d dodo "${iso}"
+
+hash=$(sha256sum "${iso}")
+if [ "${orig_hash}" != "${hash}" ]; then
+    fatal "Hash doesn't match original hash: ${hash} vs ${orig_hash}"
+fi
+
+# Test the largest karg; this technically is a detail of the padding in the
+# distro configs. Here we're assuming an FCOS/RHCOS ISO: 1024 - 2 (2 for the
+# final '\n#' to put "COREOS_KARG_EMBED_AREA on the next line).
```
bgilbert, line 46: :gift:

```
readu8() {
    dd if="${iso}" bs=1 skip="$1" count=8 status=none | od --endian=little -t d8 | head -n1 | awk '{print $2}'
}
embed_size=$(readu8 32680)
embed_used_size=$(dd if="${iso}" bs=1 skip=$(readu8 32688) count=${embed_size} status=none | grep -bo "#" | head -n 1 | cut -f1 -d:)
embed_usable_size=$((${embed_size} - ${embed_used_size} - 1))
```

tests/test-iso-kargs.sh
```diff
@@ -0,0 +1,62 @@
...
+long_karg=$(printf '%*s' 1022 | tr ' ' "k")
+coreos-installer iso kargs modify -a "${long_karg}" "${iso}"
+coreos-installer iso kargs show "${iso}" | grep -q ' k\+$'
+long_karg=$(printf '%*s' 1023 | tr ' ' "k")
+if coreos-installer iso kargs modify -a "${long_karg}" "${iso}" 2>err.txt; then
```
bgilbert, line 51: This doesn't prove much unless we `reset` first.

tests/test-iso-kargs.sh
```diff
@@ -0,0 +1,62 @@
...
+if [ "${orig_hash}" != "${hash}" ]; then
+    fatal "Hash doesn't match original hash: ${hash} vs ${orig_hash}"
+fi
+
+# Test the largest karg; this technically is a detail of the padding in the
+# distro configs. Here we're assuming an FCOS/RHCOS ISO: 1024 - 2 (2 for the
+# final '\n#' to put "COREOS_KARG_EMBED_AREA on the next line).
+long_karg=$(printf '%*s' 1022 | tr ' ' "k")
+coreos-installer iso kargs modify -a "${long_karg}" "${iso}"
+coreos-installer iso kargs show "${iso}" | grep -q ' k\+$'
```
bgilbert, line 49: Let's grep for `${long_karg}` to prove we got the entire argument back.

src/live.rs
```diff
@@ -160,6 +165,197 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+    default_kargs_offset: u64,
+    kargs_offsets: Vec<u64>,
+}
+
+impl<'a> KargEmbedAreas<'a> {
+    fn for_file(file: &'a mut File) -> Result<Self> {
+        let mut buf: [u8; 8] = [0; 8];
+        // The ISO 9660 System Area is 32 KiB. Karg embed area information is located in the 72 bytes
+        // before the initrd embed area (see EmbedArea below):
+        // 8 bytes: magic string "corekarg"
```
bgilbert, line 219: `coreKarg`

src/live.rs
```diff
@@ -160,6 +165,197 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+        }
+
+        let metadata = file.metadata().chain_err(|| "reading metadata for ISO")?;
+        let iso_size = metadata.len();
+
+        // default kargs
+        file.read_exact(&mut buf)
+            .chain_err(|| "reading karg embed header")?;
+        let default_kargs_offset: u64 = u64::from_le_bytes(buf);
+        if default_kargs_offset > iso_size {
```
bgilbert, line 255: We should check that the entire buffer is inside the ISO, so `if default_kargs_offset + length > iso_size`.

src/live.rs
```diff
@@ -160,6 +165,197 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+
+        // offsets
+        let mut kargs_offsets: Vec<u64> = Vec::new();
+        while kargs_offsets.len() < COREOS_KARG_EMBED_AREA_HEADER_MAX_OFFSETS {
+            file.read_exact(&mut buf)
+                .chain_err(|| "reading karg embed header")?;
+            let offset: u64 = u64::from_le_bytes(buf);
+            if offset == 0 {
+                break;
+            } else if offset > iso_size {
```
bgilbert, line 271: Same.

src/live.rs
```diff
@@ -160,6 +165,197 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+    fn get_current_kargs(&mut self) -> Result<String> {
+        // really, we could just get the kargs from the first file, but let's sanity-check that all
+        // the offsets have the same kargs
+        let mut first_kargs: Option<String> = None;
+        for offset in &self.kargs_offsets {
+            let kargs = get_kargs_at_offset(self.file, self.length, *offset)?;
+            if let Some(ref first_kargs) = first_kargs {
+                if &kargs != first_kargs {
+                    bail!(
+                        "kargs don't match at all offsets! (defaults are '{}', but offset {} has: '{}')",
```
bgilbert, line 303: Not actually the default kargs.  `expected '{}'`?

tests/test-iso-kargs.sh
jlebon: Heh thanks! I liked the idea of just getting it from the ISO instead of assuming a value. I ended up taking it a bit further and adding a `--header` hidden switch so we just get it from `coreos-installer` itself.

tests/test-iso-kargs.sh
jlebon: Good catch!

src/live.rs
jlebon: Ahh yup, good catch.

jlebon:
Updated!

bgilbert (approved):

PR #342, opened 2020-08-05
ci/travis: various changes

jlebon:
```
commit 163d0a37e7ff1af97de36458a39172d6bfdeedec
Date:   Wed Aug 5 09:38:27 2020 -0400

    ci/travis: bump MSRV to 1.44.1

    This is in the el8 buildroot now.

commit df8224c979b66a4d954c0624db21709e5b56e924
Date:   Wed Aug 5 09:39:27 2020 -0400

    ci/travis: use MSRV for clippy

    It's not helpful to developers to have to keep around two separate
    toolchain versions for hacking on this codebase. Let's just use the MSRV
    for clippy too. It moves fast enough nowadays too.

commit 92de71b3befe8d71998066f35fdb84461ba66652
Date:   Wed Aug 5 09:42:52 2020 -0400

    ci/travis: also run `rustfmt --check`

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: rust
 rust:
   - 1.44.0  # pinned toolchain for clippy
-  - 1.41.0  # minimum supported toolchain
+  - 1.44.1  # minimum supported toolchain
   - stable
   - beta
   - nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: rust
 rust:
-  - 1.44.0  # pinned toolchain for clippy
-  - 1.44.1  # minimum supported toolchain
+  - 1.44.1  # minimum supported toolchain; used for clippy as well
   - stable
   - beta
   - nightly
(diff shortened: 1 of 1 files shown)


jlebon:
The MSRV bump here should fix CI in https://github.com/coreos/coreos-installer/pull/339.

jlebon:
The trade-off we're making here is that we'll have to deal with any clippy/rustfmt changes at the same time as we bump the MSRV, but OTOH we only have one pin to maintain.

lucab:
> we only have one pin to maintain

I'm not sure this is a real advantage (they are two different things, in theory and in practice), but LGTM anyway.

jlebon:
> I'm not sure this is a real advantage (they are two different things, in theory and in practice), but LGTM anyway.

Right, I don't disagree they're different things. But all we really want from clippy/rustfmt is to not be a moving target, right? Why doesn't the MSRV pin already fulfill that?

I think having a single toolchain version users can `rustup override` when hacking on coreos-installer (and our other Rust projects) is worthwhile.

lucab:
As stated, I'm fine with this specific PR here and I'll proceed to merge it. The comment below applies more generally though, so I'm noting it here for future reference.

> Why doesn't the MSRV pin already fulfill that? I think having a single toolchain version users can rustup override when hacking on coreos-installer (and our other Rust projects) is worthwhile.

From my experience so far this is not a common flow, and personally not very ergonomic. The point of MSRV is to make sure older toolchains keep working; I've contributed to projects which care about retro-compatibility and where that version is more than 2y old at this point, and hypothetically it could go as far back as 1.0.
That version is only for compatibility check though, not to force developers into developing with ancient toolchains. Overall, that would be a big ergonomic loss, as those lack a lot of goodies and performance/productivity boosts.
Most people develop with whatever is the latest stable toolchain, and our goal as maintainers IMHO should be to intercept them there by keeping our linting CI close to that. The pinned MSRV does not (generally) fulfill that, unless you are in a case where retro-compatibility is not actually a real concern, which I'd argue is the case here (1.44.1 is ~50d old currently).

jlebon:
Just circling back to this since I was doing another MSRV bump in rpm-ostree.

>  Most people develop with whatever is the latest stable toolchain, and our goal as maintainers IMHO should be to intercept them there by keeping our linting CI close to that. The pinned MSRV does not (generally) fulfill that, unless you are in a case where retro-compatibility is not actually a real concern, which I'd argue is the case here (1.44.1 is ~50d old currently).

Ack, thanks for expanding on this. I agree with your assessment. I guess we're just lucky in our case because the Rust maintainers in Fedora and RHEL are really great about getting us the latest versions pretty quickly!

PR #343, opened 2020-08-05
Deprecate "iso (.*)" in favor of "iso ignition $1"

bgilbert:
Rename `iso embed` to `iso ignition embed` and similarly for the other `iso` commands.  Rename `iso ignition embed -c/--config` to `-i/--ignition-file` for consistency with the `install` subcommand.  Hide the old subcommands.

See https://github.com/coreos/coreos-installer/pull/341#issuecomment-669224124.


src/live.rs
```diff
@@ -85,13 +99,13 @@ pub fn iso_show(config: &IsoShowConfig) -> Result<()> {
     if buf == embed.new_buffer() {
         bail!("No embedded Ignition config.");
     }
-    let config =
-        String::from_utf8(extract_cpio(&buf)?).chain_err(|| "couldn't decode Ignition config")?;
-    print!("{}", config);
+    stdout()
+        .write_all(&extract_cpio(&buf)?)
+        .chain_err(|| "writing output")?;
```
lucab, line 104: Does this need flushing afterwards?

lucab:
Command shuffling looks good to me, but leaving some space for @jlebon to go through this.

src/live.rs
bgilbert: Yup.  Fixed.

src/live.rs
```diff
@@ -26,7 +26,22 @@ use crate::io::*;
 
 const FILENAME: &str = "config.ign";
 
-pub fn iso_embed(config: &IsoEmbedConfig) -> Result<()> {
+pub fn iso_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
+    eprintln!("iso embed is deprecated; use iso ignition embed.  Continuing.");
```
jlebon, line 30: Minor/optional:

```suggestion
    eprintln!("`iso embed` is deprecated; use `iso ignition embed`.  Continuing.");
```

?

Aside: I'm kinda surprised `clap` doesn't have a built-in deprecation mechanism to emit this warning.

PR #344, opened 2020-08-06
live: add pxe ignition {wrap|unwrap} subcommands

### LINKED ISSUE #322: Consider adding subcommand to wrap an Ignition config in an initrd for appending
# Feature Request #

## Desired Feature ##

The Ignition Dracut glue knows how to pull a user config from an appended initrd.  (See bullet 4 [here](https://docs.fedoraproject.org/en-US/fedora-coreos/migrate-cl/#_operator_notes)).  The process of constructing a compatible cpio.gz is clunky from the command line, but would be trivial as a coreos-installer subcommand.

## Other Information ##

This could be useful for some PXE use cases, notably bootstrapping a custom TLS CA via an HTTPS fetch by an iPXE config that already trusts that CA.

bgilbert:
Add subcommands to generate or show a cpio.xz-wrapped Ignition config for appending to a PXE initrd.

Also switch `iso ignition embed` to using xz (and `iso ignition show` to accepting either gz or xz); it uses storage more efficiently and is supported by both FCOS and RHCOS.

Requires #343.  Fixes #322.

### DIFF
--- a/src/live.rs
+++ b/src/live.rs
@@ -14,9 +14,11 @@
 
 use cpio::{write_cpio, NewcBuilder, NewcReader};
 use error_chain::bail;
+use nix::unistd::isatty;
 use std::convert::TryInto;
-use std::fs::{read, remove_file, File, OpenOptions};
+use std::fs::{read, remove_file, write, File, OpenOptions};
 use std::io::{copy, stdin, stdout, BufReader, BufWriter, Cursor, Read, Seek, SeekFrom, Write};
+use std::os::unix::io::AsRawFd;
 
 use crate::cmdline::*;
 use crate::errors::*;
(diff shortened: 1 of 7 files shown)


src/live.rs
```diff
@@ -112,6 +114,49 @@ pub fn iso_ignition_remove(config: &IsoIgnitionRemoveConfig) -> Result<()> {
...
+        None => {
+            stdout().write_all(&cpio).chain_err(|| "writing output")?;
+            stdout().flush().chain_err(|| "flushing output")?;
+        }
+    }
+    Ok(())
+}
+
+pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
+    let buf = read(&config.input).chain_err(|| format!("reading {}", config.input))?;
```
jlebon, line 152: Can we limit the size here to something reasonable? (I'm imagining someone passing the base initrd by accident instead of the Ignition cpio and we `malloc` 700M).

bgilbert: Discussed OOB.  The large-file case is not great, but it's also not clear where to set a threshold, since people might encode large binaries into their Ignition configs.

PR #358, opened 2020-08-20
rdcore/rootmap: also inject rootflags karg

jlebon:
Because we now always inject a `root` karg, it means that it's systemd
which does the mount. But then, systemd doesn't know to use the
`prjquota` flag we want by default for xfs.

Call out to a `coreos-rootflags` helper where the source of truth for
default root mount flags lives and embed its output if any is provided.

### DIFF
--- a/src/bin/rdcore/rootmap.rs
+++ b/src/bin/rdcore/rootmap.rs
@@ -53,6 +53,12 @@ pub fn rootmap(config: &RootMapConfig) -> Result<()> {
     // systemd-fstab-generator, and it defaults to read-only otherwise
     kargs.push("rw".into());
 
+    let rootflags = runcmd_output!("coreos-rootflags", &config.root_mount)?;
+    let rootflags = rootflags.trim();
+    if !rootflags.is_empty() {
+        kargs.push(format!("rootflags={}", rootflags));
+    }
+
     let boot_mount = if let Some(path) = &config.boot_mount {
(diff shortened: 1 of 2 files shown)


src/bin/rdcore/rootmap.rs
```diff
@@ -53,6 +53,12 @@ pub fn rootmap(config: &RootMapConfig) -> Result<()> {
     // systemd-fstab-generator, and it defaults to read-only otherwise
     kargs.push("rw".into());
 
+    let rootflags = runcmd_output!("coreos-rootflags")?;
```
bgilbert, line 56: We could pass the path to the device node, in case the script wants to do some investigation.

src/bin/rdcore/rootmap.rs
```diff
@@ -53,6 +53,12 @@ pub fn rootmap(config: &RootMapConfig) -> Result<()> {
...
+    let rootflags = rootflags.trim();
```
bgilbert, line 57: Nit: could be combined into previous line.

jlebon: We need to bind string return value before getting an `&str` out of it with `trim()`.

src/bin/rdcore/rootmap.rs
jlebon: Good idea. Slight tweak on your suggestion: we pass it the root mountpoint instead of the device node.

PR #361, opened 2020-08-24
Don't reboot after hook unit failure; document sample custom units

bgilbert:


doc/customizing-install.md
```diff
@@ -0,0 +1,142 @@
...
+systemd:
+  units:
+    - name: install.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Run CoreOS Installer
+        Requires=coreos-installer-pre.target
+        After=network-online.target
+        Wants=network-online.target
```
cgwalters, line 110: This would only be needed for remote Ignition right?

bgilbert: Sure.  I can add a comment.

doc/customizing-install.md
```diff
@@ -0,0 +1,85 @@
...
+variant: fcos
+version: 1.1.0
+
+storage:
+  files:
+    - path: /usr/local/bin/pre-install-hook
+      mode: 0755
+      contents:
+        inline: |
+          #!/bin/bash
```
jlebon, line 24: Let's `set` a good example with `set -euo pipefail`? :)

doc/customizing-install.md
```diff
@@ -0,0 +1,85 @@
...
+          echo "post-hook"
+
+systemd:
+  units:
+    - name: pre-install-hook.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Run before install
+        Before=coreos-installer.service
```
jlebon, line 44: To make the two hooks symmetrical, WDYT about:

1. using `RequiredBy=coreos-installer.target` on both units
2. using `Before=coreos-installer.target` on both units (I think this is actually implied by `RequiredBy=` for target units BTW, but it doesn't hurt either -- see https://www.freedesktop.org/software/systemd/man/systemd.target.html#Default%20Dependencies).

So essentially the only difference is the `Before=` and `After=` for `coreos-installer.service`.

jlebon: Also, should we recommend `After=coreos-installer-pre.target` in both?

doc/customizing-install.md
```diff
@@ -83,3 +83,63 @@ list:
...
+version: 1.1.0
+
+systemd:
+  units:
+    - name: install.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Run CoreOS Installer
+        Requires=coreos-installer-pre.target
```
jlebon, line 108: Also `After` no?

doc/customizing-install.md
bgilbert: `coreos-installer-pre.target` runs so early that we shouldn't need to sequence on it.  The pre-hook needs `RequiredBy=coreos-installer.service` because we shouldn't run the install if a pre-hook fails.

bgilbert: Actually, the single unit currently pulled in by `coreos-installer-pre.target` runs early, but the target is not constrained to do so in future.  I've added `After` to the pre-hook; it's redundant for the post-hook.

bgilbert:
Travis hit an ENOSPC in one s390x builder.  It's not offering a rerun button for some reason, and Travis queues seem to be very long right now.  Merging over red.

PR #371, opened 2020-08-28
rdcore/rootmap: add rd.luks.options=_netdev if needed

jlebon:
If the root device has `_netdev` we need to add that to the kernel
cmdline so systemd knows to order the generated unit after networking.


jlebon:
Noticed this while testing https://github.com/latchset/clevis/pull/224. This is one issue with Tang pinned LUKS devices. Now hitting another issue where it seems like NM is just restarting over and over in a loop when bringing up the network.

cgwalters (approved):

Issue #372, opened 2020-09-01
s390x: Crash observed when installing on zVM with DASD disks


Prashanth684:
Similar to #360, recently installs on zVMs with DASD disks have been failing with : 

```
[core@bootstrap-0 ~]$ sudo ./coreos-installer install /dev/dasda --ignition-url=http://10.16.120.32:8080/ignition/ignitionv3basic.conf --insecure-ignition --firstboot-args "rd.neednet=1 ip=10.16.120.46::10.16.120.254:255.255.255.0:::none nameserver=10.16.120.32"
Installing Red Hat Enterprise Linux CoreOS 46.82.202008301639-0 (Ootpa) s390x (4096-byte sectors)
Error: deserialization failed
Caused by: Segmentation fault
```

This is caused only when building optimized images. When turning `lto` off it works fine. After chasing this down a bit looks like the bump of serde from 114 to 115 (#349 ) seems to have caused this in conjunction with lto being set. This is the commit where the crash starts occurring from.

A backtrace with gdb doesn't reveal much: 

```
(gdb) bt
#0  0x000002aa16ffd69a in <bincode::error::ErrorKind as core::fmt::Display>::fmt ()
#1  0x000002aa1709237a in core::fmt::Formatter::write_fmt ()
#2  0x000002aa170b4d8c in <error_chain::DisplayChain<T> as core::fmt::Display>::fmt ()
#3  0x000002aa1709a876 in core::fmt::write ()
#4  0x000002aa174c0d68 in <std::io::stdio::Stderr as std::io::Write>::write_fmt ()
#5  0x000002aa16feb8bc in coreos_installer::main ()
#6  0x000002aa170b4c22 in std::rt::lang_start::{{closure}} ()
#7  0x000002aa174d0726 in std::panicking::try::do_call ()
#8  0x000002aa1723eb08 in __rust_try ()
#9  0x000002aa170b4774 in std::rt::lang_start ()
#10 0x000003ff7e223462 in __libc_start_main () from /lib64/libc.so.6
#11 0x000002aa16fe6764 in _start ()
```

cc @nikita-dubrovskii @bgilbert

bgilbert:
@nikita-dubrovskii Does the crash fixed by #360 occur with LTO disabled?  If not, I wonder if we should revert #360 and instead disable LTO in the RHEL package on s390x.

Prashanth684:
[...]

PR #373, opened 2020-09-02
Revert "blockdev: use loop instead of iterator"

bgilbert:
Miscompilation also affects other parts of the codebase, and the original code is cleaner.

This reverts #360.  See #372.

cc @nikita-dubrovskii @Prashanth684


Prashanth684:
lgtm

cgwalters (approved):

Issue #327, opened 2020-07-29
Improve documentation via GitHub pages


travier:
# Feature Request #

Provide documentation for easier references via GitHub pages.

travier:
See also: https://github.com/coreos/coreos-installer/issues/207

PR #374, opened 2020-09-04
docs: Add initial GitHub Pages support and command line reference

travier:
This still a WIP but I think it would benefit from a quick look to figure out the next step / what's a priority.

Preview: https://travier.github.io/coreos-installer/

Fixes: https://github.com/coreos/coreos-installer/issues/327


docs/man.md
```diff
@@ -0,0 +1,52 @@
+---
+layout: default
+has_children: true
+nav_order: 4
+has_toc: false
+---
+
+# Command line reference
```
lucab, line 8: I think it would be better to have only the support commands in this user docpage. The others are not really aimed at users, and may contribute in confusing them.

If we want to also cover them for our reference, we can add a few minor pages like `osmet.md` and `rdcore.md` under the "development" section.

docs/man.md
```diff
@@ -0,0 +1,52 @@
...
+
+- [coreos-installer](coreos-installer.md)
+
+- [coreos-installer-install](coreos-installer-install.md)
+- [coreos-installer-download](coreos-installer-download.md)
+- [coreos-installer-list-stream](coreos-installer-list-stream.md)
+
+- [coreos-installer-iso](coreos-installer-iso.md)
+- [coreos-installer-iso-ignition](coreos-installer-iso-ignition.md)
+- [coreos-installer-iso-ignition-embed](coreos-installer-iso-ignition-embed.md)
```
lucab, line 22: I think these three could be just bundled into `iso-ignition.md`. Same for `pxe-ignition-*` sub-commands below.

docs/man/coreos-installer-install.md
```diff
@@ -0,0 +1,41 @@
...
+
+Install Fedora CoreOS or RHEL CoreOS
+
+## Usage
+
+* coreos-installer install [OPTIONS] &lt;device&gt;
+
+## Options
+
+* **-s**, **--stream** &lt;name&gt;  Fedora CoreOS stream
```
bgilbert, line 19: These lines are pretty unreadable when rendered.  Can we use definition lists or tables?

docs/man/coreos-installer-install.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Description
+
+Install Fedora CoreOS or RHEL CoreOS
+
+## Usage
+
+* coreos-installer install [OPTIONS] &lt;device&gt;
+
+## Options
```
bgilbert, line 17: Several of these will need a lot more text, but that can be left for a followup.

docs/customizing-install.md
```diff
@@ -1,3 +1,8 @@
+---
+layout: default
+nav_order: 2
```
bgilbert, line 3: This is an advanced topic and should be much further down the list.

docs/iso-embed-ignition.md
```diff
@@ -1,3 +1,8 @@
+---
+layout: default
+nav_order: 3
```
bgilbert, line 3: This should be grouped under development documentation.

docs/index.md
```diff
@@ -0,0 +1,221 @@
+---
+layout: default
+nav_order: 1
+---
+
+# CoreOS Installer
```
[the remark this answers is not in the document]
travier: The docs directory has to been self standing for the GitHub Pages support. The easiest way to reduce the duplication is to move the instructions from the README.md to the docs. Will do that.

docs/man/coreos-installer-install.md
travier: Converted everything to a format closer to what's currently used by podman as they use it to generate man pages (something that we may want to do even though it's not a priority).

travier:
Updated to take into account review comments. Thanks!

travier:
Updated to improve the command line reference and address other issues.

docs/cmd.md
```diff
@@ -0,0 +1,20 @@
...
+
+# Command line reference
+
+## Usage
+
+**coreos-installer** *subcommand*
+
+## Subcommands
+
+| [install](man/coreos-installer-install.md) | Install Fedora CoreOS or RHEL CoreOS |
```
bgilbert, line 16: These links are all incorrect.

docs/cmd/iso.md
```diff
@@ -0,0 +1,57 @@
+---
+layout: default
+parent: Command line reference
+nav_order: 4
+---
+
+# coreos-installer iso ignition embed
```
bgilbert, line 7: This means the page shows up in the index as `coreos-installer iso ignition embed`, but it should probably be overridden to `coreos-installer iso`.  Same for the PXE page.

bgilbert (commented):
Let's TOC the customizing-install, development, iso, and pxe pages also.

travier:
Updated.

bgilbert (approved):
:pray:

jlebon:
This looks great @travier!

PR #378, opened 2020-09-20
Properly update MBR when installing with saved partitions

bgilbert:
When installing with saved partitions, we were failing to clear the MBR boot code during install and on error, and failing to write the new boot code afterward.  Fixes regressions in #340.

Also update the size of the protective MBR partition after install, for consistency with the GPT.

https://bugzilla.redhat.com/show_bug.cgi?id=1879690


lucab (approved):
LGTM (and thanks for the unit tests)

ashcrow (approved):
:+1:

PR #390, opened 2020-10-07
systemd: add context to systemd-resolved unit order

dustymabe:
Follow up to 139c2db. Add a comment to the open issue for retrying
HTTP requests in coreos-installer.


bgilbert (approved):
Thanks!

PR #391, opened 2020-10-12
Dockerfile: update to F33

bgilbert:


sohankunkerkar (commented):
LGTM
I would appreciate if you could provide some background for this change.

bgilbert:
The Fedora 33 release is due within the next week or two, and we usually base the container on the most recent stable release.  It seems useful to bump the container now so we start building with F33 in CI.

sohankunkerkar (approved):

PR #411, opened 2020-11-24
rdcore: add `kargs` subcommand

jlebon:
We already have the code for modifying kargs in BLS configs. This
command just exposes it.

It's useful generally, but the motivation right now is for the root
filesystem migration script in RHCOS 4.7 which will use this once moved
to the initramfs for openshift/os#448.

### DIFF
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -26,8 +26,8 @@ pub enum Config {
 pub struct KargsConfig {
     pub boot_device: Option<String>,
     pub boot_mount: Option<String>,
-    pub append_kargs: Option<Vec<String>>,
-    pub delete_kargs: Option<Vec<String>>,
+    pub append_kargs: Vec<String>,
+    pub delete_kargs: Vec<String>,
 }
 
 pub struct RootMapConfig {
@@ -153,10 +153,12 @@ fn parse_kargs(matches: &ArgMatches) -> Result<Config> {
         boot_mount: matches.value_of("boot-mount").map(String::from),
         append_kargs: matches
             .values_of("append")
-            .map(|v| v.map(String::from).collect()),
+            .map(|v| v.map(String::from).collect())
+            .unwrap_or_else(Vec::new),
         delete_kargs: matches
             .values_of("delete")
-            .map(|v| v.map(String::from).collect()),
+            .map(|v| v.map(String::from).collect())
+            .unwrap_or_else(Vec::new),
     }))
 }
 
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -18,8 +18,16 @@ use error_chain::bail;
 use libcoreinst::errors::*;
 
 pub enum Config {
+    Kargs(KargsConfig),
     RootMap(RootMapConfig),
     StreamHash(StreamHashConfig),
+}
+
+pub struct KargsConfig {
+    pub boot_device: Option<String>,
+    pub boot_mount: Option<String>,
+    pub append_kargs: Option<Vec<String>>,
+    pub delete_kargs: Option<Vec<String>>,
 }
 
 pub struct RootMapConfig {
@@ -76,6 +84,45 @@ pub fn parse_args() -> Result<Config> {
                 ),
         )
         .subcommand(
+            SubCommand::with_name("kargs")
+                .about("Modify kargs in BLS configs")
+                // see comment block in rootmap command above
+                .arg(
+                    Arg::with_name("boot-device")
+                        .long("boot-device")
+                        .help("Boot device containing BLS entries to modify")
+                        .conflicts_with("boot-mount")
+                        .value_name("DEVPATH")
+                        .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("boot-mount")
+                        .long("boot-mount")
+                        .help("Boot mount containing BLS entries to modify")
+                        .conflicts_with("boot-device")
+                        .value_name("BOOT_MOUNT")
+                        .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("append")
+                        .long("append")
+                        .value_name("ARG")
+                        .help("Append kernel arg")
+                        .takes_value(true)
+                        .number_of_values(1)
+                        .multiple(true),
+                )
+                .arg(
+                    Arg::with_name("delete")
+                        .long("delete")
+                        .value_name("ARG")
+                        .help("Delete kernel arg")
+                        .takes_value(true)
+                        .number_of_values(1)
+                        .multiple(true),
+                ),
+        )
+        .subcommand(
             SubCommand::with_name("stream-hash")
                 .about("Copy data from stdin to stdout, checking piecewise hashes")
                 .arg(
@@ -89,10 +136,28 @@ pub fn parse_args() -> Result<Config> {
         .get_matches();
 
     match app_matches.subcommand() {
+        ("kargs", Some(matches)) => parse_kargs(&matches),
         ("rootmap", Some(matches)) => parse_rootmap(&matches),
         ("stream-hash", Some(matches)) => parse_stream_hash(&matches),
         _ => bail!("unrecognized subcommand"),
     }
+}
+
+fn parse_kargs(matches: &ArgMatches) -> Result<Config> {
+    // we could enforce these via clap's ArgGroup, but I don't like how the --help text looks
+    if !(matches.is_present("boot-device") || matches.is_present("boot-mount")) {
+        bail!("at least one of --boot-device or --boot-mount required");
+    }
+    Ok(Config::Kargs(KargsConfig {
+        boot_device: matches.value_of("boot-device").map(String::from),
+        boot_mount: matches.value_of("boot-mount").map(String::from),
+        append_kargs: matches
+            .values_of("append")
+            .map(|v| v.map(String::from).collect()),
+        delete_kargs: matches
+            .values_of("delete")
+            .map(|v| v.map(String::from).collect()),
+    }))
 }
 
 fn parse_rootmap(matches: &ArgMatches) -> Result<Config> {
--- a/src/bin/rdcore/kargs.rs
+++ b/src/bin/rdcore/kargs.rs
@@ -1,0 +1,43 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use libcoreinst::errors::*;
+use libcoreinst::install::*;
+
+use crate::cmdline::*;
+use crate::rootmap::get_boot_mount_from_cmdline_args;
+
+pub fn kargs(config: &KargsConfig) -> Result<()> {
+    // the unwrap() here is safe because we checked in cmdline that one of them must be provided
+    let mount = get_boot_mount_from_cmdline_args(&config.boot_mount, &config.boot_device)?.unwrap();
+    visit_bls_entry_options(mount.mountpoint(), |orig_options: &str| {
+        let new_options = bls_entry_delete_and_append_kargs(
+            orig_options,
+            config.delete_kargs.as_ref(),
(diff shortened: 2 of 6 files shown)


src/bin/rdcore/cmdline.rs
```diff
@@ -75,6 +83,45 @@ pub fn parse_args() -> Result<Config> {
                         .takes_value(true),
                 ),
         )
+        .subcommand(
+            SubCommand::with_name("kargs")
```
arithx, line 87: WDYT about also exposing an argument for just fetching the kargs from the BLS config? That way we can keep all karg interactions (read, append, delete) inside of calls to `rdcore kargs`.

jlebon: `rdcore kargs` now always prints out the final kargs (after optionally modifying it).

jlebon:
OK, updated this now! Should be ready to go.

arithx (approved):
The general idea for everything makes sense to me. Not proficient enough at Rust to say if there are better ways to do things though 😄

src/bin/rdcore/kargs.rs
```diff
@@ -0,0 +1,58 @@
...
+    let boot_mount = if let Some(path) = &config.boot_mount {
+        Some(Mount::from_existing(path)?)
+    } else if let Some(devpath) = &config.boot_device {
+        let devinfo = lsblk_single(Path::new(devpath))?;
+        let fs = devinfo
+            .get("FSTYPE")
+            .chain_err(|| format!("failed to query filesystem for {}", devpath))?;
+        Some(Mount::try_mount(devpath, fs, mount::MsFlags::empty())?)
+    } else {
+        unreachable!(); // checked in cmdline
```
lucab, line 34: FYI this can take an argument to print, in case you want to drop the comment.

src/install.rs
```diff
@@ -408,9 +419,48 @@ pub fn edit_bls_entries(mountpoint: &Path, f: impl Fn(&str) -> Result<String>) -
...
+    if found_path.is_none() {
+        bail!("Found no BLS entries in {}", config_path.display());
+    }
+
     Ok(())
 }
 
+/// Wrapper around `visit_bls_entry` to specifically visit just the BLS entry's `options` line and
+/// optionally update it if the function returns new content. Errors out if none or more than one
+/// `options` field was found.
```
lucab, line 431: The latter seems to be a valid case according to specs, though.

src/bin/rdcore/kargs.rs
jlebon: I left the comment but made the string useful in case somehow the unreachable is reached.

src/install.rs
jlebon: I think I'd prefer to error out. OSTree never emits multiple `options` and it actually doesn't even support parsing it (later `options` just override earlier ones).

jlebon:
Updated!

src/install.rs
```diff
@@ -293,6 +293,12 @@ pub fn bls_entry_delete_and_append_kargs(
     delete_args: Option<&Vec<String>>,
     append_args: Option<&Vec<String>>,
 ) -> Result<Option<String>> {
+    if (delete_args.is_none() || delete_args.unwrap().is_empty())
+        && (append_args.is_none() || append_args.unwrap().is_empty())
```
bgilbert, line 297: It seems like the `Option<>` wrappers are redundant in the function signature?

src/bin/rdcore/cmdline.rs
```diff
@@ -75,6 +83,45 @@ pub fn parse_args() -> Result<Config> {
...
+                        .long("boot-mount")
+                        .help("Boot mount containing BLS entries to modify")
+                        .conflicts_with("boot-device")
+                        .value_name("BOOT_MOUNT")
+                        .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("append")
+                        .long("append")
+                        .value_name("arg")
```
bgilbert, line 109: I suppose this should be `ARG` for consistency. :stuck_out_tongue_winking_eye:

src/bin/rdcore/kargs.rs
```diff
@@ -0,0 +1,58 @@
...
+        Some(Mount::from_existing(path)?)
+    } else if let Some(devpath) = &config.boot_device {
+        let devinfo = lsblk_single(Path::new(devpath))?;
+        let fs = devinfo
+            .get("FSTYPE")
+            .chain_err(|| format!("failed to query filesystem for {}", devpath))?;
+        Some(Mount::try_mount(devpath, fs, mount::MsFlags::empty())?)
+    } else {
+        unreachable!("neither --boot-mount nor --boot-device specified"); // checked in cmdline
+    };
```
bgilbert, line 35: Can we move this into a helper function shared with `rootmap()`?

bgilbert (commented):
LGTM generally.  Needs rebase to pick up #422.

src/install.rs
```diff
@@ -383,11 +385,22 @@ pub fn edit_bls_entries(mountpoint: &Path, f: impl Fn(&str) -> Result<String>) -
...
+        let maybe_previous_path = found_path.replace(path);
+        let path = found_path.as_ref().unwrap();
+
+        if let Some(previous_path) = maybe_previous_path {
+            bail!(
+                "Found more than one BLS entry: {} and {}",
+                previous_path.display(),
+                path.display()
+            );
+        }
```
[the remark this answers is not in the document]
jlebon: Sure WFM. I just really like `Option::replace()`. :)

src/install.rs
```diff
@@ -436,6 +420,41 @@ pub fn visit_bls_entry(
...
+            } else if found_options {
+                bail!("Multiple 'options' lines found");
+            } else {
+                let r = f(line["options ".len()..].trim()).chain_err(|| "visiting options")?;
+                if let Some(new_options) = r {
+                    new_contents.push_str("options ");
+                    new_contents.push_str(new_options.trim());
+                } else {
+                    // caller isn't interested in modifying options; so we're done
+                    return Ok(None);
```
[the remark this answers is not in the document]
jlebon: Good point. It's easy enough to add, so I'll add it!

src/install.rs
jlebon: You mean, canonicalize `None` to empty `Vec`s? Yeah, I think we could do that with an `unwrap_or_else()` in the two `cmdline.rs` code which deal with this.

Let me do that in a follow-up.

src/install.rs
```diff
@@ -436,6 +420,41 @@ pub fn visit_bls_entry(
     Ok(())
 }
 
+/// Wrapper around `visit_bls_entry` to specifically visit just the BLS entry's `options` line and
+/// optionally update it if the function returns new content. Errors out if none or more than one
+/// `options` field was found.
+pub fn visit_bls_entry_options(
```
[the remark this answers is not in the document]
bgilbert: Yeah, I see your point.  Counterpoint: it makes our naming inconsistent.  We're passing `bls_entry_delete_and_append_kargs` to `visit_bls_entry_options`, but `kargs` isn't a subset of `options`, it's exactly `options`.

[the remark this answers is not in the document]
jlebon: I renamed `bls_entry_delete_and_append_kargs` to `bls_entry_delete_and_append_options_kargs` to be more explicit. Hope that scratches that itch at least a bit. If you still feel strongly about it, I can change the function name.

src/install.rs
jlebon: Did this in this PR now!

jlebon:
Comments addressed!

src/install.rs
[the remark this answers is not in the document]
bgilbert: I can live with `options_kargs` but we should bind those terms together in the visitor, not the callback.  `bls_entry_delete_and_append_options_kargs` implies that `options` describes some sort of property of the kargs, but actually it's just a synonym for kargs.  In fact, that function no longer cares about BLS either, so it should probably just be `delete_and_append_kargs`.  The visitor can then be `visit_bls_entry_options_kargs`, which makes some sense as it's visiting the kargs inside the `options` field.

src/cmdline.rs
```diff
@@ -49,8 +49,8 @@ pub struct InstallConfig {
     pub ignition_hash: Option<IgnitionHash>,
     pub platform: Option<String>,
     pub firstboot_kargs: Option<String>,
-    pub append_kargs: Option<Vec<String>>,
-    pub delete_kargs: Option<Vec<String>>,
+    pub append_kargs: Vec<String>,
+    pub delete_kargs: Vec<String>,
```
bgilbert, line 53: I was imagining the `unwrap_or_else()` would happen in the immediate caller of `bls_entry_delete_and_append_kargs`, not in the command-line parser, for consistency.  But it's a nit.

src/install.rs
```diff
@@ -408,9 +413,51 @@ pub fn edit_bls_entries(mountpoint: &Path, f: impl Fn(&str) -> Result<String>) -
...
+                }
+                found_options = true;
+            }
+            new_contents.push('\n');
+        }
+        if !found_options {
+            bail!("Couldn't locate 'options' line");
+        }
+        if modified {
+            Ok(None)
```
bgilbert, line 454: Logic is inverted.

src/install.rs
```diff
@@ -345,56 +331,75 @@ fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
...
+            continue;
+        }
+
+        if let Some(previous_path) = found_path {
+            bail!(
+                "Found more than one BLS entry: {} and {}",
+                previous_path.display(),
+                path.display()
+            );
+        }
```
arithx, line 382: Do we actually need to bail if there is more than one BLS entry? Trying to use this in the upgrade script where we'll always have more than 1 BLS entry.

src/cmdline.rs
jlebon: I did it in the cmdline code so as to avoid the awkward/ambiguous `Option<Vec<String>>` type entirely.

src/install.rs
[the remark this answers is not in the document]
jlebon: I think the naming should convey though that the callbacks are meant to be used by those visitors. They implement a specific signature after all.

How about this:
- `visit_bls_entry`: generic visitor for the full contents
- `visit_bls_entry_options`: visitor for just the `options` key
- `bls_entry_*`: callbacks meant to be used with those visitors (with those meant for `visit_bls_entry_options` using `bls_entry_options_...`)

src/install.rs
jlebon: Ouch, yes good catch. In basically all our cases *except* this one we really expect a single BLS entry. Rather than making it configurable, I just relaxed it so we don't really care how many BLS entries there. But instead, we make sure to only act on the latest/default BLS entry, which is what we want for `rdcore kargs`.

jlebon:
Updated!

src/install.rs
bgilbert: In the case of the RAID conversion, any boot entries we don't rewrite will no longer be bootable.  Are we okay with that, or should we think about rewriting all of the entries?

jlebon: Do we support doing the RAID conversion post-firstboot, though?

bgilbert: Sorry, I meant to say the RHCOS LUKS conversion.  And yes, that will only happen post-firstboot.

jlebon: Ahhh OK. Hmm, I lean towards accepting that risk. We don't even know if the previous deployments could even support the changes by just applying the same kargs (e.g. a user could've pinned a 6 month old deployment without necessary patches!). The only deployment we know supports the new configuration is the one we're running from, since it contains the code driving the modification. So IMO it only really makes sense to modify that deployment's BLS config.

(So actually, one thing worth pointing out here is that `rdcore kargs` assumes that you're booted in the latest/default BLS entry, because that's the one it's modifying. If a human does an `rpm-ostree deploy` and then `rpm-ostree rollback` to put the new OSTree with migration code in second place, rebooting into that second entry by manually selecting it at the boot menu will affect the wrong BLS config... we could catch that I think, though not sure how much we should care about that.)

bgilbert:
The last commit also renames `bls_entry_write_platform` but the commit message doesn't mention it.

src/install.rs
bgilbert: 1. Okay, I'm okay with that.
2. Oof.  Well, we'd be affecting a BLS config other than the one we're booting, but is that actually the "wrong" config?  The user has opted to manually boot from a non-current config, but arguably that doesn't mean we should switch to maintaining an older one.

bgilbert (approved):
LGTM pending update to the last commit message.  Thanks!

jlebon:
Fixed commit message to make it more generic!

bgilbert (approved):

src/install.rs
jlebon: It's the "wrong" config in the sense that we don't even know if the modifications we're making will work with the default deployment. But the whole problem is mainly theoretical; I don't think it's reasonable for migration code to handle the case where they're not the default config.

Issue #413, opened 2020-11-25
Delete kargs by regex


YuviGold:
# Feature Request #

## Desired Feature ##
`install --delete-kargs` should support regular expressions.

For example `--delete-kargs 'console=tty[\\w,]+'` to delete all console kargs.

bgilbert:
`--save-partlabel` supports glob patterns, so if we were going to do this, I think globs would be better for consistency.

jlebon:
Hi @YuviGold, we're going to be [removing all the `console` kargs](https://github.com/coreos/fedora-coreos-tracker/issues/567) from the metal images, so the original motivation for this enhancement should go away.

Tentatively closing.

PR #414, opened 2020-11-25
Add regex support for install delete-kargs

YuviGold:
Resolves https://github.com/coreos/coreos-installer/issues/413


jlebon:
https://github.com/coreos/coreos-installer/issues/413#issuecomment-891958618

PR #425, opened 2020-12-09
s390x: trigger 'udevadm settle' after checking DASD's validity

nikita-dubrovskii:
`fdasd -p` also needs `udevadm settle` to avoid possible race condition.
This a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1900699

### DIFF
--- a/src/s390x/dasd.rs
+++ b/src/s390x/dasd.rs
@@ -181,7 +181,15 @@ fn is_invalid(dasd: &str) -> Result<bool> {
     let mut cmd = Command::new("fdasd");
     // we're looking for a hardcoded string in the output
     cmd.env("LC_ALL", "C").arg("-p").arg(dasd);
-    Ok(cmd_output(&mut cmd)?.contains("disk label block is invalid"))
+    let invalid = cmd_output(&mut cmd)?.contains("disk label block is invalid");
+    // Older versions of `fdasd` open the device O_RDWR, which causes udev
+    // to re-probe the device node.  This can cause the fdasd call in
+    // default_format() to fail on 'Error while rereading partition table' or
+    // 'Disk in use'.  To avoid this, wait for udev to settle.
+    // https://bugzilla.redhat.com/1900699
+    // Fixed by https://github.com/ibm-s390-tools/s390-tools/commit/3d74c53
+    udev_settle()?;
+    Ok(invalid)
 }
 
 /// Perform low-level format. This step is necessary before any further disk usage


jlebon:
> `fdasd -p` also needs `udevadm settle`, because it opens the device RW.

Hmm, I guess an RFE here is to make `fdasd` open ro for read-only operations. :) 

Does this fix an issue you've encountered? I'm surprised this would be even required if it's not actually making any modifications.

src/s390x/dasd.rs
```diff
@@ -181,7 +181,9 @@ fn is_invalid(dasd: &str) -> Result<bool> {
     let mut cmd = Command::new("fdasd");
     // we're looking for a hardcoded string in the output
     cmd.env("LC_ALL", "C").arg("-p").arg(dasd);
-    Ok(cmd_output(&mut cmd)?.contains("disk label block is invalid"))
+    let invalid = cmd_output(&mut cmd)?.contains("disk label block is invalid");
+    udev_settle()?;
```
bgilbert, line 185: Can we also add a comment explaining why this is needed?

bgilbert (commented):
The commit message could use a link to the BZ and some background on why this is the right fix.

nikita-dubrovskii:
> Hmm, I guess an RFE here is to make `fdasd` open ro for read-only operations. :)

Yep. This work is planned for `s390-utils` team. But as a workaround to mitigate possible race we better add extra call to `udev`.

src/s390x/dasd.rs
nikita-dubrovskii: Done

hoeppnerj:
Hi there, fyi, we've fixed the RW issue upstream a while ago with this commit: https://github.com/ibm-s390-tools/s390-tools/commit/3d74c534bfec942777ddbb106e3b2b80036d32ea

nikita-dubrovskii:
>Hi there, fyi, we've fixed the RW issue upstream a while ago with this commit: ibm-s390-tools/s390-tools@3d74c53 (It fixed something else at the time, but is related to the fdasd -p issue)

Ok, thx for the information. Missed this it in the discussion we had in Slack.

src/s390x/dasd.rs
```diff
@@ -181,7 +181,13 @@ fn is_invalid(dasd: &str) -> Result<bool> {
     let mut cmd = Command::new("fdasd");
     // we're looking for a hardcoded string in the output
     cmd.env("LC_ALL", "C").arg("-p").arg(dasd);
-    Ok(cmd_output(&mut cmd)?.contains("disk label block is invalid"))
+    let invalid = cmd_output(&mut cmd)?.contains("disk label block is invalid");
+    // It may happen, that we have low-level formatted, but invalid DASD disk, so we'd have
+    // to auto-format it. Sometimes this auto-format fails, and we end up with
+    // 'Error while rereading partition table' or 'Disk in use'.
+    // To avoid possible race-condition between subsequent `fdasd -p / fdasd -a`:
```
bgilbert, line 188: How about this?  Is it correct?

```suggestion
    // Older versions of `fdasd` open the device O_RDWR, which causes udev
    // to re-probe the device node.  This can cause the fdasd call in
    // default_format() to fail on 'Error while rereading partition table' or
    // 'Disk in use'.  To avoid this, wait for udev to settle.
    // https://bugzilla.redhat.com/1900699
    // Fixed by https://github.com/ibm-s390-tools/s390-tools/commit/3d74c53
```

nikita-dubrovskii: done

bgilbert (approved):

PR #430, opened 2020-12-16
s390x: use recommended blocksize of 4096 bytes for unformatted ECKD DASD disks

nikita-dubrovskii:
This a  fix for:
https://bugzilla.redhat.com/show_bug.cgi?id=1905159

Unformatted (low-level) ECKD DASD's sector size is 512 bytes by default:
```
$ fdasd -p /dev/dasdb
fdasd error:  Unsupported disk format
$ ioctl_sector_size
/dev/dasdb: 512
```
So `coreos-installer` chooses 512 bytes-sector `osmet` image.

After formatting (`dasdfmt`) we end up with 4096 bytes sector size, and therefore `coreos-installer` cannot install previously selected `osmet` image

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -23,6 +23,8 @@ use crate::blockdev::*;
 use crate::download::*;
 use crate::errors::*;
 use crate::io::IgnitionHash;
+#[cfg(target_arch = "s390x")]
+use crate::s390x::dasd::dasd_try_get_sector_size;
 use crate::source::*;
 
 pub enum Config {
@@ -733,7 +735,19 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         .value_of("architecture")
         .expect("architecture missing");
 
-    let sector_size = get_sector_size_for_path(Path::new(&device))
+    // Uninitialized ECKD DASD's blocksize is 512, but after formatting
+    // it changes to the recommended 4096
+    // https://bugzilla.redhat.com/show_bug.cgi?id=1905159
+    #[allow(clippy::match_bool, clippy::match_single_binding)]
+    let sector_size = match is_dasd(&device)
+        .chain_err(|| format!("checking whether {} is an IBM DASD disk", device))?
+    {
+        #[cfg(target_arch = "s390x")]
+        true => dasd_try_get_sector_size(&device).transpose(),
+        _ => None,
+    };
+    let sector_size = sector_size
+        .unwrap_or_else(|| get_sector_size_for_path(Path::new(&device)))
         .chain_err(|| format!("getting sector size of {}", &device))?
         .get();
 
--- a/src/install.rs
+++ b/src/install.rs
@@ -14,7 +14,7 @@
 
 use error_chain::{bail, ChainedError};
 use nix::mount;
-use std::fs::{canonicalize, copy as fscopy, create_dir_all, read_dir, File, OpenOptions};
+use std::fs::{copy as fscopy, create_dir_all, read_dir, File, OpenOptions};
 use std::io::{copy, Read, Seek, SeekFrom, Write};
 use std::os::unix::fs::FileTypeExt;
 use std::path::Path;
@@ -46,14 +46,14 @@ pub fn install(config: &InstallConfig) -> Result<()> {
 
     #[cfg(target_arch = "s390x")]
     {
-        if is_dasd(config)? {
+        if is_dasd(&config.device)? {
             if !config.save_partitions.is_empty() {
                 // The user requested partition saving, but SavedPartitions
                 // doesn't understand DASD VTOCs and won't find any partitions
                 // to save.
                 bail!("saving DASD partitions is not supported");
             }
-            s390x::prepare_dasd(&config)?;
+            s390x::prepare_dasd(&config.device)?;
         }
     }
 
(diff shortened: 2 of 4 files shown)


src/install.rs
```diff
@@ -499,9 +496,9 @@ fn stash_saved_partitions(disk: &mut File, saved: &SavedPartitions) -> Result<()
     Ok(())
 }
 
-fn is_dasd(config: &InstallConfig) -> Result<bool> {
-    let target = canonicalize(&config.device)
-        .chain_err(|| format!("getting absolute path to {}", config.device))?;
+pub fn is_dasd(device: &str) -> Result<bool> {
```
bgilbert, line 499: Let's move this to `blockdev.rs`.

src/cmdline.rs
```diff
@@ -733,6 +737,16 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         .value_of("architecture")
         .expect("architecture missing");
 
+    // Uninitialized ECKD DASD's blocksize is 512, but after formatting
+    // it changes to the recommended 4096
+    // https://bugzilla.redhat.com/show_bug.cgi?id=1905159
+    #[cfg(target_arch = "s390x")]
+    {
+        if is_dasd(&device)? {
+            s390x::prepare_dasd(&device)?;
```
bgilbert, line 746: I'm not very comfortable writing to the target disk from inside the command-line parser.  I recognize that formatting an unformatted disk isn't very destructive, but even so -- if we fail on a bad command-line argument, it's fair to expect that the target is unchanged.

Instead, how about we add a sector size query function in `dasd.rs` that returns `Result<Option<NonZeroU32>>`, call it here, and if it returns `Ok(Some(_))`, accept its answer instead of calling `get_sector_size()`?  That function can just call `is_formatted()` and return `4096` if false.  In other words, if we know we're going to low-level format the disk, return the sector size we plan to select when formatting.

(And by `4096` I mean a named constant.)

nikita-dubrovskii: I was thinking about using 4096 by default, but:
- `low-level format` takes time, for 20Gb disk about 2 minutes on my machine, so after those 2 minutes all opened network connections are closed, and `installer` exits with failure: `Caused by: Broken pipe (os error 32)`
- ECKD should be ready for use, so this case shouldn't happen on customer side . But if ECKD is `unformatted`, it's better to touch and initialize it ASAP
- FBA/ECKD could be 512 and 4096 (4096 is recommended and better for ECKD)
- FBA doesn't require low-level format

@bgilbert  , what do you think - should i modify this PR now or after FBA is merged?

bgilbert: Let's go ahead and fix this bug, before FBA merges.  It seems like we only need to special-case unformatted ECKD DASDs, and everything else can use `get_sector_size()`?

And let's defer the network timeout issue for now.  AIUI osmet works for DASDs, so there's a workaround at least.  I've filed https://github.com/coreos/coreos-installer/issues/431 for that issue.

src/install.rs
nikita-dubrovskii: done

src/cmdline.rs
nikita-dubrovskii: done

src/s390x/dasd.rs
```diff
@@ -33,18 +32,30 @@ use crate::runcmd;
 // IBM DASD Support
 /////////////////////////////////////////////////////////////////////////////
 
+const ECKD_DASD_BLOCKSIZE: u32 = 4096;
```
bgilbert, line 35: This constant should be used in `low_level_format()` also.

src/s390x/dasd.rs
```diff
@@ -33,18 +32,30 @@ use crate::runcmd;
...
 }
 
-pub fn prepare_dasd(config: &InstallConfig) -> Result<()> {
-    low_level_format(&config.device)?;
-    if is_invalid(&config.device)? {
-        eprintln!("Disk {} is invalid, formatting", &config.device);
-        default_format(&config.device)?
+/// Returns default block size ony for unformatted ECKD DASD disk,
+/// otherwise None
+pub fn unformatted_sector_size(device: &str) -> Result<Option<NonZeroU32>> {
```
bgilbert, line 46: The function name implies that the caller should already know whether the device is formatted, but that's this function's job.  What about something like `dasd_try_get_sector_size()`?

src/s390x/dasd.rs
```diff
@@ -33,18 +32,30 @@ use crate::runcmd;
...
     length: u64,
 }
 
-pub fn prepare_dasd(config: &InstallConfig) -> Result<()> {
-    low_level_format(&config.device)?;
-    if is_invalid(&config.device)? {
-        eprintln!("Disk {} is invalid, formatting", &config.device);
-        default_format(&config.device)?
+/// Returns default block size ony for unformatted ECKD DASD disk,
+/// otherwise None
```
bgilbert, line 45: This doesn't really explain how the caller should use it.  How about something like

```suggestion
/// Returns expected sector size if this is a DASD we'll format later,
/// or None if the caller should use get_sector_size()
```

src/cmdline.rs
```diff
@@ -733,7 +735,17 @@ fn parse_install(matches: &ArgMatches) -> Result<Config> {
         .value_of("architecture")
         .expect("architecture missing");
 
-    let sector_size = get_sector_size_for_path(Path::new(&device))
+    // Uninitialized ECKD DASD's blocksize is 512, but after formatting
+    // it changes to the recommended 4096
+    // https://bugzilla.redhat.com/show_bug.cgi?id=1905159
+    #[allow(clippy::match_bool, clippy::match_single_binding)]
+    let sector_size = match is_dasd(&device)? {
```
bgilbert, line 742: `chain_err()`

bgilbert (commented):
LGTM generally.

bgilbert (approved):

PR #432, opened 2020-12-22
iso: stream iso ignition embed to stdout

### LINKED ISSUE #382: send output of "iso ignition embed" to stdout
# Feature Request #
Enable the `coreos-installer iso ignition embed` command to send its output to stdout so that the caller can send that data to its final destination (through whatever means are appropriate to its use case) without needing to write the entire ISO out to the local filesystem.

## Use Case ##
As an example use case, the OpenShift [assisted-service](https://github.com/openshift/assisted-service/) makes RHCOS-based live ISOs available for download over HTTP with embedded ignition that is specific to a user. Rather than a three-step approach that A) makes an entire copy on disk of the ISO, B) streams the resulting file contents via HTTP, then C) deletes that file, it could combine those steps by streaming stdout from the `coreos-installer iso ignition embed` command.

This is particularly valuable when running in a live ISO. In that case, the local filesystem is backed by RAM, so the impact of making a copy of the ISO can be significant. OpenShift's [assisted-service](https://github.com/openshift/assisted-service/) will be delivered to users for [on-premise use as a live ISO](https://github.com/openshift/enhancements/pull/411/files).

## Other Information ##
For syntax, one option is to just support `-` as a valid value for `--output`.

`coreos-installer iso ignition embed -o -`

djzager:
When given the `-o -` option, coreos-installer will stream the resulting
ISO to stdout.

Fixes #382

### DIFF
--- a/src/live.rs
+++ b/src/live.rs
@@ -19,7 +19,7 @@ use openat_ext::FileExt;
 use serde::Serialize;
 use std::convert::TryInto;
 use std::fs::{read, remove_file, write, File, OpenOptions};
-use std::io::{stdin, stdout, BufReader, Cursor, Read, Seek, SeekFrom, Write};
+use std::io::{copy, stdin, stdout, BufReader, BufWriter, Cursor, Read, Seek, SeekFrom, Write};
 use std::os::unix::io::AsRawFd;
 
 use crate::cmdline::*;
@@ -51,8 +51,10 @@ pub fn iso_remove(config: &IsoIgnitionRemoveConfig) -> Result<()> {
 }
 
 pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
-    let mut holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
-    let mut embed = EmbedArea::for_file(&mut holder.file)?;
+    let use_stdout = config.output.as_deref() == Some("-");
+    if use_stdout && isatty(stdout().as_raw_fd()).chain_err(|| "checking if stdout is a TTY")? {
+        bail!("Refusing to write binary data to terminal");
+    }
 
     let ignition = match config.ignition {
         Some(ref ignition_path) => {
@@ -66,6 +68,23 @@ pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
             data
         }
     };
+
+    let mut holder: CopiedFileHolder;
+    let mut embed: EmbedArea;
+
+    if use_stdout {
+        holder = CopiedFileHolder {
+            file: OpenOptions::new()
+                .read(true)
+                .open(&config.input)
+                .chain_err(|| format!("opening {}", &config.input))?,
+            copied_path: None,
+            complete: false,
+        };
+    } else {
+        holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
+    }
+    embed = EmbedArea::for_file(&mut holder.file)?;
 
     let cpio = make_cpio(&ignition)?;
     if cpio.len() > embed.length {
@@ -85,12 +104,18 @@ pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
             bail!("This ISO image already has an embedded Ignition config; use -f to force.");
         }
     }
-    // delete any existing config
-    embed.clear()?;
-    // write new config
-    embed.seek_to_start()?;
-    embed.write(&cpio)?;
-    holder.complete();
+
+    if use_stdout {
+        embed.stream(&cpio, &mut stdout())?;
+    } else {
+        // delete any existing config
+        embed.clear()?;
+        // write new config
+        embed.seek_to_start()?;
+        embed.write(&cpio)?;
+        holder.complete()
+    }
+
     Ok(())
 }
 
@@ -116,10 +141,25 @@ pub fn iso_ignition_show(config: &IsoIgnitionShowConfig) -> Result<()> {
 }
 
 pub fn iso_ignition_remove(config: &IsoIgnitionRemoveConfig) -> Result<()> {
-    let mut holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
-    let mut embed = EmbedArea::for_file(&mut holder.file)?;
-    embed.clear()?;
-    holder.complete();
+    let use_stdout = config.output.as_deref() == Some("-");
+    if use_stdout && isatty(stdout().as_raw_fd()).chain_err(|| "checking if stdout is a TTY")? {
+        bail!("Refusing to write binary data to terminal");
+    }
+
+    if use_stdout {
+        let mut file = OpenOptions::new()
+            .read(true)
+            .open(&config.input)
+            .chain_err(|| format!("opening {}", &config.input))?;
+        let mut embed = EmbedArea::for_file(&mut file)?;
+        embed.stream(&[], &mut stdout())?;
+    } else {
+        let mut holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
+        let mut embed = EmbedArea::for_file(&mut holder.file)?;
+
+        embed.clear()?;
+        holder.complete();
+    }
     Ok(())
 }
 
@@ -490,6 +530,32 @@ impl<'a> EmbedArea<'a> {
         self.file
             .write_all(buf)
             .chain_err(|| "writing embed area")?;
+        Ok(())
+    }
+
+    fn stream(&mut self, cpio: &[u8], writer: &mut (impl Write + ?Sized)) -> Result<()> {
+        let mut buf = [0u8; BUFFER_SIZE];
+        self.file
+            .seek(SeekFrom::Start(0))
+            .chain_err(|| "seeking to start")?;
+        copy_exactly_n(&mut self.file, writer, self.offset, &mut buf)
+            .chain_err(|| "copying file")?;
+        if cpio.len() > self.length {
+            bail!("buffer larger than embed area");
+        }
+        writer.write_all(cpio).chain_err(|| "writing embed area")?;
+        let zeroes = vec![0; self.length - cpio.len()];
+        writer.write_all(&zeroes).chain_err(|| "writing zeros")?;
+        self.file
+            .seek(SeekFrom::Start(self.offset + self.length as u64))
+            .chain_err(|| "seeking to end of embed area")?;
+        let mut write_buf = BufWriter::with_capacity(BUFFER_SIZE, writer);
+        copy(
+            &mut BufReader::with_capacity(BUFFER_SIZE, self.file),
+            &mut write_buf,
+        )
+        .chain_err(|| "copying file")?;
+        write_buf.flush().chain_err(|| "flushing output")?;
         Ok(())
     }
 


src/live.rs
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::cmdline::*;
+use crate::errors::*;
+use crate::io::*;
```
djzager, line 17: This was done by the language server on format. Let me know if I should undo this.

djzager:
This is my attempt to address the feature request submitted in #382 and I don't want to pretend like I know what I'm doing :sunglasses: . Comments welcome.

src/live.rs
bgilbert: Yes, please.

src/live.rs
```diff
@@ -82,7 +85,13 @@ pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
     // write new config
     embed.seek_to_start()?;
     embed.write(&cpio)?;
-    holder.complete();
+    if config.output.as_deref() == Some("-") {
+        stdout().write_all(&cpio).chain_err(|| "writing output")?;
+        stdout().flush().chain_err(|| "flushing output")?;
+        drop(holder)
```
bgilbert, line 91: This doesn't save any disk I/O, though.  We're still copying the image to a temporary file (using reflinks to avoid I/O, if reflinks are available) and then deleting it afterward.

src/live.rs
```diff
@@ -43,6 +42,10 @@ pub fn iso_remove(config: &IsoIgnitionRemoveConfig) -> Result<()> {
 }
 
 pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
+    if config.output.as_deref() == Some("-")
```
djzager, line 45: I suspect this shouldn't be magic but wasn't certain the correct way to define it...maybe, `const USE_STDOUT: str = "-"`?

src/live.rs
djzager: There are now 2 commits on this PR. The first uses the easiest method (creating the temporary file and deleting afterward), the second skips the holder.

src/live.rs
```diff
@@ -63,34 +63,37 @@ pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
...
-    }
-    // delete any existing config
-    embed.clear()?;
-    // write new config
-    embed.seek_to_start()?;
-    embed.write(&cpio)?;
-    if config.output.as_deref() == Some("-") {
+    if config.output.as_deref() == USE_STDOUT {
+        let cpio = make_cpio(&ignition)?;
         stdout().write_all(&cpio).chain_err(|| "writing output")?;
```
bgilbert, line 68: This isn't embedding the CPIO in the ISO image, it's just generating the CPIO and sending it directly to stdout.  `coreos-installer pxe ignition wrap` already does that, if desired.

djzager: whoosh. it didn't feel right when I was doing it but I went for it anyway.

djzager:
Making it so it compiles. Now that I recognize where this falls short. I'll try to take some time over break to fix where I went wrong.

djzager:
Updated so that it now works as expected, `coreos-installer iso ignition embed -o -` dumps the entire iso to stdout. I verified my work with:

**ignition file**

```
{
  "ignition": {
    "version": "3.1.0"
  },
  "passwd": {
    "users": [
      {
        "groups": [
          "sudo"
        ],
        "name": "core",
        "sshAuthorizedKeys": [
          "replace-with-your-ssh-public-key"
        ]
      }
    ]
  },
  "storage": {
    "files": [
      {
        "path": "/etc/assisted-service/create_certs.sh",
        "contents": {
          "compression": "gzip",
          "source": "data:;base64,H4sIAAAAAAAC/5yPsU7DMBCG9zzFEeaLK0grKMoQhYUlCJUXcOxrY+LYwXeF5O1RUdW1Urd/+b5P//2d6lxQneY+GwfrEuAEisQozexYyCJT+nGGVDi4MKOhJJzFiQKzh0TfgPN69QwYoiUGtHpheNysAfnYfUGumqqp1e6z+mjUe1WfpfAWWLT3lFTTVpeUnlzho9G++A/0bi+FizmgtpZmgfzkJCO1l1aPVL22u+11ONDvQAsk1tuHVfkEONASj3L95nkz+T27QyBbnDx4I2uSvGSmH6OFTVneWM/+AgAA//9PXnJesQEAAA=="
        },
        "mode": 420
      },
      {
        "path": "/etc/assisted-service/environment",
        "contents": {
          "compression": "gzip",
          "source": "data:;base64,H4sIAAAAAAAC/1SOXY+iMBiF7/kVhnuKHyxrSHpRoaskLO221c1cNY1UbYKFoZ2JE+N/n5A4jnP5vOfkeQ8lXKwZ5v8qWSCBVohjaKzzqm31EDylFHH+n7ACquZs7HOy5Zjdr8VKbggXcDb/DaZgCmbjhRIm4K9kMR/hZ3mUflON/j4/55jtyhzLcZPcsgqevO9dFsfKOeO8biLVG9B2e9WCrtfWnczBA9NlyyRZBAWmFXmRArE1FrCz/aDPD2WOZI6ZkBSJDYwb5VVsj8ZeIqfbgzNHqxuwH3xAKK75pvwj5A4zXpKaw2uYgDTMJtdw0K1WTktzVkcdZpPw9U19ANPFjzHRvRI1+j3u9v0XZwlIQRJdlqlMk/B2C+h2VZW5zEktUFljJhlel1ywEnN4twafAQAA//8OxvIhqgEAAA=="
        },
        "mode": 420
      },
      {
        "path": "/etc/assisted-service/nginx.conf",
        "contents": {
          "compression": "gzip",
          "source": "data:;base64,H4sIAAAAAAAC/3yQsW7zMAyEdz8FBwOZfivBnyFNxi59gc6CKtO2EEUSSCawUfjdC9keCiGNBwNHfScdj5EeSPBdAQB4x4IBTsfjf2D2l2XI7LVFEtc5awRBZcEq9C6M/xh9x64P2DaW5KlBX3H603TFaTMtOXQwNwS9jihGAWVSWqULLY7rvxnktqXz0RpxMWTQbWvkL1EcJ50MMwwi6axUJv0QWc6n/dv+UpAZ0g8kzncdmkN5zih6QNMiwWfqybQI9eK5r+oF/x5DQLuE3G307gX+EVmgzkFLyBo7oP6alq2evD4XjfyqQ2jSnfPIUN/JgSprnKu5+gkAAP//yEhoiwsCAAA="
        },
        "mode": 420
      },
      {
        "path": "/etc/assisted-service/auth.json",
        "contents": {
          "source": "data:,replace-with-your-urlencoded-pull-secret"
        },
        "mode": 420
      },
      {
        "path": "/etc/assisted-service/startup_script.sh",
        "contents": {
          "compression": "gzip",
          "source": "data:;base64,H4sIAAAAAAAC/0SOwUorQRBF9/0V9/VLiIKT/oGZoEgWsxMDbowknZmaTEFS3XQ1QVD/XXoUU5u7KDjn/P/nDizu4HU0HLWZ3YxBs/gzoWpvTSLfo0qoPDg+pIS6rmFnHNWaISRwBAvs7GP6vt6/fVkDAH3AtOViYskD7FbmCq/KmqmvfOTlKXT+tAyRREce8pLDVuyEt1it4Ch3ruToL1OoNO46bfbUjQGlA5949+lYVqlHRViog7tzx8XeXNWb9fNL+7jetU+bZq4/kkK6iv7KlNKFO3IkF05BziTZfAcAAP//9XIuQCcBAAA="
        },
        "mode": 420
      }
    ]
  },
  "systemd": {
    "units": [
      {
        "contents": "[Unit]\nAfter=network-online.target\n\n[Service]\nType=oneshot\nRestart=no\nExecStart=/bin/bash /etc/assisted-service/create_certs.sh\n\n[Install]\nWantedBy=multi-user.target\n",
        "enabled": true,
        "name": "assisted-service-certs.service"
      },
      {
        "contents": "[Unit]\nAfter=network-online.target\n\n[Service]\nType=forking\nRestart=no\nExecStart=/bin/bash /etc/assisted-service/startup_script.sh\n\n[Install]\nWantedBy=multi-user.target\n",
        "enabled": true,
        "name": "assisted-service-ip-configuration.service"
      },
      {
        "contents": "[Unit]\nAfter=network-online.target\n\n[Service]\nType=oneshot\nRestart=no\nExecStart=/bin/bash -c 'cp /dev/disk/by-label/rhcos* /etc/assisted-service/livecd.iso'\n\n[Install]\nWantedBy=multi-user.target\n",
        "enabled": true,
        "name": "assisted-service-livecd.service"
      },
      {
        "contents": "[Unit]\nAfter=assisted-service-ip-configuration.service\n\n[Service]\nType=forking\nRestart=no\nExecStart=podman pod create --name assisted-installer -p 8443\n\n[Install]\nWantedBy=multi-user.target\n",
        "enabled": true,
        "name": "assisted-service-pod.service"
      },
      {
        "contents": "[Unit]\nAfter=assisted-service-pod.service\n\n[Service]\nType=forking\nRestart=no\nExecStart=podman run -dt --pod assisted-installer --env-file /etc/assisted-service/environment --authfile /etc/assisted-service/auth.json --name db quay.io/ocpmetal/postgresql-12-centos7\nTimeoutStartSec=300\n\n[Install]\nWantedBy=multi-user.target\n",
        "enabled": true,
        "name": "assisted-service-db.service"
      },
      {
        "contents": "[Unit]\nAfter=assisted-service-db.service\n\n[Service]\nType=forking\nRestart=no\nExecStart=podman run -dt --pod assisted-installer --env-file /etc/assisted-service/environment -v /bin/coreos-installer:/data/coreos-installer:z -v /etc/assisted-service/livecd.iso:/data/livecd.iso:z -v /etc/assisted-service/nginx-certs/nginx-selfsigned.crt:/data/nginx-selfsigned.crt:z --restart always --name installer quay.io/ocpmetal/assisted-service:latest\nTimeoutStartSec=300\n\n[Install]\nWantedBy=multi-user.target\n",
        "enabled": true,
        "name": "assisted-service-installer.service"
      },
      {
        "contents": "[Unit]\nAfter=assisted-service-installer.service\n\n[Service]\nType=forking\nRestart=no\nExecStart=podman run -dt --pod assisted-installer --env-file /etc/assisted-service/environment -v /etc/assisted-service/nginx-certs:/certs:z -v /etc/assisted-service/nginx.conf:/opt/bitnami/nginx/conf/server_blocks/nginx.conf:z --name ui quay.io/ocpmetal/ocp-metal-ui:latest\nTimeoutStartSec=300\n\n[Install]\nWantedBy=multi-user.target\n",
        "enabled": true,
        "name": "assisted-service-ui.service"
      }
    ]
  }
}
```

**Using latest coreos-installer**

```
$ /path/to/latest/coreos-installer iso ignition embed -f -i onprem-iso-config.ign -o livecd2.iso rhcos-46.82.202012051820-0.iso
$ sha256sum livecd2.iso3071772315868c28ce875991b5c46e5b520ab1b23ce536e182ebb6fde7d366d9  livecd3.iso
3071772315868c28ce875991b5c46e5b520ab1b23ce536e182ebb6fde7d366d9  livecd2.iso
```

**Using this PR, to file**

```
$ target/debug/coreos-installer iso ignition embed -f -i onprem-iso-config.ign -o livecd3.iso rhcos-46.82.202012051820-0.iso
$ sha256sum livecd3.iso
3071772315868c28ce875991b5c46e5b520ab1b23ce536e182ebb6fde7d366d9  livecd3.iso
```

**Using this PR, to stdout redirected to file**

```
$ target/debug/coreos-installer iso ignition embed -f -i onprem-iso-config.ign -o - rhcos-46.82.202012051820-0.iso > livecd4.iso
$ sha256sum livecd4.iso
3071772315868c28ce875991b5c46e5b520ab1b23ce536e182ebb6fde7d366d9  livecd4.iso
```

src/live.rs
djzager: Looking into eliminating the need for the temp file I got the feeling I would need to, at a minimum, add new methods to EmbedArea. Do you have any thoughts/guidance on pursuing this route?

djzager:
@bgilbert I updated the PR to not need a temporary file. I'm not particularly happy with the `write_to_stdout` but I'm struggling to reason through the responsibilities of EmbedArea, as I see it:
1. Get the embed area offset + length
1. Seek to the beginning of the embed area
1. Clear the embed area
1. Write the embed area given some buffer

My problem is that it doesn't _feel_ as though writing the whole file + ignition is the responsibility of EmbedArea but it's tough to do it in the `iso_ignition_embed` given the file is already borrowed and I **need** EmbedArea to tell me the offset and length of the embed area.

Feedback would be greatly appreciated. Thanks.

src/live.rs
```diff
@@ -59,30 +63,52 @@ pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
...
-    }
-    if !config.force {
-        // Ensure all zero bytes
-        embed.seek_to_start()?;
-        let mut buf = embed.new_buffer();
-        embed.read(&mut buf)?;
-        // compare to zeroed buffer
-        if buf != embed.new_buffer() {
-            bail!("This ISO image already has an embedded Ignition config; use -f to force.");
+    if config.output.as_deref() == USE_STDOUT {
```
bgilbert, line 66: Let's break this into two separate `if` blocks, one to open files and one to do the writing, so we can avoid duplicating the `make_cpio()` call and the validity checks.  That'll require some fancy footwork in variable assignments, but should be doable.

src/live.rs
```diff
@@ -59,30 +63,52 @@ pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
...
+
+        let cpio = make_cpio(&ignition)?;
+        if cpio.len() > embed.length {
+            bail!(
+                "Compressed Ignition config is too large: {} > {}",
+                cpio.len(),
+                embed.length
+            );
         }
+        if !config.force {
```
bgilbert, line 95: Let's do this check in the stdout path also.  It's a bit odd, but it makes the semantics more consistent.  Note that we currently do the check with `-o`, which is in a similar situation.

src/live.rs
```diff
@@ -43,8 +44,11 @@ pub fn iso_remove(config: &IsoIgnitionRemoveConfig) -> Result<()> {
 }
 
 pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
-    let mut holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
-    let mut embed = EmbedArea::for_file(&mut holder.file)?;
+    if config.output.as_deref() == USE_STDOUT
```
bgilbert, line 47: Maybe do this test once, as `let use_stdout = config.output.as_deref() == Some("-");`?

bgilbert (commented):
This is great!  Some ideas for minor refactoring, but the idea and implementation look generally sound.

I think you could add an `EmbedArea` method to lend the `file` back to you, but honestly the current approach seems okay.  `EmbedArea` is mostly a parser bolted to a set of convenience functions, and doing a "streaming embed" doesn't seem too far outside its responsibilities.

Would you be willing to add similar logic to `iso embed remove`?  I recognize that that feature might never be used, but it's odd for `embed` to support `-` and `remove` not to support it.

For the record, after you opened this PR we added the `iso kargs` family of subcommands, which are not being handled here.  I'm not going to ask you to tackle those, though.  :slightly_smiling_face:

src/live.rs
```diff
@@ -285,6 +311,22 @@ impl<'a> EmbedArea<'a> {
...
+            .chain_err(|| "seeking to start")?;
+        copy(&mut self.file.take(self.offset), &mut stdout())
+            .chain_err(|| "copying file to stdout")?;
+        stdout().write_all(buf).chain_err(|| "writing embed area")?;
+        let zeroes = vec![0; self.length - buf.len()];
+        stdout().write_all(&zeroes).chain_err(|| "writing zeros")?;
+        self.file
+            .seek(SeekFrom::Start(self.offset + self.length as u64))
+            .chain_err(|| "seeking to end of embed area")?;
+        copy(&mut self.file, &mut stdout()).chain_err(|| "copying file to stdout")?;
```
bgilbert, line 326: I think, if you check `strace`, these copies will end up using a small I/O size.  It might make sense to wrap the input and output in `BufReader`/`BufWriter` with reasonably large buffers, to reduce syscall overhead.

djzager:
@bgilbert thanks for the review! I'll rebase tonight/tomorrow and go through the comments and see what I can do to address them all.

djzager:
Testing with latest:

```
# Based on this PR ignition embed to file, to stdout via redirection, and latest coreos-installer to file
$ target/debug/coreos-installer iso ignition embed -f -i onprem-iso-config.ign -o livecd1.iso livecd.iso
$ target/debug/coreos-installer iso ignition embed -f -i onprem-iso-config.ign -o - livecd.iso > livecd2.iso
$ ./coreos-installer iso ignition embed -f -i onprem-iso-config.ign -o livecd3.iso livecd.iso

# Based on this PR ignition remove to file, to stdout via redirection, and latest coreos-installer to file
$ target/debug/coreos-installer iso ignition remove -o livecd4.iso livecd1.iso
$ target/debug/coreos-installer iso ignition remove -o - livecd1.iso > livecd5.iso
$ ./coreos-installer iso ignition remove -o livecd6.iso livecd.iso
```

Verify:
```
sha256sum livecd.iso livecd1.iso livecd2.iso livecd3.iso livecd4.iso livecd5.iso livecd6.iso
38d8e14491758b8b7052cb8e82704fe25b042875a3fbaa5298bd4f1fb1fc416e  livecd.iso
ded05da7c8bdfce334772a9a5cf28e57d73b572cf79c24df1e4db15bc9f27f7e  livecd1.iso
ded05da7c8bdfce334772a9a5cf28e57d73b572cf79c24df1e4db15bc9f27f7e  livecd2.iso
ded05da7c8bdfce334772a9a5cf28e57d73b572cf79c24df1e4db15bc9f27f7e  livecd3.iso
38d8e14491758b8b7052cb8e82704fe25b042875a3fbaa5298bd4f1fb1fc416e  livecd4.iso
38d8e14491758b8b7052cb8e82704fe25b042875a3fbaa5298bd4f1fb1fc416e  livecd5.iso
38d8e14491758b8b7052cb8e82704fe25b042875a3fbaa5298bd4f1fb1fc416e  livecd6.iso
```

djzager:
> For the record, after you opened this PR we added the `iso kargs` family of subcommands, which are not being handled here. I'm not going to ask you to tackle those, though.

Tempted to create an issue (or is there one already?) for this and see if I can tackle it when I get some time.

src/live.rs
```diff
@@ -116,10 +141,34 @@ pub fn iso_ignition_show(config: &IsoIgnitionShowConfig) -> Result<()> {
...
+                .open(&config.input)
+                .chain_err(|| format!("opening {}", &config.input))?,
+            copied_path: None,
+            complete: false,
+        };
+    } else {
+        holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
+    }
+    embed = EmbedArea::for_file(&mut holder.file)?;
+
```
bgilbert, line 165: Since we don't have a meaningful set of common code between the two cases here, it might be cleaner to just have a single if/else block.

src/live.rs
```diff
@@ -493,6 +542,30 @@ impl<'a> EmbedArea<'a> {
...
+        }
+        writer.write_all(buf).chain_err(|| "writing embed area")?;
+        let zeroes = vec![0; self.length - buf.len()];
+        writer.write_all(&zeroes).chain_err(|| "writing zeros")?;
+        self.file
+            .seek(SeekFrom::Start(self.offset + self.length as u64))
+            .chain_err(|| "seeking to end of embed area")?;
+        copy(
+            &mut BufReader::with_capacity(BUFFER_SIZE, self.file),
+            &mut BufWriter::with_capacity(BUFFER_SIZE, writer),
```
bgilbert, line 563: The `BufWriter` [docs](https://doc.rust-lang.org/std/io/struct.BufWriter.html) recommend an explicit `flush()`.

src/live.rs
```diff
@@ -67,6 +69,23 @@ pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
...
+                .read(true)
+                .open(&config.input)
+                .chain_err(|| format!("opening {}", &config.input))?,
+            copied_path: None,
+            complete: false,
+        };
+    } else {
+        holder = CopiedFileHolder::new(&config.input, config.output.as_ref())?;
+    }
+    embed = EmbedArea::for_file(&mut holder.file)?;
```
bgilbert, line 87: This isn't really using `CopiedFileHolder` correctly; `file` is supposed to be the output file, not the input.  It happens to work with the code paths we have now, but it's confusing.

I spent some time trying to refactor this more cleanly (as you may also have done) and wasn't able to find a good solution within the existing architecture.  I ended up refactoring `CopiedFileHolder` into an enum in #478, which could have a `Stdout` variant added.

What do you think about merging this part as-is, and then I'll rebase #478 on top and refactor this part?

bgilbert (commented):
This generally looks in good shape!  Should need only a few small fixes.

> Tempted to create an issue (or is there one already?) for this and see if I can tackle it when I get some time.

Sounds good!

src/live.rs
djzager: > This isn't really using `CopiedFileHolder` correctly; `file` is supposed to be the output file, not the input. It happens to work with the code paths we have now, but it's confusing.

I agree :sunglasses: 

> I spent some time trying to refactor this more cleanly (as you may also have done) and wasn't able to find a good solution within the existing architecture. I ended up refactoring `CopiedFileHolder` into an enum in #478, which could have a `Stdout` variant added.

Played a little and it was outside what I was comfortable doing as a simple contributor.

> What do you think about merging this part as-is, and then I'll rebase #478 on top and refactor this part?

:+1:

bgilbert (approved):
Looks great.  Thanks for doing this!

PR #453, opened 2021-01-15
blockdev: use --nodeps when querying single device

jlebon:
`lsblk` by default outputs the target device followed by all its reverse
dependencies. But interestingly, when using `--pairs`, the `lsblk` in
RHCOS seems to output the reverse dependency first before the target
device.

We were relying on correct ordering in `lsblk_single`, where all we care
about is the target device. Instead, explicitly use `--nodeps` in that
case so that `lsblk` knows to only print the target device. (The
`lsblk()` helper actually used to support that, but it was removed in
88de4fc.)

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1916502

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -94,7 +94,7 @@ impl Disk {
     fn get_partitions(&self, with_holders: bool) -> Result<Vec<Partition>> {
         // walk each device in the output
         let mut result: Vec<Partition> = Vec::new();
-        for devinfo in lsblk(Path::new(&self.path))? {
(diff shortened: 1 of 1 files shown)


bgilbert (approved):
@arithx confirmed that this fixes 1916502.

jlebon:
> But interestingly, when using --pairs, the lsblk in
RHCOS seems to output the reverse dependency first before the target
device.

One clarification on this: it seems like in *some cases*, RHEL's `lsblk` does output things in the right order, which makes it all the more confusing. I think it may be dependent on the type of block devices involved.

PR #478, opened 2021-03-04
iso: refactor CopiedFileHolder as an enum and rename to ContentFile

bgilbert:
Explicitly separate the in-place and copied behaviors.  Use `NamedTempFile` for the copied case, rather than writing directly to the final location and deleting the output file on failure.  Add explicit support for streaming the output to stdout.

This will set the copied output file mode to `0600`, which seems okay because the embedded Ignition config may contain secrets.  It also ensures that the permanent filename is not used until the file is completely written.

cc @djzager @jlebon

Requires #432.

### DIFF
--- a/src/live.rs
+++ b/src/live.rs
@@ -69,11 +69,9 @@ pub fn iso_ignition_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
         }
     };
 
-    let mut holder: CopiedFileHolder;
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Looks great!

bgilbert:
Rebased on #432 and ready for review.

djzager (approved):
Nice change.

jlebon (approved):

PR #480, opened 2021-03-05
Switch from error-chain to anyhow

bgilbert:
error-chain is no longer actively maintained.  Switch to a more modern library.  This does lose us backtrace support until `std::backtrace` is stabilized.

Switch to returning a `Result` directly from `main()`, since anyhow has reasonable default error rendering.


bgilbert:
Ready for review.

src/blockdev.rs
```diff
@@ -322,18 +321,23 @@ impl Partition {
         // We multiply by 512 here: the kernel values are always in 512 blocks, regardless of the
         // actual sector size of the block device. We keep the values as bytes to make things
         // easier.
-        let start_offset: u64 = start.checked_mul(512).ok_or("start offset mult overflow")?;
+        let start_offset: u64 = start
+            .checked_mul(512)
+            .ok_or_else(|| anyhow!("start offset mult overflow"))?;
```
jlebon, line 326: Hmm, interesting. Does `ok_or("string")` not work anymore? I would think the implied `into()` would convert it fine.

jlebon (approved):
I only scanned this, but it looks sane to me!

src/blockdev.rs
bgilbert: It doesn't.  `anyhow::Error` doesn't implement `From<&str>`.

jlebon: Ack.

PR #481, opened 2021-03-06
iso: stream iso kargs {modify|reset} to stdout

### LINKED ISSUE #479: stream iso kargs modify|reset to stdout
# Feature Request #

Much like #382 (implemented in #432) it should be possible to stream the result of `coreos-installer iso kargs {modify|reset}` to stdout.

Would look like `coreos-installer iso kargs {modify|reset} -o -`

djzager:
When given the `-o -` option, coreos-installer will stream the resulting ISO to stdout.

Fixes #479

Requires #480

### DIFF
--- a/src/live.rs
+++ b/src/live.rs
@@ -178,10 +178,9 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
 
 pub fn iso_kargs_modify(config: &IsoKargsModifyConfig) -> Result<()> {
     let mut content = ContentFile::new(&config.input, config.output.as_ref())?;
-    if content.is_stdout() {
-        bail!("Writing to stdout is not supported");
-    }
+    let use_stdout = content.is_stdout();
     let mut embed = KargEmbedAreas::for_file(content.as_file_mut())?;
+
     let current_kargs = embed.get_current_kargs()?;
     let new_kargs = modify_kargs(
         &current_kargs,
@@ -189,19 +188,26 @@ pub fn iso_kargs_modify(config: &IsoKargsModifyConfig) -> Result<()> {
         &config.replace,
         &config.delete,
     )?;
-    embed.write_kargs(&new_kargs)?;
+    if use_stdout {
+        embed.stream(&new_kargs, &mut stdout())?;
+    } else {
+        embed.write_kargs(&new_kargs)?;
+    }
     content.complete()?;
     Ok(())
 }
 
 pub fn iso_kargs_reset(config: &IsoKargsResetConfig) -> Result<()> {
     let mut content = ContentFile::new(&config.input, config.output.as_ref())?;
-    if content.is_stdout() {
-        bail!("Writing to stdout is not supported");
-    }
+    let use_stdout = content.is_stdout();
     let mut embed = KargEmbedAreas::for_file(content.as_file_mut())?;
+
     let default_kargs = embed.get_default_kargs()?;
-    embed.write_kargs(&default_kargs)?;
+    if use_stdout {
+        embed.stream(&default_kargs, &mut stdout())?;
+    } else {
+        embed.write_kargs(&default_kargs)?;
+    }
     content.complete()?;
     Ok(())
 }
@@ -304,6 +310,9 @@ impl<'a> KargEmbedAreas<'a> {
             kargs_offsets.push(offset);
         }
 
+        // we need ordered offsets when streaming
+        kargs_offsets.sort_unstable();
+
         // we expect at least one
         if kargs_offsets.is_empty() {
             bail!("No karg embed areas found; corrupted CoreOS ISO image.");
@@ -343,7 +352,7 @@ impl<'a> KargEmbedAreas<'a> {
         get_kargs_at_offset(self.file, self.length, self.default_kargs_offset)
     }
 
-    fn write_kargs(&mut self, kargs: &str) -> Result<()> {
+    fn format_embed_area(&mut self, kargs: &str) -> Result<Vec<u8>> {
         let kargs: String = kargs.trim().to_string() + "\n";
         if kargs.len() > self.length {
             bail!(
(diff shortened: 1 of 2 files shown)


tests/test-iso-kargs.sh
```diff
@@ -25,7 +25,7 @@ fi
 # Test all the modification operations.
 coreos-installer iso kargs modify -a foobar=oldval -a dodo "${iso}"
 coreos-installer iso kargs show "${iso}" | grep -q 'foobar=oldval dodo'
-coreos-installer iso kargs modify -r foobar=oldval=newval "${iso}"
+coreos-installer iso kargs modify -r foobar=oldval=newval -o - "${iso}" > "${iso}"
```
bgilbert, line 28: The shell will truncate the output file before we can ever read it as input.

src/live.rs
```diff
@@ -343,6 +349,48 @@ impl<'a> KargEmbedAreas<'a> {
...
+                self.length
+            );
+        }
+
+        self.file
+            .seek(SeekFrom::Start(0))
+            .context("seeking to start")?;
+        let mut new_area = vec![b'#'; self.length];
+        new_area[..kargs.len()].copy_from_slice(kargs.as_bytes());
+        self.kargs_offsets.sort();
```
bgilbert, line 368: Let's do this in `for_file()`, rather than mutating the vector here.

src/live.rs
```diff
@@ -343,6 +349,48 @@ impl<'a> KargEmbedAreas<'a> {
...
+                kargs.len(),
+                self.length
+            );
+        }
+
+        self.file
+            .seek(SeekFrom::Start(0))
+            .context("seeking to start")?;
+        let mut new_area = vec![b'#'; self.length];
+        new_area[..kargs.len()].copy_from_slice(kargs.as_bytes());
```
bgilbert, line 367: There's enough duplicated code setting up `new_area` that it should probably just move to a helper method.

src/live.rs
```diff
@@ -504,6 +552,7 @@ impl<'a> EmbedArea<'a> {
         {
             bail!("Invalid CoreOS ISO image.");
         }
+        eprintln!("offset: {}, length: {}", offset, length);
```
bgilbert, line 555: Stray debug print.

bgilbert (commented):
General structure looks good; just some nits.

bgilbert:
#480 landed.

src/live.rs
djzager: I guessed a name b/c `let new_area = self.new_area(...)` looked silly. Let me know what you think.

src/live.rs
```diff
@@ -343,6 +350,52 @@ impl<'a> KargEmbedAreas<'a> {
...
+                "kargs too large for area: {} vs {}",
+                kargs.len(),
+                self.length
+            );
+        }
+
+        self.file
+            .seek(SeekFrom::Start(0))
+            .context("seeking to start")?;
+        let new_area = self.format_kargs(&kargs).unwrap();
```
bgilbert, line 373: In this codebase, the hierarchy of error handling is:

1. `unwrap()` - obvious programmer errors, like extracting a `Some()` value when we just checked that we're not in the `None` case
2. `expect("explanation")` - less obvious programmer errors, like an inconsistent state caused by another method, where an `unwrap()` won't be sufficiently clear for debugging
3. `?` - legitimate runtime errors that should be propagated to the caller, but we don't need to add additional context to the error produced by the callee (e.g. the callee is a helper function)
4. `context("explanation")?` or `with_context(|| format!("explanation for {}", something))?` - legitimate runtime errors where we know something the callee doesn't

I think this is case 3.

src/live.rs
bgilbert: Hmm, it's a bit generic.  Something like `format_embed_area`?

tests/test-iso-kargs.sh
bgilbert: We should still test the functionality, though.

tests/test-iso-kargs.sh
```diff
@@ -25,6 +26,14 @@ fi
 # Test all the modification operations.
 coreos-installer iso kargs modify -a foobar=oldval -a dodo "${iso}"
 coreos-installer iso kargs show "${iso}" | grep -q 'foobar=oldval dodo'
+# Stream modification to stdout.
+coreos-installer iso kargs modify -a foobar=oldval -a dodo -o - "${iso}" > "${stdout_iso}"
+coreos-installer iso kargs show "${stdout_iso}" | grep -q 'foobar=oldval dodo'
```
bgilbert, line 31: This check will pass in any case, since the previous stanza already added these arguments to `${iso}`.  Since these checks don't depend on the surrounding modifications of `${iso}`, it'd make sense to move them.

tests/test-iso-kargs.sh
```diff
@@ -25,6 +26,14 @@ fi
...
+hash=$(sha256sum "${iso}")
+stdout_hash=$(sha256sum "${stdout_iso}")
+if [ "${stdout_hash}" != "${hash}" ]; then
```
bgilbert, line 34: And this check isn't necessary if `kargs modify` is run on a clean image.

tests/test-iso-kargs.sh
```diff
@@ -25,6 +26,14 @@ fi
 # Test all the modification operations.
 coreos-installer iso kargs modify -a foobar=oldval -a dodo "${iso}"
 coreos-installer iso kargs show "${iso}" | grep -q 'foobar=oldval dodo'
+# Stream modification to stdout.
+coreos-installer iso kargs modify -a foobar=oldval -a dodo -o - "${iso}" > "${stdout_iso}"
```
jlebon, line 30: This is modifying the same ISO that was just modified, so the kargs are already present.

What you could do instead is hash the stdout version before the one that modifies in-place. E.g.:

```
# Test all the modification operations
stdout_hash=$(coreos-installer iso kargs modify -a foobar=oldval -a dodo "${iso}" -o - | sha256sum)
coreos-installer iso kargs modify -a foobar=oldval -a dodo "${iso}"
coreos-installer iso kargs show "${iso}" | grep -q 'foobar=oldval dodo'

hash=$(sha256sum < ${hash})
if [ "${stdout_hash}" != "${hash}" ]; then
    fatal "Streamed hash doesn't match modified hash: ${stdout_hash} vs ${hash}"
fi
```

tests/test-iso-kargs.sh
djzager: :man_facepalming: thanks

tests/test-iso-kargs.sh
bgilbert: The current tests are reasonable as is, but note that @jlebon's code above verifies something that the tests currently don't: that `modify` in stdout mode makes the same changes that regular `modify` makes.  All we're confirming right now is that `modify` and `remove` are symmetric, and that `show` is happy with the result.

bgilbert (commented):
The linter [is unhappy](https://github.com/coreos/coreos-installer/pull/481/checks?check_run_id=2170331945).

bgilbert (approved):
Thanks!

PR #485, opened 2021-03-08
s390x: fix build

nikita-dubrovskii:
During migration from `error-chain` to `anyhow` s390x-code wasn't fully updated. Fix this


jlebon (approved):

jlebon:
(For the commit message, maybe something like: `s390x: convert missed error-chain error to anyhow` ? But LGTM as is too!)

nikita-dubrovskii:
> (For the commit message, maybe something like: `s390x: convert missed error-chain error to anyhow` ? But LGTM as is too!)

yep, done

PR #490, opened 2021-03-13
cargo: refresh lockfile

bgilbert:
gptman and mbrman both changed their error libraries.

@nikita-dubrovskii, could you check this on s390x?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "adler"
-version = "0.2.2"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ccc9a9dd069569f212bc4330af9f17c4afb5e8ce185e83dbb14f1349dda18b10"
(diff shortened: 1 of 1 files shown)


sohankunkerkar (commented):
This change LGTM
Would ask @nikita-dubrovskii to approve this change based on the outcome on s390x

bgilbert:
gptman updated in https://github.com/coreos/coreos-installer/pull/494 and mbrman updated in https://github.com/coreos/coreos-installer/pull/492.  Converting this to just a lockfile refresh.

bgilbert:
libc 0.2.90 breaks openat; https://github.com/tailhook/openat/issues/35.

bgilbert:
The metal4k test is consistently timing out in CI.  From the logs, it appears that c-i is still writing to disk when killed, and the same `testiso` command succeeds when run against a local build.  So I'm thinking that one of the dependency bumps is accentuating #184.

arithx (approved):

nikita-dubrovskii:
Checked with custom RHCOS-48.84, works fine.
LGTM

bgilbert:
This iteration now pulls in dependencies that are incompatible with our MSRV.  I'll just close this out for now.

PR #491, opened 2021-03-13
*: fix clippy::unnecessary_wraps lint on 1.50; use newer clippy in CI

bgilbert:
https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_wraps

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,8 +8,10 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  # Minimum and linting toolchain
-  ACTIONS_PINNED_TOOLCHAIN: 1.44.1
+  # Minimum supported Rust version (MSRV)
+  ACTIONS_MSRV_TOOLCHAIN: 1.44.1
+  # Pinned toolchain for linting
+  ACTIONS_LINTS_TOOLCHAIN: 1.50.0
 
 jobs:
   tests-stable:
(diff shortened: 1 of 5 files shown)


bgilbert:
I've added a commit to switch back to a dual-pinning CI model where we test against the minimum Rust version while running lints on a reasonably current one.  If CI can't be configured to detect lint that will show up on developer systems, that's a potential source of frustration.

The first commit contains a lint override that isn't supported by the MSRV clippy version, so we'll need to override CI to merge this.

Note to self: set the new check to Required after merging.

bgilbert:
Added another commit to cross-build and run clippy on s390x, so we get some minimum amount of coverage for the s390x-specific code.

cgwalters (approved):

cgwalters:
>  lzma-sys wants to build its own private copy of liblzma, so we'd need a cross C compiler.

Right, this is where the stock GH actions are limiting.  But, one approach we could take is to by default install e.g. `yum -y install gcc-c++-s390x-linux-gnu.x86_64` (and while we're at it might as well do `ppc64le` and `aarch64`) in the `cosa-buildroot` container.   And more generally, extend coreos-assembler to handle cross build workflows.

(The problem here is the more we do this, the more distant it becomes from the current Fedora workflow, which is all about [self hosting](https://blog.verbum.org/2012/10/13/building-everything-from-source-vs-self-hosting/))

PR #497, opened 2021-03-16
rdcore/kargs: add hidden --override-options switch

jlebon:
For testing the semantics of `rdcore kargs`, it's helpful to just be
able to directly pass the `options` string we want it to modify. Let's
add a switch to make that easier.

Best viewed with whitespace ignored.

### DIFF
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -27,6 +27,7 @@ pub enum Config {
 pub struct KargsConfig {
     pub boot_device: Option<String>,
     pub boot_mount: Option<String>,
+    pub override_options: Option<String>,
     pub append_kargs: Vec<String>,
     pub delete_kargs: Vec<String>,
 }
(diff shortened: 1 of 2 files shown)


src/bin/rdcore/cmdline.rs
```diff
@@ -143,12 +153,16 @@ pub fn parse_args() -> Result<Config> {
 
 fn parse_kargs(matches: &ArgMatches) -> Result<Config> {
     // we could enforce these via clap's ArgGroup, but I don't like how the --help text looks
-    if !(matches.is_present("boot-device") || matches.is_present("boot-mount")) {
+    if !(matches.is_present("boot-device")
+        || matches.is_present("boot-mount")
+        || matches.is_present("override-options"))
+    {
```
bgilbert, line 159: Maybe also say `// --override-options is undocumented` to prevent someone from updating the `bail!()`?

bgilbert (approved):

jlebon:
Ahh right, we need to fix CI here too: https://github.com/coreos/coreos-installer/pull/506

jlebon:
Just rebased for #506.

PR #498, opened 2021-03-16
rdcore/kargs: add --should-exist and --should-not-exist

jlebon:
See individual commit messages.

For https://github.com/coreos/ignition/issues/1168.

Requires: https://github.com/coreos/coreos-installer/pull/497

### DIFF
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -29,6 +29,7 @@ pub struct KargsConfig {
     pub boot_mount: Option<String>,
     pub override_options: Option<String>,
     pub append_kargs: Vec<String>,
+    pub append_kargs_if_missing: Vec<String>,
     pub delete_kargs: Vec<String>,
 }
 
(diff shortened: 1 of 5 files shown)


arithx:
Hmm, WDYT about adding an option or default behavior that optionally would either output to a file or to stdout if any kargs were added/removed? That way we could avoid reboots if kargs were requested but no changes were actually needed?

bgilbert:
Are we trying to avoid adding an ignition-kargs wrapper script around this subcommand?  Could a script capture kargs before and after, and do a comparison?

LGTM otherwise.

arithx:
> Are we trying to avoid adding an ignition-kargs wrapper script around this subcommand? Could a script capture kargs before and after, and do a comparison?

We'd need one anyways because we need to mount `/boot` & pass in the `boot-mount` to `rdcore`. I was mostly suggesting it to avoid re-doing the check in bash when we already know that information in the rust code.

jlebon:
Rebased and added `--touch-if-changed` in a separate commit!

As noted in the commit message there, this does add a new dep. If we're strongly opposed, I can add a wrapper for `utimensat` and drop the dep.

jlebon:
OK, now changed to use `--create-if-changed` and dropped the dep! This should be good for a final review.

bgilbert (approved):

PR #499, opened 2021-03-17
blockdev: strengthen device mapper path detection

jlebon:
Apart from the `/dev/mapper` symlinks, there's a bunch of other possible
symlinks which also point to the same devmapper device. Instead of
trying to catch them all, simply resolve all symlinks and then check
that it starts with `dm-`.

See also: https://bugzilla.redhat.com/show_bug.cgi?id=1886229#c39

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -157,7 +157,7 @@ impl Disk {
         // Our investigation found nothing.  If the device is expected to be
         // partitionable but reread failed, we evidently missed something,
         // so error out for safety
-        if !self.is_dm_device() {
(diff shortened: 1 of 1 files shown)


jlebon:
(Not tested yet.)

jlebon:
OK, yup this fixes https://bugzilla.redhat.com/show_bug.cgi?id=1886229#c39! :heavy_check_mark:

jlebon:
We should probably add a `testiso` scenario which tests installing on a device mapper block device.

bgilbert (approved):

PR #513, opened 2021-04-03
Resolve symlink disk paths when creating `Disk` objects

### LINKED ISSUE #512: Confusing error when attempting to install on busy disks given by `/dev/disk/by-id/*` or similar paths
# Bug #
coreos-installer shows a confusing error message when attempting to install on busy disks that have partitions on them when those disks are given by `/dev/disk/by-id/*` or `/dev/disk/by-path/*`

## Host Operating System Version ##
Linux 5.9.11-200.fc33.x86_64 SMP Tue Nov 24 18:18:01 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

## Target Operating System Version ##
Irrelevant

## coreos-installer Version ##
coreos-installer 0.8.1-alpha.0

## Expected Behavior ##
`/dev/disk/by/id/*` are symlinks to `/dev/*` files, coreos-installer should say the disks are busy and show the busy partitions

## Actual Behavior ##
With example commandline: `sudo coreos-installer install /dev/disk/by-id/wwn-0x500a075115660ab6`

coreos-installer panics with the following error:

```
Error: checking for exclusive access to /dev/disk/by-id/wwn-0x5555555555555555

Caused by:
    couldn't find /sys/block directory for partition /dev/sda4 of /dev/disk/by-id/wwn-0x5555555555555555
```

This error is misleading

## Reproduction Steps ##

Note the `/dev/disk/by-id` path of a real, busy disk on your machine, e.g. `/dev/disk/by-id/wwn-0x5555555555555555`

  1. Run `sudo coreos-installer install /dev/disk/by-id/wwn-0x5555555555555555`

## Other Information ##

This is caused because `Partition` structs save their "parent" disk's path stored just as the user entered it. Later some function assumes the parent disk has a `/dev/*` format, and that causes problems because that assumption is false, the parent disk has a `/dev/disk/by-id/*` format instead.

This problematic function is `get_sysfs_dir`:
https://github.com/coreos/coreos-installer/blob/6734a1705ef0ada2e379644adeef8338c7e32168/src/blockdev.rs#L354

The **partition** path is typically something like `/dev/sda1`.

It tries to find the partition's sysfs path by basically constructing `/sys/block/$(basename ${parent_disk_path})/$(basename ${partition_path})`, this yields something that looks like `/sys/block/wwn-0x5555555555555555/sda1` - this path doesn't exist.

It then tries to check maybe the partition path (`/dev/sda1`) is a symlink as well, it's not - it then panics with a confusing error.

Note that this code path is not reached if the `BLKRRPART` ioctl [call](https://github.com/coreos/coreos-installer/blob/6734a1705ef0ada2e379644adeef8338c7e32168/src/blockdev.rs#L843) succeeds (indicating the disk is not busy) - in that case, installation goes on as usual.

omertuc:
Proposed solution for #512

Sometimes disks may be given by a path that looks something
like `/dev/disk/by-id/wwn-0x5555555555555555`. This path is
a symlink to the actual `/dev/*` device.

This commit prevents a confusing error message from showing
when the installer is given a busy disk that is a symlink to
a `/dev/*` device and not an actual `/dev/*` directly.

The solution used here is simply resolving the symlink before
creating a `Disk` object. This is done because code later down
the line assumes that the path is going to have a `/dev/*` format.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -46,10 +46,23 @@ pub struct Disk {
 }
 
 impl Disk {
-    pub fn new(path: &str) -> Self {
-        Disk {
-            path: path.to_string(),
-        }
+    pub fn new(path: &str) -> Result<Self> {
+        let canon_path = Path::new(path)
+            .canonicalize()
+            .with_context(|| format!("canonicalizing {}", path))?;
+
+        let canon_path = canon_path
+            .to_str()
+            .with_context(|| {
+                format!(
+                    "path {} canonicalized from {} is not UTF-8",
+                    canon_path.display(),
+                    path
+                )
+            })?
+            .to_string();
+
+        Ok(Disk { path: canon_path })
     }
 
     pub fn mount_partition_by_label(
--- a/src/install.rs
+++ b/src/install.rs
@@ -82,7 +82,7 @@ pub fn install(config: &InstallConfig) -> Result<()> {
     // get reference to partition table
     // For kpartx partitioning, this will conditionally call kpartx -d
     // when dropped
-    let mut table = Disk::new(&config.device)
+    let mut table = Disk::new(&config.device)?
         .get_partition_table()
         .with_context(|| format!("getting partition table for {}", &config.device))?;
 
@@ -121,7 +121,7 @@ pub fn install(config: &InstallConfig) -> Result<()> {
 }
 
 fn ensure_exclusive_access(device: &str) -> Result<()> {
-    let mut parts = Disk::new(device).get_busy_partitions()?;
+    let mut parts = Disk::new(device)?.get_busy_partitions()?;
     if parts.is_empty() {
         return Ok(());
     }
@@ -181,7 +181,7 @@ fn write_disk(
         || config.network_config.is_some()
         || cfg!(target_arch = "s390x")
     {
-        let mount = Disk::new(&config.device).mount_partition_by_label(
+        let mount = Disk::new(&config.device)?.mount_partition_by_label(
             "boot",
             false,
             mount::MsFlags::empty(),
--- a/src/osmet/mod.rs
+++ b/src/osmet/mod.rs
@@ -84,7 +84,7 @@ pub fn osmet_pack(config: &OsmetPackConfig) -> Result<()> {
     // MS_RDONLY; this also ensures that the partition isn't already mounted rw elsewhere.
     // Allow the root partition to be in a child holder device to allow for the RHCOS
     // crypto_LUKS partition.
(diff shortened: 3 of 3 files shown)


src/blockdev.rs
```diff
@@ -48,7 +48,11 @@ pub struct Disk {
 impl Disk {
     pub fn new(path: &str) -> Self {
         Disk {
-            path: path.to_string(),
+            path: if let Ok(target) = canonicalize(path) {
+                target.to_string_lossy().to_string()
+            } else {
+                path.to_string()
+            },
```
jlebon, line 55: I don't think we should gracefully handle `canonicalize` failing here. I.e. let's just change `new` to return a `Result`?

jlebon (commented):
Thanks for the patch! I agree this should fix https://github.com/coreos/coreos-installer/pull/513.

One comment, otherwise LGTM.

Re. the commit message, could you reference the issue using a `Closes: #512` footer instead? Also, for the commit title, can you prefix it with `blockdev:`?

src/blockdev.rs
omertuc: Sure, will get to that tomorrow :+1:

omertuc: Commit updated as requested

src/blockdev.rs
```diff
@@ -46,10 +46,10 @@ pub struct Disk {
 }
 
 impl Disk {
-    pub fn new(path: &str) -> Self {
-        Disk {
-            path: path.to_string(),
-        }
+    pub fn new(path: &str) -> Result<Self> {
+        Ok(Disk {
+            path: canonicalize(path)?.to_string_lossy().to_string(),
```
jlebon, line 51: Let's add a context here to make error messages more helpful? E.g.

```suggestion
            path: canonicalize(path)
                .with_context(|| format!("canonicalizing {}", path))?
                .to_string_lossy().to_string(),
```

Also, I know we use `to_string_lossy()` in a bunch of places, though here for this it'd be better IMO if we just hard error if it's not valid UTF-8 so we don't get weirder errors down the line on paths which build on top of it. We do this already in some places (see e.g. `is_dm_device()`).

jlebon (approved):
One final comment, otherwise LGTM!

src/blockdev.rs
omertuc: Will add context.

And will remove lossy. I initially added it because I didn't want to change the function signature / handle utf-8 errors, but now that we return Result there's no point in ignoring utf-8 errors.

PR #533, opened 2021-04-30
rdcore/kargs: call `zipl` on s390x after editing BLS

jlebon:
Otherwise the changed kargs won't take effect.

### DIFF
--- a/src/bin/rdcore/kargs.rs
+++ b/src/bin/rdcore/kargs.rs
@@ -15,6 +15,8 @@
 use anyhow::{Context, Result};
 
 use libcoreinst::install::*;
+#[cfg(target_arch = "s390x")]
+use libcoreinst::s390x;
 
 use crate::cmdline::*;
 use crate::rootmap::get_boot_mount_from_cmdline_args;
@@ -26,10 +28,15 @@ pub fn kargs(config: &KargsConfig) -> Result<()> {
         // the unwrap() here is safe because we checked in cmdline that one of them must be provided
         let mount =
             get_boot_mount_from_cmdline_args(&config.boot_mount, &config.boot_device)?.unwrap();
-        visit_bls_entry_options(mount.mountpoint(), |orig_options: &str| {
+        let _changed = visit_bls_entry_options(mount.mountpoint(), |orig_options: &str| {
             modify_and_print(config, orig_options)
         })
         .context("visiting BLS options")?;
+
+        #[cfg(target_arch = "s390x")]
+        if _changed {
+            s390x::run_zipl(mount.mountpoint())?;
+        }
     }
 
     Ok(())
--- a/src/bin/rdcore/kargs.rs
+++ b/src/bin/rdcore/kargs.rs
@@ -35,7 +35,7 @@ pub fn kargs(config: &KargsConfig) -> Result<()> {
 
         #[cfg(target_arch = "s390x")]
         if _changed {
-            s390x::run_zipl(mount.mountpoint())?;
+            s390x::zipl(mount.mountpoint())?;
         }
     }
 
--- a/src/s390x/mod.rs
+++ b/src/s390x/mod.rs
@@ -17,5 +17,6 @@ pub mod zipl;
 
 pub use dasd::{dasd_try_get_sector_size, image_copy_s390x, prepare_dasd};
 pub use zipl::install_bootloader;
+pub use zipl::run_zipl;
 mod eckd;
 mod fba;
--- a/src/s390x/mod.rs
+++ b/src/s390x/mod.rs
@@ -16,7 +16,7 @@ pub mod dasd;
 pub mod zipl;
 
 pub use dasd::{dasd_try_get_sector_size, image_copy_s390x, prepare_dasd};
-pub use zipl::install_bootloader;
-pub use zipl::run_zipl;
+pub use zipl::chreipl;
+pub use zipl::zipl;
 mod eckd;
 mod fba;
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -15,6 +15,7 @@
 use crate::install::{bls_entry_options_delete_and_append_kargs, visit_bls_entry_options};
 use crate::runcmd;
 use anyhow::{anyhow, Context, Result};
+use regex::Regex;
 use std::fs::{copy, create_dir_all, read_dir};
 use std::io::Write;
 use std::path::Path;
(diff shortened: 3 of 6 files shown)


src/bin/rdcore/kargs.rs
```diff
@@ -26,10 +28,14 @@ pub fn kargs(config: &KargsConfig) -> Result<()> {
...
         let mount =
             get_boot_mount_from_cmdline_args(&config.boot_mount, &config.boot_device)?.unwrap();
-        visit_bls_entry_options(mount.mountpoint(), |orig_options: &str| {
+        let changed = visit_bls_entry_options(mount.mountpoint(), |orig_options: &str| {
             modify_and_print(config, orig_options)
         })
         .context("visiting BLS options")?;
+
+        if changed && cfg!(target_arch = "s390x") {
+            runcmd!("zipl", "--target", mount.mountpoint())?;
```
jlebon, line 37: Not 100% sure if that's all we need. /cc @nikita-dubrovskii

Basing this off of `ignition-firstboot-complete`, which just calls `zipl`, except that here the boot partition isn't necessarily mounted on `/boot` so we need to point `zipl` at it: https://github.com/coreos/ignition/blob/3885b9f07c96baf7dcc12c85b914ded151c8fd6e/systemd/ignition-firstboot-complete.service#L26

bgilbert: Also, if a bare `zipl` works, I'm confused by all the explicit arguments in `install_bootloader()`.  If zipl indeed knows how to read BLS configs, could it do that in the `install_bootloader()` case too?

nikita-dubrovskii: I'll check it.

nikita-dubrovskii: So, `--target` option doesn't work without specifying `--image` and/or others. what could be used is `--blsdir` , but in such case `/boot` shoud be `rw` accessible.

jlebon: > what could be used is `--blsdir` , but in such case `/boot` shoud be `rw` accessible. 

Huh interesting, does it actually need write access to `/boot`, or just for the boot partition to be mounted at `/boot`? If the former, what does it write there?

jlebon:
OK, updated this now! One thing that became obvious is that we need a little more sophistication here because I want `rdcore kargs` to work equally well whether it's the first boot or not. This meant having the s390x code auto-detect it and doing the right thing.

@nikita-dubrovskii Would you be able to test this on s390x? Ideally using both `--firstboot-kargs` and an Ignition config which uses Ignition kargs.

nikita-dubrovskii:
Thanks @jschintag!

jlebon:
OK weird, I really thought `zipl` was already in the initramfs for FCOS at this point. @nikita-dubrovskii Can you submit that patch to the f-c-c repo?

nikita-dubrovskii:
> OK weird, I really thought `zipl` was already in the initramfs for FCOS at this point. @nikita-dubrovskii Can you submit that patch to the f-c-c repo?

Done - https://github.com/coreos/fedora-coreos-config/pull/1146

nikita-dubrovskii:
LGTM

jlebon:
Updated now to drop the tempdir move! (Good catch!) And also added @nikita-dubrovskii's `zipl` dracut patch here.

Because I'm very wary of code I can't test (I did compile-test it at least by fudging with the conditionals), I would appreciate still a final test of this PR using `--firstboot-kargs` and also using Ignition kargs.

nikita-dubrovskii:
> I would appreciate still a final test of this PR using `--firstboot-kargs` and also using Ignition kargs.

done, works well

jlebon:
Awesome, thanks! Will let @bgilbert have a final look before we merge.

src/s390x/zipl.rs
```diff
@@ -32,15 +32,11 @@ use tempfile::Builder;
 /// * `disk` - optional device path on which to run chreipl
 /// * `firstboot` - adds ignition.firstboot karg
 /// * `firstboot_kargs` - enables ignition firstboot and adds provided kargs
```
bgilbert, line 34: Stale comment.

src/s390x/zipl.rs
```diff
@@ -86,9 +79,12 @@ fn run_zipl<P: AsRef<Path>>(
                 copy(src.path(), blsdir.join(src.file_name())).unwrap();
             });
         let mut extra = vec!["ignition.firstboot".to_string()];
-        if let Some(kargs) = firstboot_kargs {
+        let firstboot_kargs = std::fs::read_to_string(firstboot_file)
+            .context("reading firstboot kargs from ignition.firstboot")?;
+        let firstboot_kargs = firstboot_kargs.trim();
```
bgilbert, line 84: The `ignition.firstboot` file [isn't that simple](https://github.com/coreos/coreos-installer/blob/d14fe5dbb5a864eee4d45070a6a7eb3943b54249/src/install.rs#L307), unfortunately.

src/s390x/mod.rs
```diff
@@ -17,5 +17,6 @@ pub mod zipl;
 
 pub use dasd::{dasd_try_get_sector_size, image_copy_s390x, prepare_dasd};
 pub use zipl::install_bootloader;
+pub use zipl::run_zipl;
```
bgilbert, line 20: The distinction between these two functions isn't immediately obvious from the names.  Can we rename one or both?

bgilbert (commented):
LGTM generally, and this is a nice cleanup, but I think `--firstboot-kargs` will put some extra garbage in the kernel command line.

src/s390x/zipl.rs
jlebon: Ahhh of course. There's no easy way to solve this generically, but at least we can special-case the format we expect. Fix added!

src/s390x/mod.rs
jlebon: Added another commit that cleans things up a bit more.

src/s390x/zipl.rs
```diff
@@ -95,9 +86,51 @@ pub fn install_bootloader<P: AsRef<Path>>(
...
+            extract_firstboot_kargs("set ignition_network_kcmdline=\"foobar\"").unwrap(),
+            Some("foobar".into())
+        );
+        assert_eq!(
+            extract_firstboot_kargs("set ignition_network_kcmdline=\"foo bar\"").unwrap(),
+            Some("foo bar".into())
+        );
+        assert!(extract_firstboot_kargs("set ignition_network_kcmdline=\"\n\"").is_err());
+        assert!(extract_firstboot_kargs("set ignition_network_kcmdline=\"foo\nbar\"").is_err());
+        assert!(extract_firstboot_kargs("foobar").is_err());
```
bgilbert, line 134: Other things to possibly test:
- Valid `set` line followed by a newline
- Valid `set` line preceded by something else
- Valid `set` line followed by something else

bgilbert (approved):

bgilbert (approved):

jschintag:
Thanks for the work. It works on s390x.

PR #540, opened 2021-05-12
Revert container to F33 to fix quay.io build failures

bgilbert:
Fedora 34 includes glibc 2.33, which tries to opportunistically implement `faccessat()` using `faccessat2()`, which is blocked on quay.io builders by a seccomp policy, which returns `-EPERM` instead of `-ENOSYS`, which glibc treats as a fatal error.

https://issues.redhat.com/browse/PROJQUAY-2006

This reverts commit 64d0ee4208ec01b14a6022b0aab1d3b978d08b27.


dustymabe (approved):
LGTM Thanks for digging in!

PR #544, opened 2021-05-13
Release 0.9.1

sohankunkerkar:
Major changes:
- Add Fedora 35 signing key; drop Fedora 32 signing key

Minor changes:
- install: Fix block device path in error message when disk is busy
- install: Ignore corrupt GPT on target disk unless saving partitions

Internal changes:
- rootmap: Ignore multipath devices

Edited

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -164,7 +164,7 @@ checksum = "ea221b5284a47e40033bf9b66f35f984ec0ea2931eb03505246cd27a963f981b"
 
 [[package]]
 name = "coreos-installer"
-version = "0.9.1"
+version = "0.9.2-alpha.0"
 dependencies = [
  "anyhow",
  "bincode",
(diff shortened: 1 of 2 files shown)


jlebon:
> blockdev: Fix symlink disk paths when creating disk objects

Maybe something like this instead:

> install: Fix block device path in error message when disk is busy

?

We should also add:

> Update container to Fedora 34

jlebon (approved):
PR itself LGTM.

bgilbert (approved):
PR contents look good.

Release notes:

- I concur with @jlebon's comment.
- Let's drop the master -> main rename; it's not user-facing.
- The prefixes are user-facing (subcommands) rather than developer-facing (modules).  Both of the `blockdev`s should be `install` in this case.
- Similarly, the rootmap one should be `rootmap: Ignore multipath devices`.
- For the corrupt GPT one, let's say something like `install: Ignore corrupt GPT on target disk unless saving partitions`.

PR #549, opened 2021-05-28
cmdline: don't require `-h` to list subcommands

bgilbert:
The top-level `coreos-installer` command prints a list of subcommands if run without arguments, but subcommands with subcommands don't:

    $ target/debug/coreos-installer iso
    Error: parsing arguments

    Caused by:
        unrecognized 'iso' subcommand

Enable the automatic subcommand list for subcommands with subcommands.  Don't enable it globally because that would enable it for leaf commands, which precludes zero-argument commands like `list-stream`.


jlebon (approved):
Oh yes nice! Though about this in the past but never looked into whether it's configurable.

PR #551, opened 2021-06-02
s390x: Add installation support for virtio-blk(vda) device backed by …

madeelibm:
…DASD/ECKD disk

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -976,6 +976,28 @@ pub fn is_dasd(device: &str) -> Result<bool> {
     Ok(target.to_string_lossy().starts_with("/dev/dasd"))
 }
 
+/// Checks if virtio-blk device is backed by an IBM DASD disk
+#[cfg(target_arch = "s390x")]
+pub fn is_vda_dasd(device: &str) -> Result<bool> {
+    let target =
+        canonicalize(device).with_context(|| format!("getting absolute path to {}", device))?;
+    let valid;
+    let cmd = Command::new("fdasd").arg("-i").arg(target).output()?;
+    if cmd.status.success() {
+        valid = true;
+        eprintln!("{} is backed by /dev/dasd", device)
+    }
+    else {
+        valid = false;
+        eprintln!("{} is NOT backed by /dev/dasd", device)
+    }
+    Ok(valid)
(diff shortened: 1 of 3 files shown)


src/blockdev.rs
```diff
@@ -976,6 +976,28 @@ pub fn is_dasd(device: &str) -> Result<bool> {
     Ok(target.to_string_lossy().starts_with("/dev/dasd"))
 }
 
+/// Checks if virtio-blk device is backed by an IBM DASD disk
+#[cfg(target_arch = "s390x")]
+pub fn is_vda_dasd(device: &str) -> Result<bool> {
+    let target =
+        canonicalize(device).with_context(|| format!("getting absolute path to {}", device))?;
+    let valid;
+    let cmd = Command::new("fdasd").arg("-i").arg(target).output()?;
```
mihajlov, line 985: It might be good to check whether fdasd is available.

mihajlov:
LGTM. Potentially, you could log a message if fdasd is not present, but that has no functional impact.

src/blockdev.rs
Prashanth684: correct. it is actually not available and would need to be added here: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh

madeelibm: It is available as part of s390x-utils.

Prashanth684: ah..my bad..that was for the initramfs...yeah all good then

bgilbert:
Thanks for the PR, @madeelibm!  @nikita-dubrovskii has also been working on a fix for [RHBZ 1960485](https://bugzilla.redhat.com/show_bug.cgi?id=1960485), and [his version](https://github.com/nikita-dubrovskii/coreos-installer/commit/a7917b404b78dc22d811f0a6a152d7898768e764) of the fix is closer to a mergeable state, so I'll close out this PR.

My apologies for the duplicated work; I realize it wasn't completely clear that Nikita was working on this.  I don't expect this problem to happen again.

bgilbert:
#552

PR #552, opened 2021-06-04
s390x: add support for virtio dasds

nikita-dubrovskii:
This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1960485

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -970,10 +970,36 @@ pub fn detect_formatted_sector_size(buf: &[u8]) -> Option<NonZeroU32> {
 }
 
 /// Checks if underlying device is IBM DASD disk
-pub fn is_dasd(device: &str) -> Result<bool> {
+pub fn is_dasd(device: &str, fd: Option<&mut File>) -> Result<bool> {
     let target =
         canonicalize(device).with_context(|| format!("getting absolute path to {}", device))?;
-    Ok(target.to_string_lossy().starts_with("/dev/dasd"))
+    if target.to_string_lossy().starts_with("/dev/dasd") {
+        return Ok(true);
+    }
+    let read_magic = |device: &str, disk: &mut File| -> Result<[u8; 4]> {
+        let offset = disk
+            .seek(SeekFrom::Current(0))
+            .with_context(|| format!("saving offset {}", device))?;
+        disk.seek(SeekFrom::Start(8194))
+            .with_context(|| format!("seeking {}", device))?;
+        let mut lbl = [0u8; 4];
+        disk.read_exact(&mut lbl)
+            .with_context(|| format!("reading label {}", device))?;
+        disk.seek(SeekFrom::Start(offset))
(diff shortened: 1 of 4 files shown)


src/blockdev.rs
```diff
@@ -970,10 +970,42 @@ pub fn detect_formatted_sector_size(buf: &[u8]) -> Option<NonZeroU32> {
...
+
+    if target.starts_with("/dev/vd") {
+        let cdl_magic = [0xd3, 0xf1, 0xe5, 0xd6];
+        let lbl;
+        if let Some(t) = fd {
+            lbl = read_magic(device, t)?;
+        } else {
+            let mut disk = File::open(device).with_context(|| format!("opening {}", device))?;
+            lbl = read_magic(device, &mut disk)?;
+        }
```
bgilbert, line 1002:
```suggestion
        let lbl = if let Some(t) = fd {
            read_magic(device, t)?
        } else {
            let mut disk = File::open(device).with_context(|| format!("opening {}", device))?;
            read_magic(device, &mut disk)?
        }
```

src/blockdev.rs
```diff
@@ -970,10 +970,42 @@ pub fn detect_formatted_sector_size(buf: &[u8]) -> Option<NonZeroU32> {
 }
 
 /// Checks if underlying device is IBM DASD disk
-pub fn is_dasd(device: &str) -> Result<bool> {
+pub fn is_dasd(device: &str, fd: Option<&mut File>) -> Result<bool> {
     let target =
         canonicalize(device).with_context(|| format!("getting absolute path to {}", device))?;
-    Ok(target.to_string_lossy().starts_with("/dev/dasd"))
+    if target.starts_with("/dev/dasd") {
```
bgilbert, line 976: Do these checks work?  `PathBuf.starts_with()` is [documented](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.starts_with) to compare only whole path components.

src/blockdev.rs
```diff
@@ -970,10 +970,42 @@ pub fn detect_formatted_sector_size(buf: &[u8]) -> Option<NonZeroU32> {
...
+        let cdl_magic = [0xd3, 0xf1, 0xe5, 0xd6];
+        let lbl;
+        if let Some(t) = fd {
+            lbl = read_magic(device, t)?;
+        } else {
+            let mut disk = File::open(device).with_context(|| format!("opening {}", device))?;
+            lbl = read_magic(device, &mut disk)?;
+        }
+        if cdl_magic == lbl {
+            eprintln!("detected virtblk ECKD CDL disk: {}", device);
```
bgilbert, line 1004: Let's not print from a utility function that's called multiple times.

bgilbert (commented):
LGTM generally.

src/blockdev.rs
nikita-dubrovskii: missed that part, fixed

bgilbert (approved):
:+1:

PR #553, opened 2021-06-07
s390x: simplify zipl installation

nikita-dubrovskii:
On s390x this simplifies installation of `zipl`. 
Also could be used for https://github.com/coreos/coreos-installer/pull/533

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -216,7 +216,8 @@ fn write_disk(
         #[cfg(target_arch = "s390x")]
         s390x::install_bootloader(
             mount.mountpoint(),
-            &config.device,
+            Some(&config.device),
+            true,
             config.firstboot_kargs.as_ref().map(|s| s.as_str()),
         )?;
     }
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use anyhow::{bail, Context, Result};
-use regex::RegexBuilder;
-use std::fs::read_to_string;
-use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
-use walkdir::WalkDir;
+use crate::install::{bls_entry_options_delete_and_append_kargs, visit_bls_entry_options};
+use crate::runcmd;
+use anyhow::{anyhow, Context, Result};
+use std::fs::{copy, create_dir_all, read_dir};
+use std::io::Write;
+use std::path::Path;
+use std::process::Command;
+use tempfile::Builder;
 
 /////////////////////////////////////////////////////////////////////////////
 // IBM Z Bootloader Support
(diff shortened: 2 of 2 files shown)


src/install.rs
```diff
@@ -185,6 +185,7 @@ fn write_disk(
             "boot",
             false,
             mount::MsFlags::empty(),
+            Some("/boot"),
```
bgilbert, line 188: We don't own `/boot` and can't mount an arbitrary filesystem on it.  For example, coreos-installer might be running on an arbitrary Linux distro on bare metal, to set up the virtual disk for a FCOS VM.  If this code raced with the installation of OS updates on the host, both the host and guest `/boot` filesystems could be corrupted.  That's why we mount `/boot` onto a temporary mountpoint.

Aside from simplifying zipl invocation, how is this change needed for #533?

nikita-dubrovskii: running `zipl` without any options requires `/boot` , otherwise we'd have to specify kernel, initrd,... . 
what if we make mounting of `/boot`  s390x-arch dependent?

src/s390x/zipl.rs
```diff
@@ -28,109 +27,54 @@ use walkdir::WalkDir;
 /// # Arguments
 /// * `boot` - path to boot partition mountpoint, i.e. smth like /boot
 /// * `disk` - smth like /dev/dasda
+/// * `firstboot_kargs` - enables ignition firstboot and adds provided kargs
 pub fn install_bootloader<P: AsRef<Path>>(
     boot: P,
-    disk: &str,
-    extra_kargs: Option<&str>,
+    disk: Option<&str>,
```
jlebon, line 33: I'm confused by this; why is this an `Option` now?

src/s390x/zipl.rs
```diff
@@ -28,109 +27,54 @@ use walkdir::WalkDir;
...
-    let bls_config_path = get_bls_config_path(&boot)?;
-    let kernel_path = get_kernel_path(&boot)?;
-    let initramfs_path = get_initramfs_path(&boot)?;
-
-    let kargs = format!(
-        "{} ignition.firstboot {}",
-        get_boot_kargs(bls_config_path)?,
-        extra_kargs.unwrap_or("")
+    let boot = boot.as_ref();
+    let conffile = "/tmp/zipl.conf";
```
jlebon, line 38: We should use the `tempfile` API instead here.

src/s390x/zipl.rs
```diff
@@ -28,109 +27,54 @@ use walkdir::WalkDir;
...
-        }
-        let name = entry.file_name().to_string_lossy();
-        if pat.starts_with('.') {
-            if !name.ends_with(pat) {
-                continue;
-            }
-        } else if !name.starts_with(pat) {
-            continue;
-        }
+    let blsdir = format!("{}/loader/entries", boot.to_string_lossy());
```
jlebon, line 65: `runcmd!` should let you pass an `OsString`, so you can do e.g.:

```suggestion
    let blsdir = boot.as_ref().join("loader/entries");
```

src/s390x/zipl.rs
```diff
@@ -28,109 +27,54 @@ use walkdir::WalkDir;
...
+        let mut extra = vec!["ignition.firstboot".to_string()];
+        extra.extend_from_slice(
+            &kargs
+                .split_whitespace()
+                .map(|s| s.to_string())
+                .collect::<Vec<String>>(),
+        );
+        Some(extra)
+    } else {
+        None
```
jlebon, line 56: Even if there are no firstboot kargs provided, we still need to append `ignition.firstboot`, no?

src/s390x/zipl.rs
nikita-dubrovskii: i want to make this function re-usable by `rdcore`, in this case we don't need to re-ipl machine again, so `disk` is an Option

src/s390x/zipl.rs
nikita-dubrovskii: we reboot machine right after `zipl`, so `/tmp` is clean again on the next boot. do we really need to use `tempfile` here?

src/s390x/zipl.rs
nikita-dubrovskii: i don't think so, during installation on s390x we have `firstboot_kargs`, so we add `ignition.firstboot`.

src/s390x/zipl.rs
jlebon: Ack, gotcha. Maybe we should change the docstring to something like:

> /// * `disk` - optional device path on which to run `chreipl`

?

src/s390x/zipl.rs
jlebon: But we're not guaranteed to have any `firstboot_kargs`.

src/s390x/zipl.rs
```diff
@@ -28,109 +29,59 @@ use walkdir::WalkDir;
...
-        } else if !name.starts_with(pat) {
-            continue;
-        }
+    let blsdir = boot.join("loader/entries");
+    runcmd!("zipl", "--blsdir", blsdir, "--config", conffile.path())?;
 
-        return Ok(entry.path().to_path_buf());
+    if let Some(extra) = extra {
+        visit_bls_entry_options(boot, |orig_options: &str| {
+            bls_entry_options_delete_and_append_kargs(orig_options, extra.as_slice(), &[], &[])
```
bgilbert, line 77: This is not the inverse of the append operation, since it can remove arguments that existed by default.  Would it make sense instead to make a copy of the `loader` directory in a tempdir, modify the kargs in the copy, and pass the copy to `--blsdir`?

src/s390x/zipl.rs
```diff
@@ -28,109 +29,59 @@ use walkdir::WalkDir;
...
-        let name = entry.file_name().to_string_lossy();
-        if pat.starts_with('.') {
-            if !name.ends_with(pat) {
-                continue;
-            }
-        } else if !name.starts_with(pat) {
-            continue;
-        }
+    let blsdir = boot.join("loader/entries");
+    runcmd!("zipl", "--blsdir", blsdir, "--config", conffile.path())?;
```
bgilbert, line 73: This is no longer sending stdout to `/dev/null`.

src/s390x/zipl.rs
```diff
@@ -28,109 +29,59 @@ use walkdir::WalkDir;
...
-    // filter out variable substitutions such as $ignition_firstboot
-    let opts = RegexBuilder::new(r"(^| )\$[a-zA-Z0-9_]+")
-        .build()
-        .unwrap()
-        .replace_all(opts, "")
-        .to_string();
-    Ok(opts)
+    if let Some(disk) = disk {
+        eprintln!("Updating re-IPL device");
+        runcmd!("chreipl", disk)?;
```
bgilbert, line 84: Also not sending stdout to `/dev/null`.

bgilbert (commented):
Overall, this looks like a nice cleanup, thanks!

src/s390x/zipl.rs
nikita-dubrovskii: yep, i made it by intention, on s390x it maybe useful to have zipl output (for debugging)

src/s390x/zipl.rs
nikita-dubrovskii: done. but not sure that simplification is that simple now ))

jlebon (approved):

PR #555, opened 2021-06-08
blockdev: mount filesystems into private namespace

### LINKED ISSUE #208: Consider using mount namespaces
In the various parts in which we mount partitions to read or write files into, consider using a mount namespace instead if supported since we normally require root anyway in most of those workflows. That way we guarantee that the mounts are just for us. This applies at least to `coreos-install install` and `coreos-install osmet pack`.

bgilbert:
This ensures that other processes on the system can't get file descriptors to files in our mounts, which would impede unmount.  It also guarantees that our mounts are unmounted even if we crash hard (though `kpartx` wouldn't be undone, if applicable).

In contrast to the discussion in #208, we hard-fail if unshare fails.  `CAP_SYS_ADMIN` is required for both `mount()` and `unshare(CLONE_NEWNS)`, and we recommend running containers privileged, so there shouldn't be any expected unshare failures.  This avoids accidentally depending on private namespace semantics while not necessarily getting them.

Fixes #208.

cc @jlebon @dustymabe @nikita-dubrovskii

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -15,10 +15,11 @@
 use anyhow::{anyhow, bail, Context, Result};
 use gptman::{GPTPartitionEntry, GPT};
 use nix::sys::stat::{major, minor};
-use nix::{errno::Errno, mount};
+use nix::{errno::Errno, mount, sched};
 use regex::Regex;
 use std::collections::HashMap;
 use std::convert::TryInto;
+use std::env;
 use std::fs::{
     canonicalize, metadata, read_dir, read_to_string, remove_dir, symlink_metadata, File,
     OpenOptions,
@@ -31,6 +32,7 @@ use std::os::unix::fs::FileTypeExt;
 use std::os::unix::io::AsRawFd;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::sync::Once;
 use std::thread::sleep;
 use std::time::Duration;
 use uuid::Uuid;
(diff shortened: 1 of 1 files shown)


dustymabe:
forgive my lack of understanding of rust.. does this account for https://github.com/coreos/coreos-installer/issues/208#issuecomment-608709194 ?

dustymabe:
> forgive my lack of understanding of rust.. does this account for [#208 (comment)](https://github.com/coreos/coreos-installer/issues/208#issuecomment-608709194) ?

Ahh, I think you are referring to that with: 

> In contrast to the discussion in #208, we hard-fail if unshare fails. CAP_SYS_ADMIN is required for both mount() and unshare(CLONE_NEWNS), and we recommend running containers privileged, so there shouldn't be any expected unshare failures. This avoids accidentally depending on private namespace semantics while not necessarily getting them.

So you're saying that we won't fallback in that case and it's OK that we don't?

bgilbert:
> So you're saying that we won't fallback in that case and it's OK that we don't?

Right.

src/blockdev.rs
```diff
@@ -431,6 +431,10 @@ impl Mount {
         // the partition contents if umount failed
         let mountpoint = tempdir.into_path();
 
+        // ensure we're in a private mount namespace so the mount isn't
+        // visible to the rest of the system
+        sched::unshare(sched::CloneFlags::CLONE_NEWNS).context("unsharing mount namespace")?;
```
jlebon, line 436: Just to hedge our bets re.

> In contrast to the discussion in #208, we hard-fail if unshare fails. CAP_SYS_ADMIN is required for both mount() and unshare(CLONE_NEWNS), and we recommend running containers privileged, so there shouldn't be any expected unshare failures.

should we add an env var to allow skipping this call?

Also, WDYT about using std::sync::Once for this? AFAIK, it's harmless to call multiple times, but it's not necessary either and could be misleading.

jlebon (approved):
LGTM overall.

src/blockdev.rs
bgilbert: With `Once`, the code turns into this:

```rust
        static UNSHARE_INIT: Once = Once::new();
        static mut UNSHARE_ERR: Option<String> = None;
        // Once handles synchronization, so mutable static is safe
        unsafe {
            UNSHARE_INIT.call_once(|| {
                // anyhow::Error isn't Copy; convert to String
                UNSHARE_ERR = sched::unshare(sched::CloneFlags::CLONE_NEWNS)
                    .context("unsharing mount namespace")
                    .map_err(|e| e.to_string())
                    .err();
            });
            match UNSHARE_ERR {
                None => (),
                Some(ref e) => bail!(e),
            }
        }
```

I think I'd rather stick with `// multiple unshare calls are safe`.  WDYT?

jlebon: Yuck, the error handling goop here is unfortunate.

> I think I'd rather stick with // multiple unshare calls are safe. WDYT?

SGTM and thanks for investigating!

bgilbert: Added environment variable and filed https://github.com/coreos/coreos-installer/issues/557 to remove it again.

nikita-dubrovskii:
LGTM

jlebon (approved):

PR #564, opened 2021-06-23
installer-service: add 'coreos.inst.mpath' boolean option to turn on multipath for the firstboot

nikita-dubrovskii:
During installation on s390x/zVM we cannot just pass extra commandline arguments to the installer,
but we can use a special karg which is later processed by `coreos-installer-service`.

### DIFF
--- a/scripts/coreos-installer-service
+++ b/scripts/coreos-installer-service
@@ -97,6 +97,10 @@ if karg_bool coreos.inst.insecure; then
     args+=("--insecure")
 fi
 
+if karg_bool coreos.inst.mpath; then
+   args+=("--append-karg" "rd.multipath=default" "--append-karg" "root=/dev/disk/by-label/dm-mpath-root" "--append-karg" "rw")
+fi
+
 # Ensure device nodes have been created
 udevadm settle
 


jlebon:
This is tricky. So what this PR is getting at is that the UX for enabling multipath at install time is not great, because instead of staying in the `coreos.inst.*` realm it requires either scripting the install or running it manually.

This will be greatly improved with https://issues.redhat.com/browse/GRPA-3512 though, which will make operations like "add kargs" much more ergonomic. That epic was originally planned for 4.9, but has been pushed back. (Though that doesn't mean we *can't* work on it before then.)

Personally, my vote is to just hold off until we have https://issues.redhat.com/browse/GRPA-3512 rather than adding more magic kargs. Meanwhile, we should already have documentation describing how to customize the coreos-installer invocation via Ignition (see e.g. [this](https://coreos.github.io/coreos-installer/customizing-install/#custom-coreos-installer-invocation) upstream doc).

Interested what others think.

nikita-dubrovskii:
> This is tricky. So what this PR is getting at is that the UX for enabling multipath at install time is not great, because instead of staying in the `coreos.inst.*` realm it requires either scripting the install or running it manually.

This is more for s390x/zVM users: now they add `rd.multipath=default` to the parmfile, but it doesn't enable multipath for the `first/subsequent- boot`. So anyhow they script it. The only benefit of this option - they don't have to enable multipath after installation.

jlebon:
> > This is tricky. So what this PR is getting at is that the UX for enabling multipath at install time is not great, because instead of staying in the `coreos.inst.*` realm it requires either scripting the install or running it manually.
> 
> This is more for s390x/zVM users: now they add `rd.multipath=default` to the `parmfile`, but it doesn't enable multipath for the `first/subsequent- boot`. So anyhow they script it. The only benefit of this option - they don't have to enable multipath after installation.

Right yup, this is what I mean by "it requires either scripting the install or running it manually", so that they can add `rd.multipath=default` via `--append-karg`.

> > Meanwhile, we should already have documentation describing how to customize the coreos-installer invocation via Ignition
> 
> yep, but on s390x it's more error-prone, because we'd have to call `zipl` twice. and anyhow we'd have to modify `parmfile`

Hmm, not sure I follow. Why would they have to modify the parmfile and call `zipl` twice themselves? `coreos-installer` will take care of doing that in the `--append-karg` case, right?

But yeah, it's *really* tempting to just automatically forward `rd.multipath=default` to the installed system but ughhh, the more implicit behaviour we add like this, the more it constrains us in the future when we want to clean up this stuff.

nikita-dubrovskii:
> Right yup, this is what I mean by "it requires either scripting the install or running it manually", so that they can add `rd.multipath=default` via `--append-karg`.
> Hmm, not sure I follow. Why would they have to modify the parmfile and call `zipl` twice themselves? `coreos-installer` will take care of doing that in the `--append-karg` case, right?

hmm, when we install on zVM we directly copy `coreos-live-kernel`, `coreos-live-initramfs`, `kenel-options.parm` (regular text file with contents equal to`/proc/cmdline`). 
I don't know how can we add more `--append-karg` options for the installer . 
If we specify extra config for the post-install , than we'd have to call `zipl` for the 2nd time

jlebon:
> > Right yup, this is what I mean by "it requires either scripting the install or running it manually", so that they can add `rd.multipath=default` via `--append-karg`.
> > Hmm, not sure I follow. Why would they have to modify the parmfile and call `zipl` twice themselves? `coreos-installer` will take care of doing that in the `--append-karg` case, right?
> 
> hmm, when we install on zVM we directly copy `coreos-live-kernel`, `coreos-live-initramfs`, `kenel-options.parm` (regular text file with contents equal to`/proc/cmdline`).
> I don't know how can we add more `--append-karg` options for the installer .

Yes, currently only a subset of kargs in the live environment are automatically forwarded to the installed system (mostly network-related). For anything else, they can't use `coreos.inst.*` kargs, so they need to specify an Ignition config which performs the custom install instead (see https://coreos.github.io/coreos-installer/customizing-install/#custom-coreos-installer-invocation). So there, they'd be able to do `--append-karg 'rd.multipath=default root=/dev/disk/by-label/dm-mpath-root rw'`.

This is very cumbersome, but we're hoping to improve the UX soon by making it trivial to generate such an Ignition config.

On other arches which support layered initrds, they can just add the CPIO created by `coreos-installer pxe ignition wrap` as a second initrd. On s390x, this isn't supported, though they could just append the CPIO to a copy of the live initramfs and load that. (Or of course, they can also use `ignition.config.url=http://...` in the parm file.)

jlebon:
Closing as per above discussion.

PR #566, opened 2021-06-29
blockdev: try rereading partition table up to 5 times

nikita-dubrovskii:
When we use DeviceMapper disk as `coreos.inst.install_dev=`, it may happen that `kpartx` fails to reread
partition table, and therefore fails the installation. Let's try up to 5 times, before giving up

```
[   16.056589] coreos-installer-service[1333]: device-mapper: resume ioctl on mpatha4  failed: Invalid argument
[   16.057027] coreos-installer-service[1333]: resume failed on mpatha4
[   16.057050] coreos-installer-service[1333]: Error: "kpartx" "-u" "-n" "/dev/dm-0" failed with exit code: 1
[   16.057069] coreos-installer-service[1333]: Retrying in 1 second
[   18.362478] coreos-installer-service[1333]: Read disk 20.3 MiB/3.5 GiB (0%)
[   23.399255] coreos-installer-service[1333]: Read disk 116.7 MiB/3.5 GiB (3%)
[  216.402889] coreos-installer-service[1333]: Read disk 3.5 GiB/3.5 GiB (100%)
[  216.757601] EXT4-fs (dm-1): mounted filesystem with ordered data mode. Opts: (null)
[  216.759492] coreos-installer-service[1333]: Writing Ignition config
[  216.759932] coreos-installer-service[1333]: Writing first-boot kernel arguments
[  216.759957] coreos-installer-service[1333]: Modifying kernel arguments
[  216.761553] coreos-installer-service[1333]: Installing bootloader
[  217.281371] coreos-installer-service[1333]: Updating re-IPL device
[  217.341450] coreos-installer-service[1333]: Install complete.
```

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -300,6 +300,15 @@ impl PartTableKpartx {
 
 impl PartTable for PartTableKpartx {
     fn reread(&mut self) -> Result<()> {
+        let delay = 1;
+        for _ in 0..4 {
+            match self.run_kpartx("-u") {
+                Ok(()) => return Ok(()),
+                Err(e) => eprintln!("Error: {}", e),
+            }
+            eprintln!("Retrying in {} second", delay);
+            sleep(Duration::from_secs(delay));
+        }
         self.run_kpartx("-u")
     }
 }


src/blockdev.rs
```diff
@@ -300,7 +300,16 @@ impl PartTableKpartx {
 
 impl PartTable for PartTableKpartx {
     fn reread(&mut self) -> Result<()> {
-        self.run_kpartx("-u")
+        let delay = 1;
+        for _ in 0..5 {
```
jlebon, line 304: I think this is essentially https://bugzilla.redhat.com/show_bug.cgi?id=1915965. And the recommendation from the SME there is indeed to just keep running `kpartx -u` if we want to be sure that it ran without falling back to another `udevadm settle`.

Probably worth linking to that RHBZ somewhere around here?

src/blockdev.rs
```diff
@@ -300,7 +300,16 @@ impl PartTableKpartx {
...
+            match self.run_kpartx("-u") {
+                Ok(()) => return Ok(()),
+                Err(e) => eprintln!("Error: {}", e),
+            }
+            eprintln!("Retrying in {} second", delay);
+            sleep(Duration::from_secs(delay));
```
jlebon, line 310: Minor: this will print and sleep one more time on the last iteration of the loop even though we won't really be retrying.

One approach I sometimes use in that pattern is to just retry one last time outside the loop since we'll fail anyway. That way you can also return that error itself. I.e. ...

src/blockdev.rs
```diff
@@ -300,7 +300,16 @@ impl PartTableKpartx {
...
+        }
+        bail!("Failed to reread partition table");
```
jlebon, line 312: ... basically this:

```suggestion
        self.run_kpartx("-u")
```

jlebon (approved):
Makes sense to me! LGTM overall, just some nits.

src/blockdev.rs
jlebon: Also would be good to add a e.g. `Related: https://bugzilla.redhat.com/show_bug.cgi?id=1974411` in the commit message.

jlebon (approved):

PR #567, opened 2021-07-01
docs: Add iso kargs modify/reset/show

travier:
Document additions from https://github.com/coreos/coreos-installer/pull/341

As far as I could figure out, this is the only undocumented option right now:
Fixes: https://github.com/coreos/coreos-installer/issues/568


travier:
Rendered at https://travier.github.io/coreos-installer/cmd/iso/

jlebon (approved):
Thanks a lot!

PR #571, opened 2021-07-06
install: restrict access permissions on /boot/ignition{,/config.ign}

bgilbert:
The Ignition config may contain secrets.  Don't expose it, or anything in its parent directory, to unprivileged processes.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -16,9 +16,11 @@ use anyhow::{bail, Context, Result};
 use lazy_static::lazy_static;
 use nix::mount;
 use regex::Regex;
-use std::fs::{copy as fscopy, create_dir_all, read_dir, File, OpenOptions};
+use std::fs::{
+    copy as fscopy, create_dir_all, read_dir, set_permissions, File, OpenOptions, Permissions,
+};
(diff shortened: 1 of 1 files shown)


src/install.rs
```diff
@@ -262,6 +264,13 @@ fn write_ignition(
                 config_dest.display()
             )
         })?;
+    // Ignition config may contain secrets; restrict to root
+    set_permissions(&config_dest, Permissions::from_mode(0o600)).with_context(|| {
```
travier, line 268: I think we should set the whole folder to `700` (in the `create_dir_all` above)

bgilbert:
Updated.

travier (approved):

travier:
This _might_ benefit from a conversion to https://github.com/coreos/openat-ext but this can definitely be done later.

jlebon (approved):

jlebon:
Probably worth mentioning https://github.com/coreos/fedora-coreos-tracker/issues/889 in the commit message for posterity.

On the other side of this, we should consider deleting the file after it's used. This is discussed in the tracker issue.

PR #577, opened 2021-07-23
rdcore/kargs: don't fail if --create-if-changed file already exists

bgilbert:
Other services in the initrd may want to use the same flag file to coordinate reboots.  See discussion in https://github.com/openshift/os/pull/574#discussion_r675688876.

Requires https://github.com/coreos/coreos-installer/pull/579 to fix CI.


jlebon (approved):
Thanks!

cgwalters (approved):

PR #583, opened 2021-07-28
release-checklist: clarify crates.io access

jlebon:
The before-last bullet mentions `crates.io`, but for most people, access
to it will be through the "Fedora CoreOS Crates Owners" GitHub group,
which is the very next bullet. Shift mention of `crates.io` there.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-checklist.md
+++ b/.github/ISSUE_TEMPLATE/release-checklist.md
@@ -17,8 +17,8 @@ This guide requires:
  * A verified account on crates.io
  * An account on quay.io
  * Write access to this GitHub project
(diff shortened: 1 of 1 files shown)


bgilbert (approved):
Thanks!

No good deed goes unpunished: the checklists for Afterburn, bootupd, openssh-keys, ssh-key-dir, and Zincati need variants of the same update.

jlebon:
Opened https://github.com/coreos/ssh-key-dir/pull/41 and https://github.com/coreos/openssh-keys/pull/52.

For the other ones, I'm honestly not sure whether we really should be publishing to crates.io. I'd like to discuss in the individual repos whether we should drop that step from the release process.

PR #591, opened 2021-08-10
docs: Use security-opt label=disable for ISO download

travier:
docs: Use security-opt label=disable for ISO & PXE download

Only disable SELinux labelling for ISO & PXE download as we only need
access to the current directory and not all privileges granted by
`--privileged`.

---

docs: Split long lines to improve readability


travier:
Rendered: https://travier.github.io/coreos-installer/getting-started/

bgilbert (approved):
LGTM

jlebon (approved):

PR #593, opened 2021-08-11
live: centralize embed area management in a single struct

bgilbert:
Add an `IsoConfig` struct that centralizes management of embed areas, each represented by a `Region` struct containing offset/length/data.

There's a lot of code motion here; it might make sense to just rereview `live.rs` from scratch.

Goals for this refactor:

- Move logic for high-level operations out of subcommand implementations
- Support streaming of Ignition and kargs changes in one pass
- Eliminate `ContentFile`, which has always been difficult to use

Non-goals:

- Lazy reads.  For simplicity, we eagerly read all Ignition and karg embed areas (and validate karg embed areas) during initialization.


src/live.rs
```diff
@@ -187,440 +160,450 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+impl KargEmbedAreas {
+    pub fn exists_in(file: &mut File) -> Result<bool> {
+        let region = Region::read(
+            file,
+            32768 - COREOS_IGNITION_HEADER_SIZE - COREOS_KARG_EMBED_AREA_HEADER_SIZE,
+            8,
+        )
+        .context("reading karg embed magic number")?;
+        Ok(region.contents == COREOS_KARG_EMBED_AREA_HEADER_MAGIC)
     }
```
jlebon, line 459: Would it be cleaner to fold this into `for_file` and have that return a `Result<Option<Self>>` instead?

src/live.rs
```diff
@@ -187,440 +160,450 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
-        file.read_exact(&mut buf)
-            .with_context(|| format!("reading karg area at offset {}", offset))?;
-        String::from_utf8(buf).context("invalid UTF-8 in karg area")?
-    };
-    Ok(area.trim_end_matches('#').trim().into())
-}
+#[derive(Serialize)]
+struct KargEmbedAreas {
+    #[serde(rename = "default_kargs")]
+    default_region: Region,
```
jlebon, line 440: Why not just call it `default_kargs`?

src/live.rs
```diff
@@ -187,440 +160,450 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
+    #[serde(skip_serializing)]
+    default_args: String,
 
-enum ContentFile {
-    ForStdout(File),
-    InPlace(File),
-    Copied((NamedTempFile, PathBuf)),
+    #[serde(rename = "kargs")]
+    regions: Vec<Region>,
```
jlebon, line 445: And this one just `kargs`?

jlebon (commented):
This looks really good! Love all the new tests.

The `impl Stream for [&Region]` bit is very close to what osmet does and also the minimal ISO work. Will think on whether it's worth abstracting further.

src/live.rs
```diff
@@ -177,440 +149,450 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
 
-#[derive(Serialize)]
-struct KargEmbedAreas<'a> {
-    #[serde(skip_serializing)]
-    file: &'a mut File,
-    length: usize,
-    default_kargs_offset: u64,
-    kargs_offsets: Vec<u64>,
+// output_path should be None if not outputting, or Some(output_path_argument)
+fn open_live_iso(input_path: &str, output_path: Option<Option<&String>>) -> Result<File> {
```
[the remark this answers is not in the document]
bgilbert: Hmm, I'm not sure I follow.  What function signature are you thinking of?

src/live.rs
```diff
@@ -187,440 +160,450 @@ pub fn pxe_ignition_unwrap(config: &PxeIgnitionUnwrapConfig) -> Result<()> {
...
-        let new_area = self.format_embed_area(kargs)?;
+trait Stream {
+    fn stream(&self, input: &mut File, writer: &mut (impl Write + ?Sized)) -> Result<()>;
+}
+
+impl Stream for [&Region] {
+    fn stream(&self, input: &mut File, writer: &mut (impl Write + ?Sized)) -> Result<()> {
+        input.seek(SeekFrom::Start(0)).context("seeking to start")?;
+
+        let mut regions: Vec<&&Region> = self.iter().filter(|r| r.modified).collect();
```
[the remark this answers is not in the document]
bgilbert: Hmm.  We shouldn't ever have any, and nothing bad should happen if we do, so it seems like it'd make the code slightly misleading?

src/live.rs
bgilbert: I feel like it's clearer to be explicit that these fields have regions of unprocessed bytes (and in this case, multiple copies of same), rather than kargs per se.

bgilbert:
Maybe something like `impl Stream for [Streamable]`, where the latter is a trait with `offset` and `write` methods?

src/live.rs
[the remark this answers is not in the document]
jlebon: All we want to know is whether we need to open for writing too or not. We could have `output_path` just be a `bool` (and renamed to e.g. `with_write`), and the callers do `config.output.is_none()`. That makes it clearer that we don't actually *do* anything with the path itself.

src/live.rs
jlebon: WDYT about splitting the difference and calling them `default_kargs_region` and `kargs_regions`? And then we can drop the serde renames.

src/live.rs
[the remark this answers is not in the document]
bgilbert: Oh, I see.  The idea is that `open_live_iso` and `write_live_iso` are helpers that encapsulate the semantics of the `output_path` argument (and make assumptions about each other's behavior), so the callers don't have to think about it.

src/live.rs
bgilbert: Could do, though the renames don't bother me that much.  The serde renames are for `--header`, which has a different perspective than the code does; `region` is an implementation detail there.

src/live.rs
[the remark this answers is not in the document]
jlebon: SGTM!

jlebon (approved):

Issue #124, opened 2020-01-06
Configurability for automated installs without kargs


jlebon:
This came up during review of https://github.com/coreos/coreos-installer/pull/119.

There's some interest there in expanding the kargs functionality. I'd like to make sure this is the path we want to go down, rather than rework the UX to something friendlier.

I know there's some tension here, since we kinda have to keep supporting the base kargs the legacy installer did for backwards compatibility reasons.

OTOH, now that the installer is a simple binary written in a sane language, I think it'd be pretty sweet to have e.g. a YAML file or some other input file that can be dropped in via Ignition (and/or injected into the ISO) which is read directly by coreos-installer.

Then, the legacy kargs compat path would just be translating the kargs into this YAML file. We could also support a *single* karg, which would be just a URL to the YAML file to fetch.

jlebon:
(The idea is that it encourages a workflow where this file would be checked into an SCM, changes are audited, reviewed, etc... and installs are fully reproducible.)

dustymabe:
I think the kargs approach is a nice MVP for an installer, but we obviously don't want a million options, which is part of the reason we now run in the real root. i.e. people can use kargs for simple installs, but if they need something complicated they can use an ignition config to set up the environment and then call the installer.

lucab:
My personal understanding (and preference) is that we slowly move away from kargs for all cases (both simpler and more complex).

The rationale is the one mentioned by @jlebon. The final goal looks like this: https://github.com/coreos/tectonic-installer/blob/track-1/platforms/metal/cl/coreos-install.yaml.tmpl.

For FCOS specifically, I'd jump straight to document the "real rootfs" usage.

jlebon:
> My personal understanding (and preference) is that we slowly move away from kargs for all cases (both simpler and more complex).

+1

This is why I'm leaning towards a simpler approach which doesn't require everyone writing down systemd units and scripts that call the installer. IOW:

- simple/vanilla installs can just drop a file in a specific location (e.g. `/etc/coreos-installer.yaml`)
- one can order pre/post-install tasks by just having their units ordered wrt `coreos-installer.service`
- more complex workflows can just call `coreos-installer` directly
- keep kargs as the legacy interface

Anyway, to re-iterate I don't think we need to solve this right now since the "call `coreos-installer` directly" approach should work for now.

dustymabe:
[...]

PR #594, opened 2021-08-12
cmdline: switch to structopt

bgilbert:
Preparatory to #124, and also a nice cleanup.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -12,15 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// For consistency, have all parse_*() functions return Result.
-#![allow(clippy::unnecessary_wraps)]
-
-use anyhow::{bail, Context, Result};
-use clap::{crate_version, App, AppSettings, Arg, ArgMatches, SubCommand};
+use anyhow::{bail, Context, Error, Result};
 use reqwest::Url;
+use std::default::Default;
 use std::fs::{File, OpenOptions};
 use std::num::NonZeroU32;
 use std::path::Path;
+use std::str::FromStr;
+use std::string::ToString;
+use structopt::clap::AppSettings;
+use structopt::StructOpt;
 
 use crate::blockdev::*;
 use crate::download::*;
@@ -29,6 +30,10 @@ use crate::io::IgnitionHash;
 use crate::s390x::dasd_try_get_sector_size;
 use crate::source::*;
 
+// Args are listed in --help in the order declared in these structs/enums.
+// Please keep the entire help text to 80 columns.
+
+// Exported, flattened subcommand enum with postprocessed configs
 pub enum Config {
     Install(InstallConfig),
     Download(DownloadConfig),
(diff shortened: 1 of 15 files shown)


jlebon (commented):
I haven't done a deep review yet, but overall this looks really good to me!

You're probably aware of this, though my understanding is that `structopt` will be folded into `clap` in the future. AFAIK the API should mostly remain the same, though we'll probably need to adapt a few things here.

bgilbert:
Tested all command-line arguments in both `coreos-installer` and `rdcore`, except `osmet pack` (which should run in CI) and `osmet unpack`.  Verified that all help texts are character-for-character identical with the previous implementation.  Dropping WIP.

cgwalters (commented):
I only skimmed this but it looks great to me!

While I know structopt has some cost in compilation time, to me it's very worth it.  structopt and serde are great examples of how Rust's macro system enables very high level and elegant things in a systems programming language.

src/live.rs
```diff
@@ -36,19 +36,29 @@ const COREOS_KARG_EMBED_AREA_HEADER_SIZE: u64 = 72;
...
-pub fn iso_embed(config: &IsoIgnitionEmbedConfig) -> Result<()> {
+pub fn iso_embed(config: &IsoEmbedConfig) -> Result<()> {
     eprintln!("`iso embed` is deprecated; use `iso ignition embed`.  Continuing.");
-    iso_ignition_embed(config)
+    iso_ignition_embed(&IsoIgnitionEmbedConfig {
+        force: config.force,
+        ignition_file: config.config.clone(),
+        output: config.output.clone(),
+        input: config.input.clone(),
+    })
```
jlebon, line 46: Can't the `IsoCmd` enum just directly use the same `IsoIgnition*Config` structs to avoid this?

jlebon (approved):
LGTM overall! I had more suggestions, but following commits implemented them. :)

bgilbert:
Rebased on #597.

src/cmdline.rs
```diff
@@ -49,6 +54,193 @@ pub enum Config {
...
+    // users shouldn't be interacting with this command normally
+    #[structopt(setting(AppSettings::Hidden))]
+    Osmet(OsmetCmd),
+    /// Commands to manage a CoreOS live PXE image
+    Pxe(PxeCmd),
+}
+
+#[derive(Debug, StructOpt)]
+enum IsoCmd {
+    /// Embed an Ignition config in an ISO image
```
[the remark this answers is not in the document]
bgilbert: Yeah, they're still used in the help texts of the deprecated subcommands.

src/live.rs
bgilbert: That was my plan initially, but `iso ignition embed` renamed `-c`/`--config` to `-i`/`--ignition-file`.  We could use `alias`, but that would allow both subcommands to accept both arguments and would mess up the help text for one of them.

After some thought, I think the approach here is actually better, because it clearly separates the semantics of the old and new subcommands.  This way any options added to the new subcommands will not be reflected in the old ones.

jlebon (approved):

PR #597, opened 2021-08-16
rdcore/kargs: add `--current` to do a dry run on current kargs

bgilbert:
In the live ISO/PXE case, Ignition kargs stage has nowhere to write the new kargs, but should still succeed if no changes need to be made.  Add `--current` to apply the kargs changes to the contents of `/proc/cmdline`.  The Ignition kargs wrapper can use this with `--create-if-changed` to decide whether to fail.


jlebon (approved):
Nice and simple.

PR #599, opened 2021-08-17
signing-keys: add Fedora 36 key

dustymabe:
We have a rawhide stream now and rawhide just switched to F36.


cgwalters (approved):

bgilbert:
Historically we've dropped the old key at the same time.  It seems like the F33 key can go now.

dustymabe:
> Historically we've dropped the old key at the same time. It seems like the F33 key can go now.

I hesitated. This isn't part of our normal transition (i.e. traditionally we did this as part of the transition of `next`/`next-devel`). With the rawhide stream we're shifting things forward a bit and I wasn't sure if we should also more aggressively trim the old key.

bgilbert:
We don't support non-current releases, and `stable` switched to F34 three months ago, so I think it's fine to drop.  The rebase checklist doesn't mention dropping the old key (maybe it should?) so there's a risk we'll forget.

dustymabe:
+1 - will open a PR

dustymabe:
> +1 - will open a PR

https://github.com/coreos/coreos-installer/pull/600

bgilbert:
Rebase checklist update in https://github.com/coreos/fedora-coreos-tracker/pull/930.

PR #605, opened 2021-08-24
install: make platform-specific bootloader changes when `-p` specified

bgilbert:
If `-p` is specified and `/boot/coreos/platforms.json` exists, look up the specified platform and the `metal` platform in that file.  Remove any metal-specific kernel arguments or GRUB commands specified there, and inject any kargs/commands for the new platform.  This allows us to apply a platform-specific console configuration at install time.

Set the platform before performing kargs modification to allow the latter to change a platform-specific default console.

When installing via `coreos.inst` kargs, persist any `console=` kargs to the destination system, as suggested in https://github.com/coreos/fedora-coreos-tracker/issues/567#issuecomment-901801590.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -14,6 +14,9 @@
 
 use anyhow::{bail, Context, Result};
 use nix::mount;
+use regex::{Captures, Regex};
+use serde::Deserialize;
+use std::collections::HashMap;
 use std::fs::{self, File, OpenOptions, Permissions};
 use std::io::{self, Seek, SeekFrom, Write};
 use std::num::NonZeroU32;
@@ -27,6 +30,10 @@ use crate::io::*;
 #[cfg(target_arch = "s390x")]
 use crate::s390x;
 use crate::source::*;
+
+// Match the grub.cfg console settings commands in
+// https://github.com/coreos/coreos-assembler/blob/main/src/grub.cfg
+const GRUB_CFG_CONSOLE_SETTINGS_RE: &str = r"(?P<prefix>\n# CONSOLE-SETTINGS-START\n)(?P<commands>([^\n]*\n)*)(?P<suffix># CONSOLE-SETTINGS-END\n)";
 
 pub fn install(config: InstallConfig) -> Result<()> {
     // evaluate config files
@@ -517,7 +524,16 @@ fn write_firstboot_kargs(mountpoint: &Path, args: &str) -> Result<()> {
     Ok(())
 }
 
-/// Override the platform ID.
+#[derive(Clone, Default, Deserialize)]
+struct PlatformSpec {
+    #[serde(default)]
+    grub_commands: Vec<String>,
+    #[serde(default)]
+    kernel_arguments: Vec<String>,
+}
+
+/// Override the platform ID.  Add any kernel arguments and grub.cfg
+/// directives specified for this platform in platforms.json.
 fn write_platform(mountpoint: &Path, platform: &str) -> Result<()> {
     // early return if setting the platform to the default value, since
     // otherwise we'll think we failed to set it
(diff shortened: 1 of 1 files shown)


bgilbert:
Notably, while this PR allows the user to override console kargs, it doesn't provide a way to override the GRUB console commands.

bgilbert:
Tested the contents of `grub.cfg` and kargs after `install -p` with:
- No `platforms.json`
- Target platform omitted from `platforms.json`
- Target platform with kargs + GRUB commands in `platforms.json`
- Target platform with kargs only in `platforms.json`
- Target platform with GRUB commands only in `platforms.yaml`

Ready for review.

bgilbert:
Rebased and reworked.  Tested the contents of `grub.cfg` and kargs after install with:

- no `-p`,
- `-p aws` (platform-specific kargs and GRUB commands),
- `-p aliyun` (no platform-specific kargs or commands),
- `-p metal` (redundant),

with each of:

- no `platforms.json`,
- `platforms.json` from https://github.com/coreos/fedora-coreos-config/pull/1181,
- modified `platforms.json` with `metal`-specific kargs and GRUB commands.

Also tested:

- `-p aws --delete-karg=console=ttyS0,115200n8` with `platforms.json`
- Forwarding `console` kargs when installing with `coreos.inst.install_dev`

Ready for review.

cgwalters:
I hadn't really looked at this before.  I totally understand the goals, but this is also taking us even farther away from the "fancy dd" model.  The feature requires coreos-assembler as is and is another thing that won't work for RHEL for Edge.  (Of course, it could be made to, and I do hope eventually they support Ignition at least optionally which would greatly simplify the story, but just noting)

How much do we need this versus just documenting how to use the existing kernel argument switches?

bgilbert:
Updated!

scripts/coreos-installer-service
```diff
@@ -10,6 +10,9 @@ PERSIST_KERNEL_NET_PARAMS=("ipv6.disable" "net.ifnames" "net.naming-scheme")
 # List from https://www.mankier.com/7/dracut.cmdline#Description-Network
 PERSIST_DRACUT_NET_PARAMS=("ip" "ifname" "rd.route" "bootdev" "BOOTIF" "rd.bootif" "nameserver" "rd.peerdns" "biosdevname" "vlan" "bond" "team" "bridge" "rd.net.timeout.carrier" "coreos.no_persist_ip" "coreos.force_persist_ip")
 
+# Console params to persist
+PERSIST_CONSOLE_PARAMS=("console")
```
jlebon, line 14: I have mixed feelings about this. On the one hand, I think as a heuristic this makes sense (I know I suggested it myself at least once), but OTOH this is adding more kargs magic which increases the divide between the `coreos.inst` path vs manual installs.

I guess one thing that helps is that in the `coreos.inst` path, we automatically reboot which strongly implies that the console setting should be the same.

Maybe what would help too is if we had a `coreos-installer install` option like `--copy-console-kargs`. Then we could frame it as "the `coreos.inst` path implies `--copy-console-kargs`, which can also be enabled in a manual install". (On that note, maybe we could even also have `--copy-network-kargs` and `--copy-s390x-kargs` too and drain everything here into coreos-installer proper.)

Anyway, not a blocker, just voicing some thoughts.

scripts/coreos-installer-service
```diff
@@ -97,7 +100,7 @@ for item in "${cmdline[@]}"; do
             firstboot_args+="${item} "
         fi
     done
-    for param in "${PERSIST_S390X_PARAMS[@]}"; do
+    for param in "${PERSIST_CONSOLE_PARAMS[@]}" "${PERSIST_S390X_PARAMS[@]}"; do
```
jlebon, line 103: For `console` kargs specifically, we should probably also delete all previous platform-default `console` kargs before persisting those from the live environment. Although the last one wins, non-last console kargs still have an effect, and it wouldn't be correct to say we're exactly matching the console kargs of the live environment.

coreos-installer doesn't support deleting kargs by key only, but with a `--copy-console-kargs`, it could learn to delete all the metal kargs (and not append any platform kargs) if it picks up any `console` kargs from the live environment.

scripts/coreos-installer-service
bgilbert: https://github.com/coreos/coreos-installer/pull/877#discussion_r890675050

scripts/coreos-installer-service
bgilbert: https://github.com/coreos/coreos-installer/pull/877#discussion_r890675238

bgilbert (commented):
Thanks for the review!  The last commit needs some additional thought, so I've moved it to a separate PR #877.

bgilbert:
Updated!

jlebon (approved):
LGTM!

PR #613, opened 2021-09-07
dracut: also ship zipl_helper.device-mapper on s390x

julianwiedmann:
When zipl deals with device-mapper targets, it potentially requires an
additional helper script. Install this in the initramfs.

This fixes https://github.com/openshift/os/issues/615.

### DIFF
--- a/dracut/50rdcore/module-setup.sh
+++ b/dracut/50rdcore/module-setup.sh
@@ -9,5 +9,6 @@ install() {
     if [[ "$_arch" == "s390x" ]]; then
         inst_multiple zipl
         inst /lib/s390-tools/stage3.bin
+        inst /lib/s390-tools/zipl_helper.device-mapper
     fi
 }


julianwiedmann:
cc @nikita-dubrovskii

nikita-dubrovskii:
LGTM

jlebon (approved):

jlebon:
Are there any other `zipl_helper.*` scripts we should ship?

bgilbert:
Looks like that's the only one in s390-tools.

bgilbert (approved):

cgwalters:
Another option to consider is not including `zipl` in the initramfs, but writing a stamp file that causes it to run on the real root.

bgilbert:
@cgwalters For Ignition kernel arguments and FIPS, we use rdcore to apply kernel arguments and reboot before we ever reach the real root.

julianwiedmann:
> Are there any other `zipl_helper.*` scripts we should ship?

I was wondering about the same, and also whether there's any other zipl artifacts that we're still missing. From what I can tell users are free to add their own helper scripts (to support other device types), but `device-mapper` is the only one provided by upstream.

They also ship the identical script as `chreipl_helper.device-mapper` for use by the `chreipl` tool. But that tool hasn't found its way into the CoreOS initramfs yet ...

PR #616, opened 2021-09-08
osmet: produce JSON from `fiemap` debug subcommand

bgilbert:
You never know when you might want a machine-readable extent map.

### DIFF
--- a/src/osmet/mod.rs
+++ b/src/osmet/mod.rs
@@ -74,8 +74,20 @@ struct Osmet {
     size: u64,
 }
 
+#[derive(Serialize)]
+struct FiemapOutput {
+    extents: Vec<Extent>,
+}
+
 pub fn osmet_fiemap(config: &OsmetFiemapConfig) -> Result<()> {
-    eprintln!("{:?}", fiemap_path(config.file.as_str().as_ref())?);
+    let output = FiemapOutput {
+        extents: fiemap_path(config.file.as_str().as_ref())?,
+    };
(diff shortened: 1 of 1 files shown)


src/osmet/mod.rs
```diff
@@ -74,8 +74,20 @@ struct Osmet {
...
+
 pub fn osmet_fiemap(config: &OsmetFiemapConfig) -> Result<()> {
-    eprintln!("{:?}", fiemap_path(config.file.as_str().as_ref())?);
+    let output = FiemapOutput {
+        extents: fiemap_path(config.file.as_str().as_ref())?,
+    };
+    serde_json::to_writer_pretty(std::io::stdout(), &output)
+        .context("failed to serialize extents")?;
+    std::io::stdout()
+        .write_all(b"\n")
```
cgwalters, line 89: Doesn't matter here, but I would use [the lock method](https://doc.rust-lang.org/std/io/struct.Stdout.html#method.lock).  It can start to matter when doing nontrivial amounts of I/O, and it's not a difficult pattern, so better to use it consistently.

cgwalters (approved):

src/osmet/mod.rs
bgilbert: :+1: We don't generally do that now; will leave it for a followup.

bgilbert: https://github.com/coreos/coreos-installer/pull/618

PR #620, opened 2021-09-13
iso: add minimal ISO9660 parser

jlebon:
We do a lot of sophisticated things with our live ISOs. An upcoming
patch will push things yet further by supporting embedding in the ISO a
compressed minimal version of itself.

Currently, these modifications (together with Ignition embedding and
kernel arg modifications) require support from coreos-assembler writing
pointers in the System Area section to other places in the disc. With
the minimal ISO work, we'll need to able to find every file on the disc,
which would grow the complexity of these hacks further.

This patch adds a basic parser for the ISO9660 filesystem. It only
understands the minimal set of things we need, but it should provide a
good base for cleaning up our current approach, as well as for adding
minimal ISO support in a future patch.

As a way to exercise the code, add a new `coreos-installer iso inspect`
command which simply prints the volume descriptors and file contents of
any ISO.

### DIFF
--- a/src/iso9660.rs
+++ b/src/iso9660.rs
@@ -1,0 +1,514 @@
+// Copyright 2021 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Minimal ISO 9660 parser. This is not a comprehensive parser; we only
+//! parse out the fields we care about. Extensions such as Rock Ridge
+//! and Joliet are not supported.
+//!
+//! The official specification is not free. The primary reference used
+//! for this module is https://wiki.osdev.org/ISO_9660.
+
+// An initial version of this module used the zerocopy crate to try to deserialize directly from
+// the on-disk ISO file in with zero copying. It works, but it's non-trivial and the performance
+// difference from just copying stuff didn't justify it.
+
+// Many magic numbers corresponding to offsets and lengths have not been const-ified. It should be
+// straightforward to see to what they correspond using the referenced linked above.
+
+use std::fs;
+use std::io::{BufReader, Read, Seek, SeekFrom};
+use std::path::{Path, PathBuf};
+
+use anyhow::{anyhow, bail, Context, Result};
+use bytes::{Buf, Bytes};
+use serde::Serialize;
+
+use crate::io::{LimitReader, BUFFER_SIZE};
+
+// technically the standard supports others, but this is the only one we support
+const ISO9660_SECTOR_SIZE: usize = 2048;
+
+#[derive(Debug, Serialize)]
+pub struct IsoFs {
+    descriptors: Vec<VolumeDescriptor>,
+    #[serde(skip_serializing)]
+    file: fs::File,
+}
+
+impl IsoFs {
+    pub fn from_file(mut file: fs::File) -> Result<Self> {
+        let descriptors = get_volume_descriptors(&mut file)?;
+        Ok(Self { descriptors, file })
+    }
+
+    pub fn as_file(&mut self) -> Result<&mut fs::File> {
+        self.file
+            .seek(SeekFrom::Start(0))
+            .context("seeking to start of ISO")?;
+        Ok(&mut self.file)
+    }
+
+    pub fn get_root_directory(&self) -> Result<Directory> {
+        let primary = self
+            .get_primary_volume_descriptor()
+            .context("getting root directory")?;
+        Ok(primary.root.clone())
+    }
+
+    pub fn walk(&mut self) -> Result<IsoFsWalkIterator> {
+        let root_dir = self.get_root_directory()?;
+        let buf = self.list_dir(&root_dir)?;
+        Ok(IsoFsWalkIterator {
+            iso: &mut self.file,
+            parent_dirs: Vec::new(),
+            current_dir: Some(buf),
+            dirpath: PathBuf::new(),
+        })
+    }
+
+    /// Returns an iterator over the records of a directory.
+    pub fn list_dir(&mut self, dir: &Directory) -> Result<IsoFsIterator> {
+        IsoFsIterator::new(&mut self.file, dir)
+    }
+
+    /// Returns the record for a specific path.
+    pub fn get_path(&mut self, path: &str) -> Result<DirectoryRecord> {
+        let root_dir = self.get_root_directory()?;
+        let as_path = Path::new(path);
+        let mut parent_dir = if let Some(p) = as_path.parent() {
+            p
+        } else {
+            return Ok(DirectoryRecord::Directory(root_dir));
+        };
+        if parent_dir.has_root() {
+            parent_dir = parent_dir
+                .strip_prefix("/")
+                .with_context(|| format!("making path '{}' relative", path))?;
+        }
+        let filename = as_path
+            .file_name()
+            .ok_or_else(|| anyhow!("path {} has no base", path))?;
+        let filename = filename.to_str().unwrap(); // `path` is &str
+
+        let mut dir = root_dir;
+        for component in parent_dir.components() {
+            if let std::path::Component::Normal(c) = component {
+                let c = c.to_str().unwrap(); // `path` is &str
+                match self.get_dir_record(&dir, c)? {
+                    Some(DirectoryRecord::Directory(d)) => dir = d,
+                    Some(DirectoryRecord::File(_)) => {
+                        bail!("component {:?} in path {} is not a directory", c, path)
+                    }
+                    None => bail!("intermediate directory {} does not exist", c),
+                }
+            } else {
+                bail!("path is not canonical: {}", path);
+            }
+        }
+
+        self.get_dir_record(&dir, filename)?.ok_or_else(|| {
+            anyhow!(
+                "no record for {} in directory {}",
+                filename,
+                parent_dir.display()
+            )
+        })
+    }
+
+    /// Returns the record for a specific directory.
+    pub fn get_dir(&mut self, path: &str) -> Result<Directory> {
+        match self.get_path(path)? {
+            DirectoryRecord::Directory(d) => Ok(d),
+            DirectoryRecord::File(_) => Err(anyhow!("path {} is a file", path)),
+        }
+    }
+
+    /// Returns the record for a specific file.
+    pub fn get_file(&mut self, path: &str) -> Result<File> {
+        match self.get_path(path)? {
+            DirectoryRecord::Directory(_) => Err(anyhow!("path {} is a directory", path)),
+            DirectoryRecord::File(f) => Ok(f),
+        }
+    }
+
+    /// Returns the record for a specific name in a directory if it exists.
+    pub fn get_dir_record(
+        &mut self,
+        dir: &Directory,
+        name: &str,
+    ) -> Result<Option<DirectoryRecord>> {
+        for record in self
+            .list_dir(dir)
+            .with_context(|| format!("listing directory {}", dir.name))?
+        {
+            let record = record?;
+            match &record {
+                DirectoryRecord::Directory(d) if d.name == name => return Ok(Some(record)),
+                DirectoryRecord::File(f) if f.name == name => return Ok(Some(record)),
+                _ => continue,
+            }
+        }
+        Ok(None)
+    }
+
+    /// Returns a reader for a file record.
+    pub fn read_file(&mut self, file: &File) -> Result<impl Read + '_> {
+        self.file
+            .seek(SeekFrom::Start(file.address))
+            .with_context(|| format!("seeking to file {}", file.name))?;
+        Ok(BufReader::with_capacity(
+            BUFFER_SIZE,
+            LimitReader::new(&self.file, file.length as u64, None),
+        ))
+    }
+
+    fn get_primary_volume_descriptor(&self) -> Result<&PrimaryVolumeDescriptor> {
+        for d in &self.descriptors {
+            if let VolumeDescriptor::Primary(p) = d {
+                return Ok(p);
+            }
+        }
+        Err(anyhow!("no primary volume descriptor found in ISO"))
+    }
+}
+
+#[derive(Debug, Serialize)]
+#[serde(tag = "type", rename_all = "lowercase")]
+enum VolumeDescriptor {
+    Boot(BootVolumeDescriptor),
+    Primary(PrimaryVolumeDescriptor),
+    Supplementary,
+    Unknown { type_id: u8 },
+}
+
+#[derive(Debug, Serialize)]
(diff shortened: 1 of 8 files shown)


src/live.rs
```diff
@@ -647,6 +648,53 @@ fn extract_cpio(buf: &[u8]) -> Result<Vec<u8>> {
...
+
+    let stdout = io::stdout();
+    let mut out = stdout.lock();
+    serde_json::to_writer_pretty(&mut out, &inspect_out)
+        .context("failed to serialize ISO metadata")?;
+    out.write_all(b"\n").context("failed to write newline")?;
+    Ok(())
+}
+
+fn collect_iso_records(
```
jlebon, line 676: I may fold this into the `iso9660` module in a future patch.

cgwalters (commented):
Didn't verify anything in depth, but LGTM.  

One thing I think would help a lot here to increase confidence is to execute this as part of the coreos-assembler builds, and hence require the property that if we ship an ISO, this command can parse it.

jlebon:
> One thing I think would help a lot here to increase confidence is to execute this as part of the coreos-assembler builds, and hence require the property that if we ship an ISO, this command can parse it.

The CI test added here operates on the live ISO output by cosa. But yeah, once we have the minimal ISO work, we'll implicitly test the parser (and more!) at `cosa buildextend-live` time too.

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+            }
+        }
+        Err(anyhow!("directory {} has no record for {}", dir.name, name))
+    }
+
+    /// Returns a reader for a file record.
+    pub fn read_file(&mut self, file: &File) -> Result<impl Read + '_> {
+        seek_to_sector(&mut self.file, file.address)
+            .with_context(|| format!("seeking to file {}", file.name))?;
+        Ok(LimitReader::new(&self.file, file.length as u64, None))
```
bgilbert, line 140: I just noticed that `Read.take()` exists, so we could use it and revert da13fe03ad53aab1838f31dbd9a6ab34757039bf.

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+}
+
+/// Reads the directory record at cursor and advances to the next one.
+fn get_next_directory_record(buf: &mut Bytes) -> Result<Option<DirectoryRecord>> {
+    let len = buf.get_u8() as usize;
+    if len == 0 {
+        return Ok(None);
+    }
+
+    let address = eat(buf, 1).get_u32_le();
```
bgilbert, line 295: A corrupt image that puts a partial entry at the end of the directory will cause one of these calls to panic.  Do we want to check for that case?

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+        TYPE_BOOT => Some(VolumeDescriptor::Boot(parse_boot_descriptor(buf)?)),
+        TYPE_PRIMARY => Some(VolumeDescriptor::Primary(parse_primary_descriptor(buf)?)),
+        TYPE_SUPPLEMENTARY => Some(VolumeDescriptor::Supplementary),
+        TYPE_TERMINATOR => None,
+        t => Some(VolumeDescriptor::Unknown { type_id: t }),
+    })
+}
+
+/// Parses boot descriptor at cursor.
+fn parse_boot_descriptor(buf: &mut Bytes) -> Result<BootVolumeDescriptor> {
```
bgilbert, line 233: nit: The parse functions could be associated functions on the respective types.

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+    verify_descriptor_header(buf).context("parsing boot descriptor")?;
+    Ok(BootVolumeDescriptor {
+        boot_system_id: parse_iso9660_string(buf, 32)?,
```
bgilbert, line 236: Do we want to `.trim_end_matches(" ")` these?

bgilbert (commented):
This is great!

src/iso9660.rs
bgilbert: Also, should we wrap this in a BufReader?

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+            .read_exact(&mut buf)
+            .with_context(|| format!("reading directory {}", dir.name))?;
+
+        Ok(IsoFsIterator {
+            dir: Bytes::from(buf),
+        })
+    }
+
+    /// Returns the record for a specific path.
+    pub fn get_path(&mut self, path: &str) -> Result<DirectoryRecord> {
```
bgilbert, line 87: Would it make sense to have a `get_file` helper method that rejects directories?

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+    if flags & 2 > 0 {
+        Ok(Some(DirectoryRecord::Directory(Directory {
+            name,
+            address,
+            length,
+        })))
+    } else {
+        Ok(Some(DirectoryRecord::File(File {
+            name,
+            address,
```
bgilbert, line 313: If we intend the rest of the codebase to access this field directly (e.g. for ISO embed), it seems more useful to record the byte offset here.

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    if flags & 2 > 0 {
+        Ok(Some(DirectoryRecord::Directory(Directory {
+            name,
+            address,
+            length,
+        })))
+    } else {
+        Ok(Some(DirectoryRecord::File(File {
+            name,
```
bgilbert, line 312: I assume we don't plan to use the file version number; should we drop it from the name for convenience?

src/live.rs
```diff
@@ -647,6 +648,53 @@ fn extract_cpio(buf: &[u8]) -> Result<Vec<u8>> {
...
 }
 
+#[derive(Serialize)]
+struct IsoInspectOutput {
+    header: IsoFs,
+    records: Vec<String>,
+}
+
+pub fn iso_inspect(config: &IsoInspectConfig) -> Result<()> {
+    let mut iso = IsoFs::new(&config.input)?;
```
bgilbert, line 658: For consistency, maybe we should use `open_live_iso(&config.input, None)` here.  We'll need an `IsoFs::for_file()` anyway when refactoring the other commands in this file.

bgilbert:
https://github.com/coreos/coreos-installer/pull/622 has another API consumer.

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+fn parse_iso9660_path(buf: &mut Bytes, len: usize) -> Result<String> {
+    if len == 1 && (buf[0] == 0 || buf[0] == 1) {
+        let c = buf.get_u8();
+        if c == 0 {
+            Ok(".".into())
+        } else {
+            Ok("..".into())
+        }
+    } else {
+        parse_iso9660_string(buf, len)
```
bgilbert, line 330: ISO 9660 doesn't allow `/` characters here (they're not in the D-string set), but we are, which potentially allows directory traversals.

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+            bail!("invalid string name {:?}", bytes);
+        }
+    }
+    Ok(s)
+}
+
+fn seek_to_sector(f: &mut fs::File, sector: u32) -> Result<()> {
+    let offset = (ISO9660_SECTOR_SIZE as u64)
+        .checked_mul(sector as u64)
+        .ok_or_else(|| anyhow!("sector offset mult overflow"))?;
```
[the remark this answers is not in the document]
jlebon: Ended dropping that function entirely.

src/iso9660.rs
```diff
@@ -0,0 +1,364 @@
...
+        get_next_directory_record(&mut self.dir)
+            .context("reading next record")
+            .transpose()
+    }
+}
+
+/// Reads the directory record at cursor and advances to the next one.
+fn get_next_directory_record(buf: &mut Bytes) -> Result<Option<DirectoryRecord>> {
+    let len = buf.get_u8() as usize;
+    if len == 0 {
```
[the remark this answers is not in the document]
jlebon: Good catch, fixed!

src/iso9660.rs
jlebon: Let's tackle the `Read.take()` thing separately.

Added `BufReader` wrapper.

src/iso9660.rs
jlebon: Good catch, fixed!

src/iso9660.rs
jlebon: Changed this, but might change it back when I get farther on miniso work. The miniso table will be serialized as LBA, and while it also needs to seek I'd rather have it multiply than divide.

src/live.rs
jlebon: Did this now as a new iterator accessible via `IsoFs::walk()`. The function is a bit gnarlier than I would've liked, but the interface for callers is much nicer.

jlebon:
Updated for comments! :arrow_up:

bgilbert (commented):
Some of the below gets into API Opinions.  Feel free to defer things if you just want this PR to land.

src/iso9660.rs
```diff
@@ -0,0 +1,517 @@
...
+            break;
+        } else {
+            bail!("invalid string name {:?}", bytes);
+        }
+    }
+    if matches!(kind, IsoString::StrA | IsoString::StrD) {
+        s.truncate(s.trim_end_matches(' ').len());
+    } else {
+        // IsoString::File -- let's make lowercase to be less screamy
+        s.make_ascii_lowercase();
```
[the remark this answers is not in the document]
jlebon: Hmm, this is what the kernel itself does though: https://github.com/torvalds/linux/blob/ddf21bd8ab984ccaa924f090fc7f515bb6d51414/fs/isofs/dir.c#L28-L29. Feels like we should just match that behaviour?

src/iso9660.rs
```diff
@@ -0,0 +1,517 @@
...
+enum IsoString {
+    StrA,
+    StrD,
+    File,
+}
+
+/// Reads an ISO9660 string.
+fn parse_iso9660_string(buf: &mut Bytes, len: usize, kind: IsoString) -> Result<String> {
+    const FILE_CHARS: [u8; 17] = *b"!\"%&'()*+,-.:<=>?";
+    const A_CHARS: [u8; 2] = *b";/";
```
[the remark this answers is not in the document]
jlebon: `A_CHARS` are the chars that are permitted in A-chars on top of D-chars and `FILE_CHARS`.

I'm using the kernel here as reference. There, it basically allows anything, apart from `;` and `/`. Let's just match the kernel and support the same (although non-compliant) filenames?

src/iso9660.rs
```diff
@@ -0,0 +1,517 @@
...
+        #[allow(clippy::if_same_then_else)] // I find it easier to follow this way
+        if byte.is_ascii_alphabetic() || byte.is_ascii_digit() || *byte == b'_' || *byte == b' ' {
+            s.push(char::from(*byte));
+        } else if FILE_CHARS.contains(byte) && matches!(kind, IsoString::File | IsoString::StrA) {
+            s.push(char::from(*byte));
+        } else if A_CHARS.contains(byte) && matches!(kind, IsoString::StrA) {
+            s.push(char::from(*byte));
+        } else if A_CHARS.contains(byte) && matches!(kind, IsoString::File) {
+            s.push('.'); // this matches what the kernel does
+        } else if *byte == 0 {
```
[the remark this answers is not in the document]
jlebon: Matching the kernel here too. :)

src/iso9660.rs
```diff
@@ -0,0 +1,517 @@
...
+    }
+
+    pub fn get_root_directory(&self) -> Result<Directory> {
+        let primary = self
+            .get_primary_volume_descriptor()
+            .context("getting root directory")?;
+        Ok(primary.root.clone())
+    }
+
+    pub fn walk(&mut self) -> Result<IsoFsWalkIterator> {
```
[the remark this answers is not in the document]
jlebon: The only use cases I have for this right now want to walk the whole filesystem, so it seems nicer to just leave the simpler signature. We can tweak it in the future if that's not the case.

src/iso9660.rs
```diff
@@ -0,0 +1,517 @@
...
+    /// Returns the record for a specific file.
+    pub fn get_file(&mut self, path: &str) -> Result<File> {
+        match self.get_path(path)? {
+            DirectoryRecord::Directory(_) => Err(anyhow!("path {} is a directory", path)),
+            DirectoryRecord::File(f) => Ok(f),
+        }
+    }
+
+    /// Returns the record for a specific name in a directory if it exists.
+    pub fn get_dir_record(
```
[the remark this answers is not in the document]
jlebon: Offhand, feels like if a caller is going to try some specific paths with optional `ENOENT`s allowed, there aren't going to be that many it'll try. I think the potential number of APIs in this area could get combinatorial, so let's defer for now and revisit as we gain more experience with this?

src/iso9660.rs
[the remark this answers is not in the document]
bgilbert: Yeah, but mostly it doesn't, because the filesystem usually has Rock Ridge or Joliet extensions.  I'm not saying the behavior here is wrong, I'm saying it seems more likely to cause c-i bugs when someone doesn't realize that the filenames here are different than the ones they see in `ls`.  In the old Windows days, if you saw a list of uppercase shortnames, you knew you were in 8.3 land.

src/iso9660.rs
```diff
@@ -0,0 +1,513 @@
...
+                }
+            }
+        }
+        Ok(None)
+    }
+}
+
+/// Reads the directory record at cursor and advances to the next one.
+fn get_next_directory_record(buf: &mut Bytes, length: u32) -> Result<Option<DirectoryRecord>> {
+    if buf.remaining() == 0 {
```
bgilbert, line 399: Clippy might say "use `buf.has_remaining()`" if clippy knew about `bytes`.  :stuck_out_tongue:

src/iso9660.rs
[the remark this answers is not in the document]
bgilbert: Sure.  Let's add a comment to that effect, though.

src/iso9660.rs
```diff
@@ -0,0 +1,517 @@
...
+/// Reads the directory record at cursor and advances to the next one.
+fn get_next_directory_record(buf: &mut Bytes, length: u32) -> Result<Option<DirectoryRecord>> {
+    let mut remaining = buf.remaining();
+    if remaining == 0 {
+        return Ok(None);
+    }
+
+    let mut len = buf.get_u8() as usize;
+    remaining -= 1;
+    if len == 0 && remaining > 0 {
```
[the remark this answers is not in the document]
bgilbert: This was marked resolved, but the test is still there.  In the `len == 0`, `buf.remaining() == 0` case, I think we'll currently panic computing `len - 1` in debug builds.

src/iso9660.rs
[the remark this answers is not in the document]
bgilbert: Discussed this OOB.  Let's land as-is for now, and circle back in a followup.

bgilbert (approved):
:tada:

src/iso9660.rs
bgilbert: Turns out that `take()` consumes the `File` and `File.try_clone()` returns a clone that shares the file offset with the original, so I'm thinking the `take()` change isn't worth it.

jlebon: I thought this too at first, but it can work if we wrap the `LimitReader` itself, but it's kinda awkward since they essentially do almost the same thing.

bgilbert: Ahh, yeah, good call.

PR #621, opened 2021-09-13
iso: add `coreos-installer iso extract pxe`

jlebon:
This commands allows users to easily extract the PXE artifacts from
CoreOS live ISOs without having to mount them, which may require
additional privileges.

There are other CLI tools which can do this of course, but having it
integrated makes it much easier and it's trivial to support with the new
ISO9660 parser.

It also establishes the `extract` subcommand, to which we'll add a
`miniso` command for the minimal ISO work.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -72,6 +72,8 @@ pub enum IsoCmd {
     // for testing and debugging purposes only
     #[structopt(setting(AppSettings::Hidden))]
     Inspect(IsoInspectConfig),
+    /// Commands to extract files from a CoreOS live ISO image
+    Extract(IsoExtractCmd),
 }
 
 #[derive(Debug, StructOpt)]
@@ -92,6 +94,12 @@ pub enum IsoKargsCmd {
     Reset(IsoKargsResetConfig),
     /// Show kernel args from an ISO image
     Show(IsoKargsShowConfig),
+}
+
+#[derive(Debug, StructOpt)]
+pub enum IsoExtractCmd {
+    /// Extract PXE files from an ISO image
+    Pxe(IsoExtractPxeConfig),
 }
 
 #[derive(Debug, StructOpt)]
@@ -389,6 +397,16 @@ pub struct IsoInspectConfig {
     /// ISO image
     #[structopt(value_name = "ISO")]
     pub input: String,
+}
+
+#[derive(Debug, StructOpt)]
+pub struct IsoExtractPxeConfig {
+    /// ISO image
+    #[structopt(value_name = "ISO")]
+    pub input: String,
+    /// Output directory
+    #[structopt(short, long, value_name = "PATH", default_value = ".")]
+    pub output_dir: String,
 }
 
 #[derive(Debug, StructOpt)]
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -68,6 +68,10 @@ pub enum IsoCmd {
     Ignition(IsoIgnitionCmd),
     /// Modify kernel args in a CoreOS live ISO image
     Kargs(IsoKargsCmd),
+    /// Inspect the CoreOS live ISO image
+    // for testing and debugging purposes only
+    #[structopt(setting(AppSettings::Hidden))]
+    Inspect(IsoInspectConfig),
 }
 
 #[derive(Debug, StructOpt)]
@@ -375,6 +379,13 @@ pub struct IsoKargsShowConfig {
     /// Show ISO header (for debugging/testing only)
     #[structopt(long, hidden = true)]
     pub header: bool,
+    /// ISO image
+    #[structopt(value_name = "ISO")]
+    pub input: String,
+}
+
+#[derive(Debug, StructOpt)]
+pub struct IsoInspectConfig {
     /// ISO image
     #[structopt(value_name = "ISO")]
     pub input: String,
--- a/src/live.rs
+++ b/src/live.rs
@@ -28,6 +28,7 @@ use std::path::Path;
 use crate::cmdline::*;
 use crate::install::*;
 use crate::io::*;
+use crate::iso9660::IsoFs;
 
 const FILENAME: &str = "config.ign";
 const COREOS_IGNITION_HEADER_MAGIC: &[u8] = b"coreiso+";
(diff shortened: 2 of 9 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-installer/pull/620

src/cmdline.rs
```diff
@@ -391,6 +399,16 @@ pub struct IsoInspectConfig {
     pub input: String,
 }
 
+#[derive(Debug, StructOpt)]
+pub struct IsoExtractPxeConfig {
+    /// ISO image
+    #[structopt(value_name = "ISO")]
+    pub input: String,
+    /// Output directory
+    #[structopt(short, long)]
```
bgilbert, line 408: `value_name = "path"`?

src/live.rs
```diff
@@ -695,6 +696,37 @@ fn collect_iso_records(
...
+            iso9660::DirectoryRecord::Directory(_) => continue,
+            iso9660::DirectoryRecord::File(file) => {
+                let basename = file.name.trim_end_matches(";1").trim_end_matches('.');
+                let path = Path::new(outdir).join(basename);
+                let mut outf = OpenOptions::new()
+                    .write(true)
+                    .create_new(true)
+                    .open(&path)
+                    .with_context(|| format!("opening {}", path.display()))?;
+                println!("Extracting {} to {}", basename, path.display());
```
bgilbert, line 722: `coreos-installer download` just prints the filenames without explanatory text, so that scripts can consume them easily.

src/live.rs
```diff
@@ -695,6 +696,37 @@ fn collect_iso_records(
...
+    };
+
+    let outdir = config.output_dir.as_deref().unwrap_or(".");
+    std::fs::create_dir_all(outdir)?;
+
+    for record in iso.list_dir(&pxeboot)? {
+        match record? {
+            iso9660::DirectoryRecord::Directory(_) => continue,
+            iso9660::DirectoryRecord::File(file) => {
+                let basename = file.name.trim_end_matches(";1").trim_end_matches('.');
```
bgilbert, line 715: The results are in screaming case.  Seems reasonable to lower-case them?

src/live.rs
```diff
@@ -695,6 +696,37 @@ fn collect_iso_records(
...
+pub fn iso_extract_pxe(config: &IsoExtractPxeConfig) -> Result<()> {
+    let mut iso = IsoFs::new(&config.input)?;
+    let pxeboot = match iso.get_path(COREOS_ISO_PXEBOOT_DIR)? {
+        iso9660::DirectoryRecord::Directory(d) => d,
+        iso9660::DirectoryRecord::File(_) => {
+            bail!("expected dir {}, but found file", COREOS_ISO_PXEBOOT_DIR)
+        }
+    };
+
+    let outdir = config.output_dir.as_deref().unwrap_or(".");
```
bgilbert, line 708: For clarity of help text, let's set a `default = "."` structopt property and avoid the unwrap here.

src/live.rs
```diff
@@ -695,6 +696,37 @@ fn collect_iso_records(
...
+    std::fs::create_dir_all(outdir)?;
+
+    for record in iso.list_dir(&pxeboot)? {
+        match record? {
+            iso9660::DirectoryRecord::Directory(_) => continue,
+            iso9660::DirectoryRecord::File(file) => {
+                let basename = file.name.trim_end_matches(";1").trim_end_matches('.');
+                let path = Path::new(outdir).join(basename);
```
bgilbert, line 716: A `file.name` of `../INUZ` will get through validation and perform a directory traversal.  The consequences seem limited because of the `create_new()`, but it'd still be good to fix.

src/live.rs
```diff
@@ -695,6 +696,37 @@ fn collect_iso_records(
...
+                let mut outf = OpenOptions::new()
+                    .write(true)
+                    .create_new(true)
+                    .open(&path)
+                    .with_context(|| format!("opening {}", path.display()))?;
+                println!("Extracting {} to {}", basename, path.display());
+                copy(&mut iso.read_file(&file)?, &mut outf)?;
```
bgilbert, line 723: In my tests this is using 8 KiB chunks.  It probably wants a `BufReader`/`BufWriter`.

bgilbert (commented):
Nice!

bgilbert:
I'm a little uncertain about just dumping out the contents of the ISO directory verbatim.  Alternatives include providing explicit destination arguments (`-- kernel kernel-file`) and prepending the filename of the ISO image so the version information is preserved.  Thoughts?

src/live.rs
bgilbert: https://github.com/coreos/coreos-installer/pull/620/files#r708383059

jlebon:
(Haven't addressed comments here yet, only rebased on top of the latest version of https://github.com/coreos/coreos-installer/pull/620.)

src/cmdline.rs
```diff
@@ -72,6 +72,8 @@ pub enum IsoCmd {
     // for testing and debugging purposes only
     #[structopt(setting(AppSettings::Hidden))]
     Inspect(IsoInspectConfig),
+    /// Commands to extract files from a CoreOS live ISO image
```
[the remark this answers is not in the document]
jlebon: Heh, actually took example from other "subparent" commands:

>    /// Commands to manage a CoreOS live ISO image

>    /// Commands to manage a CoreOS live PXE image

jlebon:
> I'm a little uncertain about just dumping out the contents of the ISO directory verbatim. Alternatives include providing explicit destination arguments (`-- kernel kernel-file`) and prepending the filename of the ISO image so the version information is preserved. Thoughts?

I like prepending the filenames, so did that now!

Hmm, actually we could use the volume ID to restore the exact original artifact names, which would be cool. Although, I guess overall it's still nicer to have it match the ISO filename itself so they're listed near each other.

src/live.rs
```diff
@@ -674,6 +675,42 @@ pub fn iso_inspect(config: &IsoInspectConfig) -> Result<()> {
...
+                let path = Path::new(&config.output_dir).join(&filename);
+                let mut outf = OpenOptions::new()
+                    .write(true)
+                    .create_new(true)
+                    .open(&path)
+                    .with_context(|| format!("opening {}", path.display()))?;
+                let mut bufw = BufWriter::with_capacity(BUFFER_SIZE, &mut outf);
+                println!("{}", path.display());
+                copy(&mut iso.read_file(&file)?, &mut bufw)?;
+            }
```
bgilbert, line 708: `BufWriter` recommends a manual `flush()` before drop.

src/cmdline.rs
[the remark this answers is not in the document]
bgilbert: Whoops, you're right, we're inconsistent and I didn't look carefully enough.

bgilbert (approved):

PR #622, opened 2021-09-14
live: use ISO9660 parser to find Ignition/kargs embed areas

bgilbert:
Drop support for the `coreiso+` header, since it's always possible to use the ISO9660 parser to find the Ignition embed area.  For kargs, use the new `kargs.json` from https://github.com/coreos/coreos-assembler/pull/2429 if we have it, and fall back on the `coreKarg` header otherwise.


bgilbert:
Added kargs tests for all combinations of headers.

bgilbert:
Ready for review.

bgilbert:
Added a test fixture so we can continue testing the legacy kargs header after cosa stops emitting it.  Also added some unit tests.

tests/iso-kargs.sh
```diff
@@ -21,94 +21,113 @@ cd "${tmpd}"
...
+echo "============== Both headers =============="
+run_tests
+
+echo "============== Only JSON =============="
+dd if=/dev/zero of="${iso}" seek=32672 count=8 bs=1 conv=notrunc status=none
+run_tests
+
+echo "============== Only System Area =============="
+echo -n "coreKarg" | dd of="${iso}" seek=32672 bs=1 conv=notrunc status=none
+echo -n "X" | dd of="${iso}" seek=$(grep --byte-offset --only-matching --text 'KARGS.JSO;1' "${iso}" | cut -f1 -d:) bs=1 conv=notrunc status=none
```
jlebon: So, this is overwriting the name field in the directory record to be `XARGS.JSON;1`? Feels like it's worth a comment.

src/live.rs
```diff
@@ -456,9 +457,76 @@ struct KargEmbedAreas {
...
+
+        // sanity-check size against a reasonable limit
+        if info.size > COREOS_KARG_EMBED_AREA_MAX_SIZE {
+            bail!(
+                "karg embed area size larger than {} (found {})",
+                COREOS_KARG_EMBED_AREA_MAX_SIZE,
+                info.size
+            );
+        }
+        if info.default.len() > info.size {
```
jlebon, line 498: Hmm, so then do we even need a `size` field in the JSON file given that the information is already provided via the string length itself?

src/live.rs
```diff
@@ -776,6 +776,69 @@ fn copy_file_from_iso(iso: &mut IsoFs, file: &iso9660::File, output_path: &Path)
...
+        iso_file
+    }
+
+    #[test]
+    fn test_ignition_embed_area() {
+        let mut iso_file = open_iso_file();
+        // normal read
+        let mut iso = IsoFs::from_file(iso_file.try_clone().unwrap()).unwrap();
+        let region = ignition_embed_area(&mut iso).unwrap();
+        assert_eq!(region.offset, 102400);
```
jlebon, line 798: Minor: would be nice if these offset numbers were actually calculated so it's easier to tell it's correct. Shouldn't be too hard with the IsoFs APIs.

jlebon (commented):
This appears to be missing the script that actually generated the `embed-areas.iso.xz` file.

bgilbert:
> This appears to be missing the script that actually generated the `embed-areas.iso.xz` file.

The script is a hacked-up version of `cmd-buildextend-live`, running inside cosa.  I didn't commit it because I don't expect that we'll ever refresh that fixture; it's a point-in-time snapshot of how we currently build the live ISO.

src/live.rs
bgilbert: Unlike in the old `.cmdline` file, the `default` string is not padded to the size of the embed area, so the sizes are in fact different.

src/live.rs
bgilbert: Sure, but then we're partially comparing the code itself.  It seemed better to manually generate the ground truth.

jlebon:
> > This appears to be missing the script that actually generated the `embed-areas.iso.xz` file.
> 
> The script is a hacked-up version of `cmd-buildextend-live`, running inside cosa. I didn't commit it because I don't expect that we'll ever refresh that fixture; it's a point-in-time snapshot of how we currently build the live ISO.

Right OK, but OTOH having binary files in the repo without any practical way to reproduce them isn't great. It makes it almost impossible to tell if it was generated correctly and to tweak things if needed, even if we don't envision needing it right now. That coreos-installer handles it fine is good, but it's kind of backwards testing.

So, not sure. I think I'm OK with it, and I don't think it's worth trying to convert your hacked-up `cmd-buildextend-live` to shell or something, but it feels odd to me.

bgilbert:
Yeah, I agree it's awkward.  It seems like an improvement over the previous state, where we needed to call out to an entire other subsystem to generate test data for us.

To the extent we change how images are generated in the future, I think it makes sense to continue testing compatibility with _this_ image.  I've added a datestamp to the filename to try to make it clearer that this is a point-in-time snapshot.

To the extent that the image itself may have been generated incorrectly... I don't know.  Should I add a file here containing a cosa patch?

jlebon (approved):
Yup, definitely agreed it's a nice improvement overall. I like being able to unit test at the function level. Let's roll with it, and if for whatever reason we do need to change the fixture, that might be a sign we should formalize it a bit more.

PR #627, opened 2021-09-19
Some ISO 9660 refactoring

bgilbert:
1. Remove from `LimitReader` the ability to return `Ok(0)` when the limit is reached.  Use `Read.take()` instead.
1. Drop the `IsoFs` `get_{dir,file}` wrappers in favor of new `try_into_{dir,file}` methods on `DirectoryRecord`.
1. Add `IsoFs.try_get_path()` method returning `Result<Option<DirectoryRecord>>`, for callers that can handle a missing path.

### DIFF
--- a/src/iso9660.rs
+++ b/src/iso9660.rs
@@ -402,51 +402,52 @@ impl<'a> IsoFsWalkIterator<'a> {
 
 /// Reads the directory record at cursor and advances to the next one.
 fn get_next_directory_record(buf: &mut Bytes, length: u32) -> Result<Option<DirectoryRecord>> {
-    if !buf.has_remaining() {
-        return Ok(None);
-    }
-
-    let mut len = buf.get_u8() as usize;
-    if len == 0 {
-        let jump = {
-            // calculate where we are we in the directory
-            let pos = length as usize - buf.remaining();
-            // get distance to next 2k-aligned address
-            ((pos + ISO9660_SECTOR_SIZE) & !(ISO9660_SECTOR_SIZE - 1)) - pos
-        };
-        if jump >= buf.remaining() {
+    loop {
+        if !buf.has_remaining() {
             return Ok(None);
         }
-        buf.advance(jump);
-        len = buf.get_u8() as usize;
-    }
 
-    // + 1 because len includes the length of the length byte itself, which we already read
-    if buf.remaining() + 1 < len {
-        bail!("incomplete directory record; corrupt ISO?");
-    }
+        let len = buf.get_u8() as usize;
+        if len == 0 {
+            let jump = {
+                // calculate where we are we in the directory
+                let pos = length as usize - buf.remaining();
+                // get distance to next 2k-aligned address
+                ((pos + ISO9660_SECTOR_SIZE) & !(ISO9660_SECTOR_SIZE - 1)) - pos
+            };
+            if jump >= buf.remaining() {
+                return Ok(None);
+            }
+            buf.advance(jump);
+            continue;
+        } else if len > buf.remaining() + 1 {
+            // + 1 because len includes the length of the length byte
+            // itself, which we already read
+            bail!("incomplete directory record; corrupt ISO?");
+        }
 
-    let address = (eat(buf, 1).get_u32_le() as u64) * (ISO9660_SECTOR_SIZE as u64);
-    let length = eat(buf, 4).get_u32_le();
-    let flags = eat(buf, 25 - 14).get_u8();
-    let name_length = eat(buf, 32 - 26).get_u8() as usize;
-    let name = parse_iso9660_path(buf, name_length).context("parsing record name")?;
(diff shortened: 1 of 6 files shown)


bgilbert:
In hindsight, it might have been less effort to implement Rock Ridge.

src/iso9660.rs
```diff
@@ -107,31 +107,41 @@ impl IsoFs {
...
                 "no record for {} in directory {}",
                 filename,
                 parent_dir.display()
-            )
+            )))
         })
     }
 
+    /// Returns the record for a specific path if it exists.
+    pub fn try_get_path(&mut self, path: &str) -> Result<Option<DirectoryRecord>> {
```
jlebon: Hmm, could we instead just make `NotFound` public and clients of `get_path` can just look for it if needed?

src/iso9660.rs
```diff
@@ -439,38 +440,41 @@ fn get_next_directory_record(buf: &mut Bytes, length: u32) -> Result<Option<Dire
...
-            Ok("..".into())
+            Ok(Some(DirectoryRecord::File(File {
+                name,
+                address,
+                length,
+            })))
         }
     } else {
-        parse_iso9660_string(buf, len, IsoString::File)
+        get_next_directory_record(buf, length, is_root)
```
jlebon, line 467: This recursive call is kind of awkward. I know it's subject to TCO, but... I had the same feeling about the jump to the next 2k alignment at the beginning of the function where we hackily just advance the cursor and pretend to start from scratch. Thinking it'd be cleaner at this point to just move to a loop where the start of each iteration always assumes we're at the start of a directory record.

jlebon (commented):
First 6 commits makes sense to me overall. Can you split out the last two commits in a separate PR?

jlebon:
Notably, I'm not a fan of the added complexity surrounding 8.3 path handling in the last two commits. I agree it'd probably be better to just support RR since we know the ISOs we care about do emit them.

bgilbert:
Yeah, makes sense.  I do think there's a kernel of a good idea in the last commit, though, which is normalizing paths explicitly and separately before walking them in `get_path()`.

src/iso9660.rs
bgilbert: Yup, good call.

bgilbert:
Updated for comments and moved the last two commits to #628.

bgilbert:
I've extracted the canonicalization part of #628 and added it in a new commit here.

src/iso9660.rs
```diff
@@ -522,3 +519,48 @@ fn eat(buf: &mut Bytes, n: usize) -> &mut Bytes {
     buf.advance(n);
     buf
 }
+
+/// Parse path into a PathBuf with zero or more Normal components.  Convert
+/// path to relative and resolve all "." and ".." components.
+fn canonicalize_path(s: &str) -> Result<PathBuf> {
```
jlebon, line 525: Should we just return a `Vec<&OsStr>` instead and then we can drop the `else` branch with the `panic!()` in `get_path`?

src/iso9660.rs
```diff
@@ -522,3 +519,48 @@ fn eat(buf: &mut Bytes, n: usize) -> &mut Bytes {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_canonicalize_path() {
+        // basic
+        assert_eq!(canonicalize_path("z").unwrap().to_str().unwrap(), "z");
+        // absolute path with . and ..
+        assert_eq!(
+            canonicalize_path("/a/./../b").unwrap().to_str().unwrap(),
```
jlebon, line 553: Looking at these paths... I wonder if we really need to support them. Feels like it's OK to just require callers to provide canonical paths? Or did you have a use case in mind that would leverage this?

bgilbert: I don't have a use case, but it felt awkward that `get_path()` was trying to pick-n-choose certain behaviors of `Path` and work around others.  This way we've defined and tested clear semantics.  We could restrict the semantics further, but the implementation is basically free, so I wonder whether that's worth doing.

jlebon: Ack, SGTM!

src/iso9660.rs
bgilbert: Good call.

bgilbert:
Updated!

src/iso9660.rs
```diff
@@ -512,3 +504,42 @@ fn eat(buf: &mut Bytes, n: usize) -> &mut Bytes {
...
+    use std::path::Component::*;
+    let mut ret = Vec::new();
+    for c in Path::new(s).components() {
+        match c {
+            Prefix(_) | RootDir | CurDir => (),
+            ParentDir => {
+                ret.pop();
+            }
+            Normal(c) => {
+                ret.push(c.to_str().unwrap()); // `path` is &str
```
jlebon, line 521:
```suggestion
                ret.push(c.to_str().unwrap()); // `s` is &str
```

PR #628, opened 2021-09-20
RFC: iso9660: add `IsoPath` type

bgilbert:
Add and consistently use an `IsoPath` type that canonicalizes paths and encapsulates the filename restrictions we're applying for interchange level 1: monocase lowercase filenames truncated to 8.3, with some special characters forbidden.  This makes it clear when we're working with ISO 9660 pathnames vs. regular ones, and allows callers to not have to think about the conversion rules.

Broken out of https://github.com/coreos/coreos-installer/pull/627, and requires it.  See discussion in that PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -186,6 +186,7 @@ dependencies = [
  "serde_json",
  "structopt",
  "tempfile",
+ "thiserror",
  "url",
  "uuid",
  "walkdir",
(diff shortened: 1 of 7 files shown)


bgilbert:
Discussed with @jlebon OOB.  This doesn't handle the full generality of the problem, since `genisoimage` will create shortnames with serial numbers if two records' shortnames would collide.  Also this abstraction seems like overkill for now.  Closing this out.

PR #629, opened 2021-09-21
iso9660: add unit tests with ISO9660 test fixture

bgilbert:
Test what we can in `cargo test` rather than in a separate test script.


jlebon (approved):
Nice work! The unit test-level coverage makes this definitely worth it.

Issue #516, opened 2021-04-06
Switch from `lsblk --pairs` to `lsblk --json`


jlebon:
`lsblk` supports `--json`, which would be much nicer to work with using serde than to parse the `--pairs` output using regex.

lucab:
I think we already had this but we reverted in https://github.com/coreos/coreos-installer/pull/95 because CentOS 7 didn't support that. Did something change in the meanwhile and we are reinstating this?

jlebon:
> I think we already had this but we reverted in #95 because CentOS 7 didn't support that. Did something change in the meanwhile and we are reinstating this?

Thanks for bringing this up! I wasn't aware of this.

We discussed this during open discussion and since we haven't had CentOS/RHEL7 at all on our radar, it's possible that compatibility for it broke a long time ago already. The proposed path forward is:
- Try to build latest coreos-installer on CentOS/RHEL7.
    - If it doesn't work and would require a lot of effort to get it working, then let's clearly declare it as unsupported and point people towards using it as a container instead.
    - If it still works well, or only requires minor adaptations, then let's keep supporting it, including adding it in CI, and holding off on enhancements like this one which would break compatibility. We can then discuss a deprecation period after which we point people towards using it as a container instead.

At least EPEL it seems has been receiving the latest Rust updates together with Fedora, so it shouldn't be limited by the toolchain.

nikita-dubrovskii:
So, bad news:
- `coreos-installer` could be built on CentOS7 (using `rust` from `epel`)
- `coreos-installer` still works on CentOS7 
- CentOS7 still has old `lsblk` with no `--json` support

PR #631, opened 2021-09-22
src/* : use lsblk --json output format

sohankunkerkar:
Fixes: https://github.com/coreos/coreos-installer/issues/516


src/blockdev.rs
```diff
@@ -42,6 +42,42 @@ use crate::util::*;
...
+    pub fstype: Option<String>,
+    #[serde(rename = "type")]
+    pub blktype: Option<String>,
+    pub mountpoint: Option<String>,
+    pub uuid: Option<String>,
+    pub children: Option<Vec<Device>>,
+}
+
+impl Device {
+    pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<Device>> {
```
jlebon, line 63: So I think here a really nice thing about switching to JSON is that we capture the hierarchy of the block devices. In this case, subdevices of the main device we're querying will be in `children`.

So here I think we want to output a `Result<Device>` instead. The `lsblk` output supports outputting multiple top-level devices, but in our uses of `lsblk` we're always really just interested in a single top-level device (and possibly its children if `with_deps` is enabled). So this function could sanity-check that only one device matched, and return that one.

src/blockdev.rs
```diff
@@ -105,37 +141,58 @@ impl Disk {
...
+        for deviceinfo in deviceinfos {
+            let mut partition = self.compute_partition(&deviceinfo, with_holders)?;
+            result.append(&mut partition);
+            if let Some(children) = deviceinfo.children.as_ref() {
+                if !children.is_empty() {
+                    for child in children {
+                        let mut childpartitions = self.compute_partition(child, with_holders)?;
+                        result.append(&mut childpartitions)
+                    }
+                }
```
jlebon, line 195: Hmm, right OK you had to do some refactoring here to deal with children not being in the main iteration anymore.

Though I don't think this flow is functionally equivalent, because we're only going one level down. I think we want a recursive model here instead? So e.g. you have `get_partitions` call to `get_partitions_recurse` with the top-level `Device`, and then that function calls itself for each child in the device, aggregating partitions in the vector as it goes. Then we return that vector. Does that make sense?

src/blockdev.rs
jlebon: I think we should also rename `with_deps` to `with_children`, since that's the name in the structure we return.

Though hmm, kinda tempting to drop the `with_deps` argument altogether since we'd always return one device now unlike before. `lsblk` will be doing more work than it needs to in some cases, but meh... feels like it's worth the cleaner API we get. @bgilbert might have a different opinion.

src/blockdev.rs
```diff
@@ -42,6 +42,47 @@ use crate::util::*;
...
+        if !with_children {
+            cmd.arg("--nodeps");
+        }
+        let output = cmd_output(&mut cmd)?;
+        let devs: DevicesOutput = serde_json::from_str(&output)?;
+        let devinfo = devs
+            .blockdevices
+            .into_iter()
+            .next()
+            .ok_or_else(|| anyhow!("failed to get device information"))?;
```
jlebon, line 81: This should verify that there is only one dev and return that dev. Otherwise error out.

src/blockdev.rs
```diff
@@ -506,11 +563,12 @@ impl Mount {
...
-        devinfo
-            .get("UUID")
-            .map(String::from)
-            .with_context(|| format!("filesystem {} has no UUID", self.device))
+        let devinfo = Device::lsblk(Path::new(&self.device), false)?;
+        let uuid = devinfo[0]
+            .uuid
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("missing uuid in {}", devinfo[0].name))?;
+        Ok(uuid.to_string())
```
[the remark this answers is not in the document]
jlebon: This is still cloning.

src/blockdev.rs
```diff
@@ -105,37 +146,72 @@ impl Disk {
...
+        with_holders: bool,
+    ) -> Result<Vec<Partition>> {
+        let mut result: Vec<Partition> = Vec::new();
+        if let Some(children) = deviceinfo.children.as_ref() {
+            if !children.is_empty() {
+                for child in children {
+                    let mut childpartitions = self.compute_partition(child, with_holders)?;
+                    result.append(&mut childpartitions);
+                    self.get_partitions_recurse(child, with_holders)?;
+                }
```
jlebon, line 214: OK digging more into this took me to https://github.com/coreos/coreos-installer/pull/648. Let's do that first, and then this should essentially be just `get_partitions()` enumerating over the immediate children only for partitions (so no need for recursion).

Also, maybe as a follow-up we should investigate just returning `Vec<Device>` here instead of a separate `Partition` type.

PR #634, opened 2021-09-24
Makefile: Don't build at install time, and fix up RDCORE conditional

cgwalters:
First, IMO as a general rule `make install` should *not* explicitly
depend on `make` i.e. `make all` because it's very common to use
`sudo make install` type flows which end up building the software
as root.  It's better to get an error if you just do `make install`
and require the user to invoke `make` first.

Secondly, let's not require the user to set the `RDCORE`
for both `make` and `make install`.  Instead, dynamically detect
if we built rdcore and conditionally install it.


cgwalters:
To say this all another way, this makes it Just Work to do e.g.:
`make RDCORE=0 && sudo make install`.

(Assuming you don't have a prior build tree, but we're just moving the state/hysteresis there from the user's shell history to the `target/` dir)

bgilbert:
> First, IMO as a general rule `make install` should _not_ explicitly
> depend on `make` i.e. `make all`

Automake seems to disagree with you, and its UX tends to define expectations in the Make world.  The difference, of course, is that here we don't have proper dependency tracking inside Make: we just hand off to Cargo, which decides to rebuild everything when run as root.  Given that practicality, I'm okay with this.

> (Assuming you don't have a prior build tree, but we're just moving the state/hysteresis there from the user's shell history to the `target/` dir)

It turns out that you do, though.  `make && change something && make RDCORE=0 && sudo make install RDCORE=0` will now install a stale `rdcore`, which seems likely to confuse someone eventually.

cgwalters:
> Automake seems to disagree with you, and its UX tends to define expectations in the Make world. 

True!  Though I don't agree with this part of Automake; they also have other mistakes like trying to support `make uninstall`.

There's not any hard rules here, e.g. the [bootupd Makefile](https://github.com/coreos/bootupd/blob/main/Makefile) also doesn't have the `all -> install` chain.

> The difference, of course, is that here we don't have proper dependency tracking inside Make:

Yeah, agreed.

> It turns out that you do, though. make && change something && make RDCORE=0 && sudo make install RDCORE=0 will now install a stale rdcore, which seems likely to confuse someone eventually.

True.  The more correct way to handle this is to serialize the build options into a state file and dispatch on that, which is basically what autoconf+automake does.

bgilbert:
As a cheap version of state serialization, maybe `make RDCORE=0` should delete an existing `rdcore` build artifact?  That doesn't help the `cargo build && make install` case, but I think it's fair to say that if the user mixes wrappers they get what they get.

bgilbert:
This [just bit us in the RHEL package](https://src.fedoraproject.org/rpms/rust-coreos-installer/pull-request/37).  @cgwalters, are you still interested in landing this?

cgwalters:
I think I'd hand it off to you, please feel free to make any changes you want!

bgilbert:
Okay, updated.

bgilbert (approved):
Approved for Colin's part.  A second approval would be good though.

jlebon (approved):

PR #635, opened 2021-09-24
More prep patches for miniso work

jlebon:
Split out of https://github.com/coreos/coreos-installer/pull/559. See individual commit messages.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use anyhow::{bail, ensure, Context, Error, Result};
+use bincode::Options;
 use flate2::read::GzDecoder;
 use openssl::hash::{Hasher, MessageDigest};
 use openssl::sha;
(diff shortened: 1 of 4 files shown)


jlebon:
OK, clearly I need to updated my local clippy.

src/io.rs
```diff
@@ -525,4 +570,31 @@ mod tests {
...
+        assert_eq!(data, outbuf);
+
+        // limit smaller than data
+        let mut outbuf: Vec<u8> = Vec::new();
+        let mut lim = LimitWriter::new(&mut outbuf, 90, "foo".into());
+        assert_eq!(
+            lim.write_all(&data).unwrap_err().to_string(),
+            "collision with foo at offset 90"
+        );
+    }
```
bgilbert, line 610: There's some marginal value in testing multiple write calls per writer, particularly:
- checking that a write overrun succeeds up to the limit and then fails when `write` is called again at the limit,
- checking that a zero-length write succeeds at the limit.

bgilbert (approved):
Looks great; just some optional nits.  I like the `Address` solution.

jlebon:
Updated!

bgilbert (approved):

PR #636, opened 2021-09-26
Adjust for `lzma-sys` bundling `liblzma` if it doesn't find it

bgilbert:
If `lzma-sys` can't find development headers for `liblzma`, it builds its own copy.  Add `xz-devel` to the builder container so the containerized version uses Fedora's `liblzma` instead.  For the case where the developer doesn't have `liblzma` headers installed, avoid a 2x slowdown by building `lzma-sys` with optimization in the `dev` profile.

It appears that the Fedora package currently uses the system `liblzma` (by happenstance?) and the RHEL package currently bundles.  We'll want to add `BuildRequires: xz-devel` to both in any case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,3 +80,9 @@ opt-level = 3
 opt-level = 3
 [profile.dev.package.miniz_oxide]
 opt-level = 3
+
+# If liblzma development headers aren't installed, lzma-sys builds its own
+# copy of liblzma.  The unoptimized performance is okay but optimizing gives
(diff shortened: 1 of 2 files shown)


bgilbert:
> For RHCOS and RHEL, this should be fixed in the latest branches.

And apparently [I knew about this already](https://src.osci.redhat.com/rpms/coreos-installer/pull-request/27#comment-21440).  Cool.

> For Fedora, AFAICT it gets pulled in as expected by the liblzma dep

Ah, right.

PR #642, opened 2021-09-29
install: print warning when system may have more than one boot/root partition

nikita-dubrovskii:
Inspired by: https://github.com/coreos/fedora-coreos-tracker/issues/976

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -23,6 +23,7 @@ use std::io::{copy, Read, Seek, SeekFrom, Write};
 use std::num::NonZeroU32;
 use std::os::unix::fs::{FileTypeExt, PermissionsExt};
 use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
 
 use crate::blockdev::*;
 use crate::cmdline::*;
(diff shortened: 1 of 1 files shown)


bgilbert:
If the coreos-installer run is automated (e.g. with `coreos.inst.install_dev`) this will present as an install failure.  Is that the intention?

It's also subject to false positives and especially false negatives.  If coreos-installer is being run not on the target system (for example, writing to an external disk and then moving it to a different machine, or writing to a guest VM disk from the host), it won't know what other disks will be attached to the target system.  And if it's being run from one FCOS system to install to a target disk for a different system (maybe this never happens?), or on a different OS that uses the `boot`/`root` labels, the error will be incorrect.

jlebon:
Yeah, we shouldn't make this a hard error, but as a warning or maybe even a note it'd be helpful. In interactive installs, it'll be noticed and acted upon in time. In automated installs, it'll be missed, but then we would catch it at reboot time if we go that way in https://github.com/coreos/fedora-coreos-tracker/issues/976. So this is just about opportunistically trying to save users grief when possible (because it saves them another install iteration).

nikita-dubrovskii:
> If the coreos-installer run is automated (e.g. with `coreos.inst.install_dev`) this will present as an install failure. Is that the intention?

it was.

>  If coreos-installer is being run not on the target system..

yep, missed those cases.

> Yeah, we shouldn't make this a hard error, but as a warning

done

PR #646, opened 2021-10-03
io: split into submodules; add `verify`; use `io::Result`

bgilbert:
`io.rs` is getting a bit unwieldy.  Split it into submodules, then move `verify` in from the top level.

While we're here, clean up unnecessary imports of `std::result`.


jlebon (approved):
Nice!

PR #647, opened 2021-10-04
WIP: warn user when system has several `boot` partitions

nikita-dubrovskii:
Part of: coreos/fedora-coreos-tracker#976

Requires this - https://github.com/coreos/coreos-installer/pull/631


src/bin/rdcore/rootmap.rs
```diff
@@ -27,6 +27,10 @@ use libcoreinst::runcmd_output;
 use crate::cmdline::*;
 
 pub fn rootmap(config: &RootmapConfig) -> Result<()> {
+    // fail if we have more than 1 partition with boot label
+    if !check_single_partition("boot")? {
```
jlebon, line 31: Should this be a separate command instead? E.g. `rdcore verify-single-boot` or something.

The model for `rdcore` I think we should aim for is to try to keep commands modular/narrowly-scoped where fedora-coreos-config just "orchestrates" it. That way we can reorganize things on the FCOS side without having to cut another coreos-installer release/do backports. So e.g. for this, you could imagine we have more than one place where we want to check uniqueness of `boot`. Splitting it out of `rdcore rootmap` makes that possible.

src/blockdev.rs
```diff
@@ -1515,3 +1516,270 @@ mod tests {
...
+            }
+            if let Some(children) = device.children.as_ref() {
+                disks_flatten(children, all);
+            }
+        }
+    }
+
+    let mut all = HashMap::new();
+    disks_flatten(devices, &mut all);
+    eprintln!("{:?}", all);
```
jlebon, line 1551: Hmm, could we keep all the printing in the higher-level functions and avoid printing anything in the utility functions? Maybe instead the API can be `get_partitions_with_label()` and it returns a `Vec`? And then the `install` function can print out the elements if necessary.

src/blockdev.rs
```diff
@@ -1556,6 +1556,7 @@ fn count_partition_by_label(label: &str, devices: &[Device]) -> usize {
 }
 
 pub fn check_single_partition(label: &str) -> Result<bool> {
+    runcmd!("partprobe")?;
```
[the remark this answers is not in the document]
nikita-dubrovskii: on s390x with 2 paths to same SCSI (`multipath` was `off`) i hit the issue during `coreos-gpt-setup`:
- new UUID was generated for `boot`, so `sda` had it, but `sdb` had some other from preious install

So `partprobe` forces system to update its/kernel info regarding UUID

src/bin/rdcore/rootmap.rs
nikita-dubrovskii: i was thikining abot this, so decided first to commit as it is and ask for upstream suggestions. will do

src/blockdev.rs
nikita-dubrovskii: ooops, forgot to remove. that was just for debugging purposes

>Maybe instead the API can be get_partitions_with_label() and it returns a Vec? 

ideally yes, i wanted to do that after #631

src/blockdev.rs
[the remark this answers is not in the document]
jlebon: We discussed this out-of-band today.

I think it would make sense to have `coreos-gpt-setup` do the retriggering instead. Rather than adding `partprobe`, it could use `lsblk` + `blockdev --rereadpt` or call out to e.g. `rdcore rereadpt [DEV]` since we already have the code to trigger partition reloads via `ioctl`.

PR #658, opened 2021-10-11
rdcore/installer: warn user when system has several boot partitions 

nikita-dubrovskii:
This was made to mitigate https://github.com/coreos/fedora-coreos-tracker/issues/976

### DIFF
--- a/src/bin/rdcore/unique_fs.rs
+++ b/src/bin/rdcore/unique_fs.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 CoreOS, Inc.
+// Copyright 2021 CoreOS, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,22 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-mod cmdline;
-mod kargs;
-mod rootmap;
-mod stream_hash;
-mod unique_fs;
-
-use anyhow::Result;
-use structopt::StructOpt;
-
 use crate::cmdline::*;
+use anyhow::{bail, Result};
 
-fn main() -> Result<()> {
-    match Cmd::from_args() {
-        Cmd::Kargs(c) => kargs::kargs(&c),
-        Cmd::Rootmap(c) => rootmap::rootmap(&c),
-        Cmd::StreamHash(c) => stream_hash::stream_hash(&c),
-        Cmd::VerifyUniqueFsLabel(c) => unique_fs::verify_unique_fs(&c),
+use libcoreinst::blockdev::*;
+
+pub fn verify_unique_fs(config: &VerifyUniqueFsLabelConfig) -> Result<()> {
+    let pts = get_filesystems_with_label(&config.label)?;
+    let count = pts.len();
+    if count != 1 {
+        bail!(
+            "System has {} devices with a filesystem labeled '{}': {:?}",
+            count,
+            config.label,
+            pts
+        );
     }
+    Ok(())
 }
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -17,7 +17,7 @@ use gptman::{GPTPartitionEntry, GPT};
 use nix::sys::stat::{major, minor};
 use nix::{errno::Errno, mount, sched};
 use regex::Regex;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::convert::TryInto;
 use std::env;
 use std::fs::{
@@ -806,6 +806,55 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
         bail!("no lsblk results for {}", dev.display());
     }
     Ok(devinfos.remove(0))
+}
+
+/// Rereads partition table and than returns all available filesystems
+/// rereading mitigates possble issue with outdated UUIDs on different
+/// paths to the same disk: after 'ignition-ostree-firstboot-uuid'
+/// '/dev/sdaX' path gets new UUID, but '/dev/sdbX/' path has an old one
+fn lsblk_all_and_reread_partition_tables() -> Result<Vec<HashMap<String, String>>> {
+    let mut cmd = Command::new("lsblk");
+    cmd.arg("--noheadings")
+        .arg("--nodeps")
+        .arg("--list")
+        .arg("--paths")
+        .arg("--output")
+        .arg("NAME");
+    let output = cmd_output(&mut cmd)?;
+    for dev in output.lines() {
+        if let Ok(mut fd) = std::fs::File::open(dev) {
+            let _ = reread_partition_table(&mut fd);
+        }
+    }
+    udev_settle()?;
+    let mut result: Vec<HashMap<String, String>> = Vec::new();
+    for dev in output.lines() {
+        let mut info = lsblk(Path::new(dev), true)?;
+        result.append(&mut info);
+    }
+    Ok(result)
+}
+
+/// Returns filesystems with given label.
+/// If multiple filesystems with the label have the same UUID, we only return one of them.
+pub fn get_filesystems_with_label(label: &str) -> Result<Vec<String>> {
+    let mut uuids = HashSet::new();
+    let result = lsblk_all_and_reread_partition_tables()?
+        .iter()
+        .filter(|v| v.get("LABEL").map(|l| l.as_str()) == Some(label))
+        .filter(|v| match v.get("UUID") {
+            Some(uuid) => {
+                if !uuid.is_empty() {
+                    uuids.insert(uuid)
+                } else {
+                    true
+                }
+            }
+            None => true,
+        })
+        .filter_map(|v| v.get("NAME").map(<_>::to_owned))
+        .collect();
+    Ok(result)
 }
 
 pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<HashMap<String, String>>> {
--- a/src/install.rs
+++ b/src/install.rs
@@ -241,6 +241,33 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         bail!("install failed");
     }
 
+    // Because grub picks /boot by label and the OS picks /boot, we can end up racing/flapping
+    // between picking a /boot partition on startup. So check amount of filesystems labeled 'boot'
+    // and warn user if it's not only one
+    match get_filesystems_with_label("boot") {
+        Ok(pts) => {
+            if pts.len() > 1 {
+                let rootdev = std::fs::canonicalize(&config.device)
+                    .unwrap_or_else(|_| PathBuf::from(&config.device))
+                    .to_string_lossy()
+                    .to_string();
+                let pts = pts
+                    .iter()
+                    .filter(|pt| !pt.contains(&rootdev))
+                    .collect::<Vec<_>>();
+                eprintln!("Note: detected other devices with a filesystem labeled 'boot'.");
+                for pt in pts {
+                    eprintln!("- {}", pt);
+                }
+                eprintln!("The installed OS may not work correctly if there are multiple boot filesystems.
+Before rebooting, investigate whether these filesystems are needed and consider
+wiping them with `wipefs -a`."
+                );
+            }
+        }
+        Err(e) => eprintln!("checking filesystems labeled 'boot': {:?}", e),
+    }
+
     eprintln!("Install complete.");
     Ok(())
 }
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -34,6 +34,8 @@ pub enum Cmd {
     Kargs(KargsConfig),
     /// Copy data from stdin to stdout, checking piecewise hashes
     StreamHash(StreamHashConfig),
+    /// Checks there is only one filesystem with given label
+    VerifyUniqueFsLabel(VerifyUniqueFsLabelConfig),
 }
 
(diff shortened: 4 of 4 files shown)


src/install.rs
```diff
@@ -241,6 +241,20 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         bail!("install failed");
     }
 
+    if let Ok(pts) = get_filesystems_with_label("boot") {
+        if pts.len() != 1 {
```
jlebon, line 245:
```suggestion
        if pts.len() > 1 {
```

That way we're not going to print confusing messages if somehow we can't find the boot partition on the root device we just wrote.

jlebon: I guess to be even more precise, we can filter out from `pts` those that are on the root device, and then do `if !pts.is_empty() {`.

jlebon:
See additional review comments in https://github.com/coreos/coreos-installer/pull/671.

src/blockdev.rs
```diff
@@ -809,6 +809,56 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+        .arg("--paths")
+        .arg("--output")
+        .arg("NAME");
+    let output = cmd_output(&mut cmd)?;
+    for dev in output.lines() {
+        if let Ok(mut fd) = std::fs::File::open(dev) {
+            let _ = reread_partition_table(&mut fd);
+        }
+    }
+    udev_settle()?;
```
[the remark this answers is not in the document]
nikita-dubrovskii: hmm, i'm not sure this could be helpful or could make the code more concise. What's the reason to have `reread_pt` if it would be always `true` ? imo, it's easier to have this logic just in 1 place and always return the most actual info about `filesystems`, than splitting and introducing some flags, which in turn forces user to read low-level details about how flag affects the result

src/install.rs
nikita-dubrovskii: ok, my intention was to warn if there's not just 1 `boot` pt

src/blockdev.rs
```diff
@@ -809,6 +809,50 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+    cmd.arg("--noheadings")
+        .arg("--nodeps")
+        .arg("--list")
+        .arg("--paths")
+        .arg("--output")
+        .arg("NAME");
+    let output = cmd_output(&mut cmd)?;
+    for dev in output.lines() {
+        if let Ok(mut fd) = std::fs::File::open(dev) {
+            let _ = reread_partition_table(&mut fd);
```
bgilbert, line 823: Why are we rereading the partition table at all?  The install path already rereads the partition table for the device we installed to.  Ignition does the same (via sgdisk) when modifying partition tables.  For any devices on the system whose partition tables have been modified by other programs, that program is responsible for initiating a reread.  I don't think we should do this unless there's a clear need.

src/blockdev.rs
```diff
@@ -809,6 +809,50 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+                if !uuid.is_empty() {
+                    uuids.insert(uuid)
+                } else {
+                    true
+                }
+            }
+            None => true,
+        })
+        .filter_map(|v| v.get("NAME"))
+        .map(|v| v.to_owned())
```
bgilbert, line 851: This can be merged into the previous `filter_map`.

src/blockdev.rs
```diff
@@ -809,6 +809,50 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+
+pub fn get_filesystems_with_label(label: &str) -> Result<Vec<String>> {
+    let mut uuids = HashSet::new();
+    let result = lsblk_all_and_reread_partition_tables()?
+        .iter()
+        .filter(|v| v.get("LABEL").map(|l| l.as_str()) == Some(label))
+        .filter(|v| match v.get("UUID") {
+            Some(uuid) => {
+                if !uuid.is_empty() {
+                    uuids.insert(uuid)
```
bgilbert, line 843: This function should probably have a docstring noting that if multiple filesystems with the label have the same UUID, we only return one of them.

src/install.rs
```diff
@@ -241,6 +241,27 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
 
+    match get_filesystems_with_label("boot") {
+        Ok(pts) => {
+            let rootdev = std::fs::canonicalize(&config.device)
+                .unwrap_or_else(|_| PathBuf::from(&config.device))
+                .to_string_lossy()
+                .to_string();
+            let pts = pts
+                .iter()
+                .filter(|pt| !pt.contains(&rootdev))
```
bgilbert, line 252: Using `contains` is probably okay, but is a bit sloppy.  In principle we could split out the first half of `mount_partition_by_label()` into a separate method, and use that to get the actual device path for the installed `boot` partition.  Not sure if it's worth it though.

src/install.rs
```diff
@@ -241,6 +241,27 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+                .collect::<Vec<_>>();
+            if !pts.is_empty() {
+                eprintln!("Note: detected other devices with a filesystem labeled 'boot'");
+                for pt in pts {
+                    eprintln!("- {}", pt);
+                }
+                eprintln!("Before booting the freshly installed system, consider wiping these devices with `wipefs -a`");
```
bgilbert, line 259: I wonder if we can add some more context here.  "Found some other partitions; consider destroying them" seems like dangerous advice on its own.

src/install.rs
```diff
@@ -241,6 +241,27 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         bail!("install failed");
     }
 
+    match get_filesystems_with_label("boot") {
```
bgilbert, line 244: Could we add a comment block here, briefly explaining what we're checking and why?

src/install.rs
```diff
@@ -241,6 +241,27 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+                .collect::<Vec<_>>();
+            if !pts.is_empty() {
+                eprintln!("Note: detected other devices with a filesystem labeled 'boot'");
+                for pt in pts {
+                    eprintln!("- {}", pt);
+                }
+                eprintln!("Before booting the freshly installed system, consider wiping these devices with `wipefs -a`");
+            }
+        }
+        Err(e) => eprintln!("checking filesystems labeled 'boot': {}", e),
```
bgilbert, line 262: Should be `{:?}` so it includes the cause chain.

src/bin/rdcore/unique_fs.rs
```diff
@@ -0,0 +1,33 @@
+// Copyright 2020 CoreOS, Inc.
```
bgilbert, line 1: 2021

src/bin/rdcore/unique_fs.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::cmdline::*;
+use anyhow::{bail, Result};
+
+use libcoreinst::blockdev::*;
+
+pub fn verify_unique_fs(config: &VerifyUniqueFsLabelConfig) -> Result<()> {
+    let pts = get_filesystems_with_label(&config.label)?;
+    if pts.len() != 1 {
+        for pt in &pts {
+            eprintln!("- {}", pt);
```
bgilbert, line 24: We're printing a list of devices first, without context, then failing with an error afterward.  It might make more sense to either print a header/trailer the way install does, or integrate the list into the `bail!()` call.

src/blockdev.rs
nikita-dubrovskii: on `s390x zVM` i hit an error when using one disk with several paths (multipath was off). `/dev/sdaX`path had new `UUID` after `ignition-ostree-firstboot-uuid`, but `/dev/sdbX/` path had old `UUID`

src/install.rs
nikita-dubrovskii: any ideas what to print instead?

src/blockdev.rs
bgilbert: Ah, okay.  That's not an obvious case, so let's add a comment.

src/install.rs
bgilbert: Maybe something like "The installed OS may not work correctly if there are multiple boot filesystems.  Before rebooting, investigate whether these filesystems are needed and consider wiping them with `wipefs -a`."?

If we get this completely robust (e.g. with the followup mentioned in https://github.com/coreos/coreos-installer/pull/671#issue-1039979287), so that the warning is no longer valid, I assume we'll just remove the check.

src/install.rs
```diff
@@ -241,6 +241,31 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+                let pts = pts
+                    .iter()
+                    .filter(|pt| !pt.contains(&rootdev))
+                    .collect::<Vec<_>>();
+                eprintln!("Note: detected other devices with a filesystem labeled 'boot'");
+                for pt in pts {
+                    eprintln!("- {}", pt);
+                }
+                eprintln!("The installed OS may not work correctly if there are multiple boot filesystems");
+                eprintln!("Before rebooting, investigate whether these filesystems are needed and consider wiping them with `wipefs -a`");
```
jlebon, line 263: @bgilbert may ask you to wrap these to 80 columns. :)

jlebon: Also, all these sentences seem to be missing their final period.

jlebon (approved):
Some nits but LGTM overall! Minor nit re. commit title of the first commit: it says "warn" but we're not actually warning. Maybe just... s/warn/inform/ ?

src/install.rs
```diff
@@ -241,6 +241,38 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+                );
+                for pt in pts {
+                    eprintln!("- {}", pt);
+                }
+                eprintln!(
+                    "The installed OS may not work correctly if there are\
+                 multiple boot filesystems.\n\
+                    Before rebooting, investigate whether these filesystems\
+                are needed and consider wiping them with `wipefs -a`."
+                );
```
jlebon, line 270: Sorry for not being clearer: the source lines don't have to be wrapped at the 80 column, just the output. I think this will do the trick:

```suggestion
                eprintln!("Note: detected other devices with a filesystem labeled 'boot':");
                for pt in pts {
                    eprintln!("- {}", pt);
                }
                eprintln!("The installed OS may not work correctly if there are multiple boot filesystems.");
                eprintln!("Before rebooting, investigate whether these filesystems are needed and consider");
                eprintln!("wiping them with `wipefs -a`.");
```

src/install.rs
```diff
@@ -241,6 +241,38 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+            }
+        }
+        Err(e) => eprintln!("checking filesystems labeled 'boot': {:?}.", e),
```
jlebon, line 273: Usually we don't add final periods when error-prefixing. So you can remove it from here.

src/blockdev.rs
```diff
@@ -809,6 +809,55 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
     Ok(devinfos.remove(0))
 }
 
+/// Rereads partition table and than returns all avaible filesystems
```
bgilbert, line 812: `available`

src/blockdev.rs
```diff
@@ -809,6 +809,55 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+            Some(uuid) => {
+                if !uuid.is_empty() {
+                    uuids.insert(uuid)
+                } else {
+                    true
+                }
+            }
+            None => true,
+        })
+        .filter_map(|v| v.get("NAME").map(<_>::to_owned))
```
bgilbert, line 856: I didn't know about this syntax.  Neat!

src/blockdev.rs
```diff
@@ -809,6 +809,55 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+    udev_settle()?;
+    let mut result: Vec<HashMap<String, String>> = Vec::new();
+    for dev in output.lines() {
+        let mut info = lsblk(Path::new(dev), true)?;
+        result.append(&mut info);
+    }
+    Ok(result)
+}
+
+/// Returns filesystmes with given label.
```
bgilbert, line 839: `filesystems`

src/install.rs
```diff
@@ -241,6 +241,40 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         bail!("install failed");
     }
 
+    // Because grub picks /boot by label and the OS picks /boot, we can end up racing/flapping
+    // between picking a /boot partition on startup. So check amount of filesystems labeled 'boot'
+    // and watn user if it's not only one
```
bgilbert, line 246: `warn`

src/install.rs
```diff
@@ -241,6 +241,40 @@ pub fn install(config: &InstallConfig) -> Result<()> {
...
+                for pt in pts {
+                    eprintln!("- {}", pt);
+                }
+                eprintln!(
+                    "The installed OS may not work correctly if there are multiple
+boot filesystems."
+                );
+                eprintln!(
+                    "Before rebooting, investigate whether these filesystems are
+needed and consider wiping them with `wipefs -a`."
```
bgilbert, line 271: These multiline strings have newlines embedded in them, with a narrow right margin.  Could we rebalance them to 80 characters per line shown to the user?

bgilbert (commented):
LGTM; just formatting nits.

bgilbert (approved):
LGTM

jlebon (approved):
LGTM!

PR #659, opened 2021-10-11
[release-0.10] io: check for EOF when decoding a gzip stream (CVE-2021-20319)

bgilbert:
Under normal conditions, `GzDecoder` doesn't read EOF from the underlying source; it stops reading as soon as it reaches the gzip trailer.  Since the wrapped `GpgReader` doesn't see EOF, it doesn't check the exit status of GPG, and a bad signature will not be noticed.  `XzDecoder` is not affected.

This allows bypass of signature verification under uncommon circumstances. Notably, installing from a live ISO or PXE image uses either osmet images (Fedora CoreOS or RHEL CoreOS) or a full copy of the install image (RHEL for Edge), and these are not affected because they're trusted (they come from the installation media we've booted from). These flows are affected:

1.  Installing with `--image-file`, `--image-url`, or `coreos.inst.image_url`.  For example, if a user has a local mirror of installation images, an attacker could replace an image with a gzip-compressed alternative (even if the file extension is `.xz`).  The result:

    ```
    $ coreos-installer install --image-url http://localhost:8080/image.xz /dev/loop0
    Downloading image from http://localhost:8080/image.xz
    Downloading signature from http://localhost:8080/image.xz.sig
    > Read disk 749.9 MiB/749.9 MiB (100%)
    gpg: Signature made Mon 20 Sep 2021 02:41:50 PM EDT
    gpg: using RSA key 8C5BA6990BDB26E19F2A1A801161AE6945719A39
    gpg: BAD signature from "Fedora (34) <fedora-34-primary@fedoraproject.org>" [ultimate]
    Install complete.
    ```

    GPG still complains when its stdin is closed, but coreos-installer doesn't notice.  Automation that relies on coreos-installer's exit status will not notice either.

2. `coreos-installer download --decompress --image-url`:

    ```
    $ coreos-installer download --decompress --image-url http://localhost:8080/image.xz
    > Read disk 749.9 MiB/749.9 MiB (100%)
    gpg: Signature made Mon 20 Sep 2021 02:41:50 PM EDT
    gpg: using RSA key 8C5BA6990BDB26E19F2A1A801161AE6945719A39
    gpg: BAD signature from "Fedora (34) <fedora-34-primary@fedoraproject.org>" [ultimate]
    ./image
    ```

    Again, coreos-installer exits 0.

3. Installing with default parameters, when not using live install media with osmet, if the Red Hat-controlled S3 bucket is compromised or the HTTPS connection is successfully MITMed.

4. `coreos-installer download --decompress` if the S3 bucket is compromised or the HTTPS connection is MITMed.

Fix this by reading one more byte from the source after `GzDecoder` returns EOF.  If that also returns EOF, great, and we've given the source an opportunity to do its own EOF processing.  If not, fail, because the stream contains trailing garbage.

Also add unit tests for signature verification and for detection of bad signatures in the fetch pipeline.

This is a lightweight backport of #655.  It should completely fix the vulnerability, but omits the robustness improvements from #655 to ease backporting.

Bug discovered by @raballew; thanks!


bgilbert:
These commits were pre-approved by @jlebon, who isn't available at the moment to stamp this PR.  Self-merging.

PR #668, opened 2021-10-21
[release-0.10] dracut: also ship zipl_helper.device-mapper on s390x 

julianwiedmann:
Cherry-pick a small s390x fix for multipath setups, so that the `multipath.day1` test goes green.

Missed to queue this up for v0.10.1, here's hoping that there will be a v0.10.2 to piggyback on :).


nikita-dubrovskii:
LGTM

bgilbert:
No 0.10.2 release is planned.  This fix doesn't directly affect upstream users of coreos-installer, so we'd normally just backport the patch to the downstream RPMs. The RHCOS 4.10 RPM already has the patch; if you need it in FCOS for CI, feel free to backport there.

PR #671, opened 2021-10-29
rdcore: add `bind-boot` command

jlebon:
This command does a few things:
- adds a `boot=UUID=<uuid>` karg to ensure the rootfs mounts by UUID
- "claims" the bootfs for the rootfs by adding a stamp file and erroring
  out if one already exists
- adds a GRUB dropin containing the boot UUID

A natural follow-up to this would be have `coreos-installer install`
optimistically pre-randomize the bootfs UUID and adding the stamp file,
though it's worth discussing separately.

Part of coreos/fedora-coreos-tracker#976.

### DIFF
--- a/src/bin/rdcore/rootmap.rs
+++ b/src/bin/rdcore/rootmap.rs
@@ -218,3 +218,73 @@ fn get_luks_uuid(device: &Path) -> Result<String> {
         .trim()
         .into())
 }
+
+pub fn bind_boot(config: &BindBootConfig) -> Result<()> {
+    let boot_mount = Mount::from_existing(&config.boot_mount)?;
+    let root_mount = Mount::from_existing(&config.root_mount)?;
+    let boot_uuid = boot_mount.get_filesystem_uuid()?;
+    let root_uuid = root_mount.get_filesystem_uuid()?;
+
+    let kargs = vec![format!("boot=UUID={}", boot_uuid)];
+    let changed = visit_bls_entry_options(boot_mount.mountpoint(), |orig_options: &str| {
+        if !orig_options.starts_with("boot=") && !orig_options.contains(" boot=") {
+            bls_entry_options_delete_and_append_kargs(orig_options, &[], &kargs, &[])
+        } else {
+            // boot= karg already exists; let's not add anything
+            Ok(None)
+        }
+    })
+    .context("appending boot kargs")?;
+
+    // put it in /run also for the first boot real root mount
+    // https://github.com/coreos/fedora-coreos-config/blob/8661649009/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator#L105-L108
+    if changed {
+        let boot_uuid_run = Path::new("/run/coreos/bootfs_uuid");
+        let parent = boot_uuid_run.parent().unwrap();
+        std::fs::create_dir_all(parent)
+            .with_context(|| format!("creating {}", parent.display()))?;
+        std::fs::write(boot_uuid_run, format!("{}\n", &boot_uuid))
+            .with_context(|| format!("writing {}", boot_uuid_run.display()))?;
+    }
+
+    // bind rootfs to bootfs
+    let root_uuid_stamp = boot_mount.mountpoint().join(".root_uuid");
+    if root_uuid_stamp.exists() {
+        let bound_root_uuid = std::fs::read_to_string(&root_uuid_stamp)
+            .with_context(|| format!("reading {}", root_uuid_stamp.display()))?;
+        let bound_root_uuid = bound_root_uuid.trim();
+        // Let it slide if it already matches the rootfs... that shouldn't happen unless the user
+        // is trying to force a rerun of Ignition. In that case, we'll have nicer errors and
+        // warnings elsewhere.
+        if bound_root_uuid != root_uuid {
+            bail!(
+                "boot filesystem already bound to a root filesystem (UUID: {})",
+                bound_root_uuid
+            );
+        }
+    } else {
+        std::fs::write(&root_uuid_stamp, format!("{}\n", root_uuid))
+            .with_context(|| format!("writing {}", root_uuid_stamp.display()))?;
+    }
+
+    // now bind GRUB to bootfs
+    #[cfg(not(target_arch = "s390x"))]
+    {
+        let grub_bios_path = boot_mount.mountpoint().join("grub2/bootuuid.cfg");
+        write_boot_uuid_grub2_dropin(&boot_uuid, grub_bios_path)?;
+    }
+
+    for esp in find_esps()? {
+        let mount = Mount::try_mount(&esp, "vfat", mount::MsFlags::empty())?;
+        let vendor_dir = find_efi_vendor_dir(&mount)?;
+        let grub_efi_path = vendor_dir.join("bootuuid.cfg");
+        write_boot_uuid_grub2_dropin(&boot_uuid, grub_efi_path)?;
+    }
+    Ok(())
+}
+
+fn write_boot_uuid_grub2_dropin<P: AsRef<Path>>(uuid: &str, p: P) -> Result<()> {
+    let p = p.as_ref();
+    std::fs::write(p, format!("set BOOT_UUID=\"{}\"\n", uuid))
+        .with_context(|| format!("writing {}", p.display()))
+}
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -877,6 +877,57 @@ pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<HashMap<String, String>>
     Ok(result)
 }
 
+pub fn find_esps() -> Result<Vec<String>> {
+    const ESP_TYPE_GUID: &str = "c12a7328-f81f-11d2-ba4b-00a0c93ec93b";
+    let mut cmd = Command::new("lsblk");
+    // Older lsblk, e.g. in CentOS 7.6, doesn't support PATH, but --paths option
+    cmd.arg("--pairs")
+        .arg("--paths")
+        .arg("--output")
+        .arg("NAME,PARTTYPE");
+    let output = cmd_output(&mut cmd)?;
+    output
+        .lines()
+        .filter_map(|line| {
+            let dev = split_lsblk_line(line);
+            if dev.get("PARTTYPE").map(|t| t.as_str()) == Some(ESP_TYPE_GUID) {
+                Some(
+                    dev.get("NAME")
+                        .cloned()
+                        .ok_or_else(|| anyhow!("ESP device with missing NAME")),
+                )
+            } else {
+                None
+            }
+        })
+        .collect()
+}
+
+/// This is basically a Rust version of:
+/// https://github.com/coreos/coreos-assembler/blob/d3c7ec094a02/src/cmd-buildextend-live#L492-L495
+pub fn find_efi_vendor_dir(efi_mount: &Mount) -> Result<PathBuf> {
+    let p = efi_mount.mountpoint().join("EFI");
+    let mut vendor_dir: Vec<PathBuf> = Vec::new();
+    for ent in p.read_dir()? {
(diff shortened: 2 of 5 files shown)


src/bin/rdcore/unique_fs.rs
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::cmdline::*;
+use anyhow::{bail, Result};
+
+use libcoreinst::blockdev::*;
+
+pub fn verify_unique_fs(config: &VerifyUniqueFsLabelConfig) -> Result<()> {
+    if let Ok(pts) = get_filesystems_with_label(&config.label) {
```
cgwalters, line 21: Why are we dropping errors here?  Worth a comment.

Also, I think in *most* places we drop errors, we should at least log them (even if at a debug level).

src/blockdev.rs
```diff
@@ -809,6 +809,56 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+    Ok(result)
+}
+
+pub fn get_filesystems_with_label(label: &str) -> Result<Vec<String>> {
+    let label = label.to_string();
+    let all = lsblk_all()?;
+    let all = all
+        .iter()
+        .filter(|v| v.get("LABEL") == Some(&label))
+        .collect::<Vec<_>>();
```
cgwalters, line 841: Any reason for the intermediate `.collect()` here?  Since we immediately iterate over the returned vector, ISTM we could just chain the iterators.

src/blockdev.rs
```diff
@@ -809,6 +809,56 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+    }
+    Ok(result)
+}
+
+pub fn get_filesystems_with_label(label: &str) -> Result<Vec<String>> {
+    let label = label.to_string();
+    let all = lsblk_all()?;
+    let all = all
+        .iter()
+        .filter(|v| v.get("LABEL") == Some(&label))
```
cgwalters, line 840: It looks like we're heap allocating `label` above just to do this matching; instead it should work to do:
`        .filter(|v| v.get("LABEL").as_deref() == Some(label))`
https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref

src/blockdev.rs
```diff
@@ -829,6 +879,55 @@ pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<HashMap<String, String>>
...
+    cmd.arg("--pairs")
+        .arg("--paths")
+        .arg("--output")
+        .arg("NAME,PARTTYPE");
+    let output = cmd_output(&mut cmd)?;
+    output
+        .lines()
+        .filter_map(|line| {
+            let dev = split_lsblk_line(line);
+            if dev.get("PARTTYPE").map(|t| t.as_str()) == Some(ESP_TYPE_GUID) {
```
cgwalters, line 895: Another `.as_deref()` candidate.

src/blockdev.rs
```diff
@@ -829,6 +879,55 @@ pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<HashMap<String, String>>
...
+                Some(dev.get("NAME").cloned().ok_or_else(|| {
+                    anyhow!("ESP device with missing NAMEmissing found sector size of zero")
```
cgwalters, line 897: This error message looks odd...

src/blockdev.rs
```diff
@@ -829,6 +879,55 @@ pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<HashMap<String, String>>
...
+        let ent = ent.with_context(|| format!("reading directory entry in {}", p.display()))?;
+        if ent.file_name() == "BOOT" {
+            continue;
+        }
+        if !ent.file_type()?.is_dir() {
+            continue;
+        }
+        vendor_dir.push(ent.path());
+    }
+    if vendor_dir.len() != 1 {
```
cgwalters, line 921: I think it'd be cleaner to have `let mut vendor_dir: Option<PathBuf>` and then move the error check inside like:
```
if vendor_dir.replace(ent.path()).is_some() {
   bail!(...)
```

Then there'd be no need for an `unwrap()` at the exit path.

src/install.rs
```diff
@@ -241,6 +241,20 @@ pub fn install(config: &InstallConfig) -> Result<()> {
         bail!("install failed");
     }
 
+    if let Ok(pts) = get_filesystems_with_label("boot") {
```
cgwalters, line 244: Similar comment around error squashing.

src/blockdev.rs
```diff
@@ -809,6 +809,56 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
     Ok(devinfos.remove(0))
 }
 
+fn lsblk_all() -> Result<Vec<HashMap<String, String>>> {
+    let mut cmd = Command::new("lsblk");
+    cmd.arg("--noheadings")
```
cgwalters, line 814: Optional: I find these slightly more readable as `cmd.args(&["--noheadings", "--nodeps", "--list"])` etc.

(There's also my https://docs.rs/sh-inline/0.1.0/sh_inline/macro.bash_command.html macro which can make this just `bash_command!("lsblk --noheadings --nodeps --list --paths --output NAME")` safely)

src/blockdev.rs
```diff
@@ -809,6 +809,56 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
...
+                    uuids.insert(uuid)
+                } else {
+                    true
+                }
+            }
+            _ => true,
+        })
+        .filter_map(|v| v.get("NAME"))
+        .map(|v| v.to_owned())
+        .collect::<Vec<_>>();
```
cgwalters, line 858: I suspect you could drop the turbofish here, type infererence will know it from the return value.

cgwalters (commented):
Collection of minor nits; superficially the high level looks sane.

jlebon:
Requires: https://github.com/coreos/coreos-installer/pull/658

src/blockdev.rs
jlebon: I actually did exactly that at first, but decided to match the logic in https://github.com/coreos/coreos-assembler/blob/d3c7ec094a02/src/cmd-buildextend-live#L492-L495 in that respect too since I like how it automatically handles the 0 case too.

src/blockdev.rs
jlebon: Heh, I've also tried this at first, but

```
error[E0308]: mismatched types
   --> src/blockdev.rs:895:55
    |
895 |             if dev.get("PARTTYPE").as_deref() == Some(ESP_TYPE_GUID) {
    |                                                       ^^^^^^^^^^^^^ expected struct `std::string::String`, found `str`
    |
    = note: expected reference `&std::string::String`
               found reference `&'static str`
```

I think it's because we're not dealing with an `Option<String>`, but an `Option<&String>`.

jlebon:
Updated!

Other comments will be addressed as part of https://github.com/coreos/coreos-installer/pull/658.

src/bin/rdcore/rootmap.rs
```diff
@@ -218,3 +218,62 @@ fn get_luks_uuid(device: &Path) -> Result<String> {
...
+    let root_mount = Mount::from_existing(&config.root_mount)?;
+    let boot_uuid = boot_mount.get_filesystem_uuid()?;
+    let root_uuid = root_mount.get_filesystem_uuid()?;
+
+    let kargs = vec![format!("boot=UUID={}", boot_uuid)];
+    visit_bls_entry_options(boot_mount.mountpoint(), |orig_options: &str| {
+        if !orig_options.starts_with("boot=") && !orig_options.contains(" boot=") {
+            bls_entry_options_delete_and_append_kargs(orig_options, &[], &kargs, &[])
+        } else {
+            Ok(None)
```
bgilbert, line 233: Should we have an explicit comment to call out this case?

src/bin/rdcore/rootmap.rs
```diff
@@ -218,3 +218,62 @@ fn get_luks_uuid(device: &Path) -> Result<String> {
...
+        // is trying to force a rerun of Ignition. In that case, we'll have nicer errors and
+        // warnings elsewhere.
+        if bound_root_uuid != root_uuid {
+            bail!(
+                "boot filesystem already bound to a root filesystem (UUID: {})",
+                bound_root_uuid
+            );
+        }
+    } else {
+        std::fs::write(root_uuid_stamp, root_uuid)?;
```
bgilbert, line 258: This block assumes the stamp file isn't newline-terminated, which is slightly unfriendly for debugging.

src/bin/rdcore/rootmap.rs
```diff
@@ -218,3 +218,62 @@ fn get_luks_uuid(device: &Path) -> Result<String> {
...
+    #[cfg(not(target_arch = "s390x"))]
+    {
+        let grub_bios_path = boot_mount.mountpoint().join("grub2/boot_uuid");
+        write_boot_uuid_grub2_dropin(&boot_uuid, grub_bios_path)?;
+    }
+
+    for esp in find_esps()? {
+        let mount = Mount::try_mount(&esp, "vfat", mount::MsFlags::empty())?;
+        let vendor_dir = find_efi_vendor_dir(&mount)?;
+        let grub_efi_path = vendor_dir.join("boot_uuid");
```
bgilbert, line 271: https://github.com/coreos/coreos-assembler/pull/2524#discussion_r740443214 here too.

src/bin/rdcore/rootmap.rs
```diff
@@ -218,3 +218,62 @@ fn get_luks_uuid(device: &Path) -> Result<String> {
...
+        write_boot_uuid_grub2_dropin(&boot_uuid, grub_efi_path)?;
+    }
+    Ok(())
+}
+
+fn write_boot_uuid_grub2_dropin<P: AsRef<Path>>(uuid: &str, p: P) -> Result<()> {
+    Ok(std::fs::write(p, format!("set BOOT_UUID=\"{}\"", uuid))?)
```
bgilbert, line 278: Missing newline.

src/bin/rdcore/rootmap.rs
```diff
@@ -218,3 +218,62 @@ fn get_luks_uuid(device: &Path) -> Result<String> {
         .trim()
         .into())
 }
+
+pub fn bind_boot(config: &BindBootConfig) -> Result<()> {
```
bgilbert, line 222: Could probably use some more `.context("")`s throughout this function.

src/bin/rdcore/rootmap.rs
```diff
@@ -218,3 +218,62 @@ fn get_luks_uuid(device: &Path) -> Result<String> {
...
+            Ok(None)
+        }
+    })
+    .context("appending boot kargs")?;
+
+    // put it in /run also for the first boot real root mount
+    // https://github.com/coreos/fedora-coreos-config/blob/8661649009/overlay.d/05core/usr/lib/systemd/system-generators/coreos-boot-mount-generator#L105-L108
+    let boot_uuid_run = Path::new("/run/coreos/bootfs_uuid");
+    std::fs::create_dir_all(boot_uuid_run.parent().unwrap())?;
+    std::fs::write(boot_uuid_run, &boot_uuid)?;
```
bgilbert, line 242: Missing newline.

bgilbert (commented):
LGTM generally.

jlebon:
Updated for comments!

bgilbert (approved):

jlebon:
Rebased now that https://github.com/coreos/coreos-installer/pull/658 is in and removed draft status!

bgilbert (approved):

jlebon:
> A natural follow-up to this would be have `coreos-installer install` optimistically pre-randomize the bootfs UUID and adding the stamp file, though it's worth discussing separately.

Opened https://github.com/coreos/coreos-installer/issues/798 for this.

PR #680, opened 2021-11-10
Replace `modify_kargs` and `bls_entry_options_delete_and_append_kargs` with builder struct

bgilbert:
They take multiple arguments of the same type, and the argument order isn't obvious.


bgilbert:
Rebased!

lucab (approved):

jlebon (approved):
Very nice!

PR #688, opened 2021-11-23
Cleanups and prep patches for `install --config-file`

bgilbert:


src/cmdline.rs
```diff
@@ -186,12 +186,12 @@ pub struct InstallConfig {
     pub copy_network: bool,
     /// For use with -n.
     #[structopt(long, value_name = "path")]
-    #[structopt(default_value = "/etc/NetworkManager/system-connections/")]
+    #[structopt(default_value)]
```
jlebon: Minor/optional: could unify these now for consistency.

jlebon (approved):
Very nice! One comment, but LGTM as is too if you don't want to waste the CI cycles.

src/cmdline.rs
bgilbert: Yup, good call.  Also added a commit to drop `structopt(verbatim_doc_comment)` from that field for simplicity.

jlebon (approved):

PR #689, opened 2021-11-23
install: support config files

bgilbert:
Add a repeatable `--config-file` option which specifies a YAML config file.  This contains command-line arguments that are prepended to the actual `coreos-installer install` command line before execution.  This will allow users to configure the install process without resorting to kernel arguments or replacing the `coreos-installer.service` unit.

Make the target device argument optional if a config file is provided, since the config file might specify it instead, but also do a late check in case the config file doesn't provide one either.

Implement config merging by generating a vector of synthesized command-line arguments, prepending it to the actual command line, and having structopt reparse the whole thing.  That allows structopt to apply any validation and constraints specified in struct attributes.

### DIFF
--- a/docs/cmd/install.md
+++ b/docs/cmd/install.md
@@ -40,3 +40,52 @@ Install Fedora CoreOS or RHEL CoreOS
 | **--architecture** *name* | Target CPU architecture [default: x86_64] |
 | **--preserve-on-error** | Don't clear partition table on error |
 | **--fetch-retries** *N* | Fetch retries, or string "infinite" |
+
+## Config file format
+
+Config files specified by `--config-file` are [YAML](https://yaml.org/) documents containing directives with the same names and semantics as command-line arguments.  Each specified config file is parsed in order, and other command-line arguments are parsed afterward.
+
+All parameters are optional.
+
+```yaml
+# Fedora CoreOS stream
+stream: name
+# Manually specify the image URL
+image-url: URL
+# Manually specify a local image file
+image-file: path
+# Embed an Ignition config from a file
+ignition-file: path
+# Embed an Ignition config from a URL
+ignition-url: URL
+# Digest (type-value) of the Ignition config
+ignition-hash: digest
+# Override the Ignition platform ID
+platform: name
+# Append default kernel arguments
+append-karg: [arg1, arg2]
+# Delete default kernel arguments
+delete-karg: [arg1, arg2]
+# Copy network config from install environment
+copy-network: true
+# Source directory for copy-network
+network-dir: path
+# Save partitions with this label glob
+save-partlabel: [glob, glob]
+# Save partitions with this number or range
+save-partindex: [id-or-range, id-or-range]
+# Force offline installation
+offline: true
+# Skip signature verification
+insecure: true
+# Allow Ignition URL without HTTPS or hash
+insecure-ignition: true
+# Base URL for Fedora CoreOS stream metadata
+stream-base-url: URL
+# Target CPU architecture
+architecture: name
+# Don't clear partition table on error
+preserve-on-error: true
+# Fetch retries, or string "infinite"
+fetch-retries: N
+```
--- a/docs/cmd/install.md
+++ b/docs/cmd/install.md
@@ -19,6 +19,7 @@ Install Fedora CoreOS or RHEL CoreOS
 
 ## Options
 
+| **--config-file**, **-c** *path* | YAML config file with install options |
 | **--stream**, **-s** *name* | Fedora CoreOS stream |
 | **--image-url**, **-u** *URL* | Manually specify the image URL |
 | **--image-file**, **-f** *path* | Manually specify a local image file |
--- a/scripts/coreos-installer-service
+++ b/scripts/coreos-installer-service
@@ -13,6 +13,9 @@ PERSIST_DRACUT_NET_PARAMS=("ip" "ifname" "rd.route" "bootdev" "BOOTIF" "rd.booti
 # IBM S390X params to persist
 PERSIST_S390X_PARAMS=("rd.dasd" "rd.zfcp" "rd.znet" "zfcp.allow_lun_scan" "cio_ignore")
 
+# Installer config directory
+INSTALLER_CONFIG_DIR=/etc/coreos/installer.d
+
 args=("install")
 
 cmdline=( $(</proc/cmdline) )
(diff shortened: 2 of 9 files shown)


jlebon:
I wonder if a simpler alternative would be to first check for `--config-file` args, and if found, rerun `Cmd::from_iter()`, pointing it to a `Vec` of what we read in + everything else that's left over? (And do this iteratively). That would avoid all the serialization goop I think.

src/cmdline.rs
```diff
@@ -586,10 +686,428 @@ impl DefaultString for Architecture {
...
+    #[test]
+    fn serialize_empty_command_line() {
+        let expected = ["/dev/missing"];
+        let config = InstallConfig::from_args(&expected).unwrap();
+        assert_eq!(config.to_args().unwrap(), expected);
+    }
+
+    /// Test multiple config files overlapping with command-line arguments
+    #[test]
+    fn install_config_file_overlapping_field() {
```
jlebon, line 1092: Would be good to expand this test (or add a separate test) with a YAML doc containing all the fields. Can reuse the values from `serialize_full_install_config` so we check the deserialized version against the `InstallConfig` there.

docs/cmd/install.md
```diff
@@ -19,6 +19,7 @@ Install Fedora CoreOS or RHEL CoreOS
 
 ## Options
 
+| **--config-file**, **-c** *path* | YAML config file with install options |
```
jlebon, line 22: Will need more docs for this describing the format.

src/cmdline.rs
```diff
@@ -586,10 +686,428 @@ impl DefaultString for Architecture {
...
+                }
+            }
+        }
+
+        fn output_argument<T: ToString>(&mut self, arg: T) {
+            self.output.push(arg.to_string());
+        }
+    }
+
+    // enormous pile of boilerplate
```
jlebon, line 748: Was most of this copy/pasted from some serde doc? Should we link it here?

src/cmdline.rs
```diff
@@ -215,21 +233,80 @@ pub struct InstallConfig {
...
     /// Don't clear partition table on error
     #[structopt(long)]
     pub preserve_on_error: bool,
     /// Fetch retries, or "infinite"
+    #[serde(skip_serializing_if = "is_default")]
     #[structopt(long, value_name = "N", default_value)]
     pub fetch_retries: FetchRetries,
 
     // positional args
+    // use #[serde(rename(serialize = "_[...]"))] on each of these
```
jlebon, line 248: Can you expand that comment with the why? Basically something like:

```suggestion
    // use #[serde(rename(serialize = "_[...]"))] on each of these so that
    // we don't serialize it as e.g. `--device DEV`
```

jlebon (approved):
Some minor comments but LGTM overall. The amount of boilerplate involved isn't great, and there's a maintenance cost there, though I don't have a better idea if we want to retain the nicer YAML UX (instead of just a pure list of CLI args).

bgilbert:
Discussed this some OOB with @jlebon.  We could dramatically reduce the serde attribute goo/boilerplate if the config file was just a list of command-line arguments, without schema enforcement, but I hesitate to do that.  We'll be encouraging users to write and maintain these config files, so it's nice to have a format that's a little more structured.  I've tried to implement this PR in a way that minimizes ongoing maintenance burden (e.g. by reusing field stringification/destringification needed by structopt), and I've also now added a bunch of comments for context, but there's no doubt that the current approach is a more cumbersome implementation.

I did also try directly merging `InstallConfig` structs, rather than serializing them back to command-line arguments.  That turned out to be worse, especially since structopt can't handle the constraint enforcement in that model.  A lot of this would be easier if we could use a proc macro, but those currently can't be implemented and invoked within the same crate, and it doesn't seem worth adding a second crate for this.

Updated for feedback.  Additional changes in this version:

- Added user-facing config file docs.
- Config files can now override each other, and command-line arguments can override config files.
- A later PR will need clean serialization from `InstallConfig` back to YAML, requiring more struct attributes and a worse way of detecting positional parameters (boo).
- Modified `coreos-installer-service` to run if there are files in `/etc/coreos/installer.d`, and to pass those files to coreos-installer as config files.
- Renamed the `device` config file setting to `dest-device` for clarity.

jlebon (approved):
Nice work! This looks good overall, just some minor items.

jlebon:
We should also add some coverage for this in `kola testiso`.

scripts/coreos-installer-service
```diff
@@ -44,17 +47,29 @@ copy_arg() {
     fi
 }
 
+# Config files
+have_config_file=
+for file in ${INSTALLER_CONFIG_DIR}/*; do
```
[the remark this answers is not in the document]
bgilbert: I don't see any benefit to requiring a particular extension, and it adds a footgun.  We don't need to worry about RPM backup files, editor backup files, etc., because the files will be placed in the directory by Ignition, exist for one boot, and then be deleted.  The footgun is that files lacking the extension would be ignored, which could be unpleasant for users to debug.

bgilbert:
Yeah, actually, this will let us clean up some hacks in `kola testiso`.  I'll add that to the queue.

Updated the docs page.

jlebon (approved):

scripts/coreos-installer-service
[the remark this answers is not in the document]
jlebon: Don't feel strongly about this, but IMO the benefit is simply that it's more descriptive and enforces consistency.

> The footgun is that files lacking the extension would be ignored, which could be unpleasant for users to debug.

Hmm, I'm not sure how much of a concern that is. systemd does the same for many (all?) of its dropins directories like tmpfiles.d, sysusers.d, etc...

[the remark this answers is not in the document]
bgilbert: I looked at a bunch of packages with `.d` directories in `/etc` on my system, and results were mixed:

### Require extension

- auditd
- dbus
- dhclient
- dnf
- dracut
- fwupd
- kmod
- ld.so (RPM package)
- NetworkManager
- ostree
- polkit
- sshd (RPM package)
- Xorg

### Do not require extension

- bash
- dconf
- dnsmasq (RPM package)
- grub
- kdump
- logrotate
- PAM
- sudo

It looks like there may be a trend in newer tools of requiring an extension, which makes sense for RPM packages that ship conffiles.  In this case I'm still not convinced that the extra debug overhead (for admins who may not be familiar with the convention) is worth it, though.  (And it's inconsistent with the `--config-file` option, which doesn't care about extensions.)

I'm going to go ahead and merge for now, and we can revisit in a followup if desired.

PR #690, opened 2021-11-23
docs/cmd: sync; fail CI if not in sync

bgilbert:


travier (approved):

tests/help.sh
```diff
@@ -3,17 +3,44 @@
...
+
+    longopts=$(target/debug/coreos-installer $* --help | \
+        (grep -Eo -- "--[a-zA-Z0-9_-]+" ||:) | \
+        (grep -Ev -- "--(help|version)" ||:))
+    shortopts=$(target/debug/coreos-installer $* --help | \
+        (grep -Eo -- "-[a-zA-Z0-9]," ||:) | \
+        tr -d , | \
+        (grep -Ev -- "-(h|V)" ||:))
+    # Ignore hidden subcommands, subcommands that are only containers for
+    # other subcommands, and subcommands with no options
```
jlebon, line 28: They should all have at least `-h`/`--help`, right? We can use that to verify that our grep was actually successful and we didn't regress on a format change.

jlebon (approved):
One comment, but LGTM otherwise.

That said, I wonder how hard it'd be at this point to write up a script which massages the `--help` output into docs pages. That will ensure they're in sync and make life easier everytime we add/modify things.

bgilbert:
> That said, I wonder how hard it'd be at this point to write up a script which massages the `--help` output into docs pages. That will ensure they're in sync and make life easier everytime we add/modify things.

Well, the docs pages were intended to be expanded versions of the help text, with more detail than can fit in 80 characters.  I think that's still worth pursuing when we get a chance.

bgilbert:
Okay, updated to check that subcommands without non-trivial options are documented, and did some refactoring.

jlebon (approved):

PR #694, opened 2021-11-25
miniso: handle hardlinked files

jlebon:
On ppc64le, some files are hardlinked. The way hardlinks are represented
in ISO9660 is simply that the dirents point to the same sector and have
identical sizes. So they show up as exact copies in the table. So we can
get rid of them by deduplicating. Rust's `Vec::dedup()` makes this easy.

See: https://github.com/coreos/coreos-assembler/issues/2583

### DIFF
--- a/src/iso9660.rs
+++ b/src/iso9660.rs
@@ -221,7 +221,7 @@ pub struct File {
     pub length: u32,
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
 pub struct Address(u32);
 
 impl Address {
--- a/src/miniso.rs
+++ b/src/miniso.rs
@@ -47,7 +47,7 @@ impl Table {
     fn new(
         full_files: &HashMap<String, iso9660::File>,
         minimal_files: &HashMap<String, iso9660::File>,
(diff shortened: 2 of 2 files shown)


jcajka:
Tested fixing the issue https://github.com/coreos/coreos-assembler/issues/2583

jcajka:
Thank you for quick fix.

jlebon:
Thanks for testing!

I just tweaked things a bit because I noticed in [your output](https://github.com/coreos/coreos-assembler/issues/2583#issuecomment-979428779) that we were saying:

```
Matched 238 files of 240
```

which is misleading: the two remaining files were just hardlinks. The latest push adds those up to the final count.

src/iso9660.rs
```diff
@@ -221,7 +221,7 @@ pub struct File {
     pub length: u32,
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq)]
```
bgilbert, line 224: Nit: might as well derive `Eq` in both places while we're here.

bgilbert (approved):

src/iso9660.rs
jlebon: Done!

jlebon:
Only added `Eq`. Going to get this in based on previous approval.

PR #696, opened 2021-11-30
rdcore/rootmap: properly handle linear md devices

bgilbert:
Devices that report type `linear` are md devices, even though the type doesn't start with `raid`.


lucab:
LGTM, but I do remember we were doing some `dm-linear` hacks for LUKS in the past. So I'm not sure if this will conflict with systems still having that layout. I'll defer to @jlebon review.

jlebon (approved):
LGTM!

> LGTM, but I do remember we were doing some `dm-linear` hacks for LUKS in the past. So I'm not sure if this will conflict with systems still having that layout.

It should be fine. This code is only relevant for first boots, so we don't need to worry about interfering with upgraded systems.

bgilbert:
dm-linear shows up in the lsblk `TYPE` field as `dm` rather than `linear`, so there shouldn't be a conflict in any event.

lucab:
Thanks for the feedback, sorry if was overzealous :)

PR #700, opened 2021-12-01
bind-boot: look for ESPs on the same disk(s) as bootfs

jlebon:
Right now, to find the ESP devices, we just query all the devices on the
system with the matching partition type. This is clumsy though. E.g. in
case multipathing is in use, this will match both the multipathed
version and the non-multipathed versions.

Another case is multiple re-installs of the OS on different disks: we're
working to ensure that users don't have multiple boot filesystems, but
haven't so far required them to also wipe old ESPs. So we want to ensure
we only modify the ESPs on *our* boot disk.

Rework things by trying to be smarter: since we know what the right boot
filesystem is (`coreos-boot-edit.service` uses `--boot-mount`), use that
knowledge to derive the ESPs by only looking for them in the same
storage hierarchy.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -877,30 +877,71 @@ pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<HashMap<String, String>>
     Ok(result)
 }
 
-pub fn find_esps() -> Result<Vec<String>> {
-    const ESP_TYPE_GUID: &str = "c12a7328-f81f-11d2-ba4b-00a0c93ec93b";
+/// This is a bit fuzzy, but... this function will return every block device in the parent
+/// hierarchy of `device` capable of containing other partitions. So e.g. parent devices of type
+/// "part" doesn't match, but "disk" and "mpath" does.
+pub fn find_parent_devices(device: &str) -> Result<Vec<String>> {
     let mut cmd = Command::new("lsblk");
-    // Older lsblk, e.g. in CentOS 7.6, doesn't support PATH, but --paths option
     cmd.arg("--pairs")
         .arg("--paths")
+        .arg("--inverse")
         .arg("--output")
-        .arg("NAME,PARTTYPE");
+        .arg("NAME,TYPE")
+        .arg(device);
     let output = cmd_output(&mut cmd)?;
-    output
-        .lines()
-        .filter_map(|line| {
+    let mut parents = Vec::new();
+    // skip first line, which is the device itself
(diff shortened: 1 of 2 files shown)


jlebon:
I tested this + https://github.com/coreos/fedora-coreos-config/pull/1316 with:

```
$ kola run multipath.day1 ext.config.reboot coreos.boot-mirror
```

Sadly I already closed that tmux pane, but I can assure you all three tests passed. :)

bgilbert:
> `coreos.boot-mirror`

At present, that only proves that we found a parent device for `/boot` on RAID 1, not that we successfully found an ESP on it.  It'd be good to test that explicitly.

jlebon:
> > `coreos.boot-mirror`
> 
> At present, that only proves that we found a parent device for `/boot` on RAID 1, not that we successfully found an ESP on it. It'd be good to test that explicitly.

I've manually checked it, though will look at adding it to `core.boot-mirror` as well.

src/blockdev.rs
```diff
@@ -877,30 +877,71 @@ pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<HashMap<String, String>>
...
 }
 
-pub fn find_esps() -> Result<Vec<String>> {
-    const ESP_TYPE_GUID: &str = "c12a7328-f81f-11d2-ba4b-00a0c93ec93b";
+/// This is a bit fuzzy, but... this function will return every block device in the parent
+/// hierarchy of `device` capable of containing other partitions. So e.g. parent devices of type
+/// "part" doesn't match, but "disk" and "mpath" does.
+pub fn find_parent_devices(device: &str) -> Result<Vec<String>> {
     let mut cmd = Command::new("lsblk");
-    // Older lsblk, e.g. in CentOS 7.6, doesn't support PATH, but --paths option
```
bgilbert, line 883: Isn't this comment still relevant?

bgilbert (approved):

src/blockdev.rs
jlebon: https://github.com/coreos/coreos-installer/pull/702

jlebon:
> > > `coreos.boot-mirror`
> > 
> > 
> > At present, that only proves that we found a parent device for `/boot` on RAID 1, not that we successfully found an ESP on it. It'd be good to test that explicitly.
> 
> I've manually checked it, though will look at adding it to `core.boot-mirror` as well.

https://github.com/coreos/coreos-assembler/pull/2594

Issue #545, opened 2021-05-20
Support injecting NM keyfiles into the live ISO


jlebon:
# Feature Request #

Right now, when booting the ISO, one cannot use e.g. pointer Ignition configs, or reference remote resources in their Ignition configs if the system requires non-default network configuration. With the new `coreos-installer iso kargs`, it's possible to inject kernel arguments to set up networking, but (1) complex network setups may be difficult or impossible to express on the cmdline, and (2) we should encourage NM keyfiles over kargs in general because they're much easier to manage.

## Desired Feature ##

Something like

```
$ coreos-installer iso nmconfigs <embed|show|remove>
```

This would add NM configs to a pre-allocated CPIO in the ISO, much like we do for the Ignition CPIO. These would then be picked up by `coreos-copy-firstboot-network` (which would learn to run in the live ISO path, see https://github.com/coreos/fedora-coreos-tracker/issues/841), propagated to the real root of the live ISO, and optionally could be copied into the installed system using `--copy-network`. That way, you only have to define network configs *once* in the whole flow.

And to be symmetrical with PXE, we'd have:

```
$ coreos-installer pxe nmconfigs <wrap|unwrap>
```

Though clearly in the PXE case, one can also just use the kernel command-line (but again, network kargs are not great).

## Other Information ##

This would also be prep for having a "`coreos-installer iso uproot`" command which would be able to convert a full live ISO into a minimal live ISO (where the only difference is the removal of the rootfs CPIO). I've been playing with this idea a bit in the background to see what it would entail. But the primary thing is that obviously the system will need the ability to fetch the rootfs remotely.

One of the end goals for this would be to replace the Assisted Installer's need to remaster their own minimal ISO. Though the feature is useful to a larger audience.

(Consider this a separate RFC from the original feature request here. They do not need to happen together. Will open a separate discussion at some point for that.)

PR #706, opened 2021-12-07
Add high-level subcommands for customizing live ISO images and PXE initrds

bgilbert:
Fully customizing the installation process can involve a lot of moving parts, including installer config files, NetworkManager key files, kernel arguments, and nested Ignition configs.  Add user-friendly `iso customize` and `pxe customize` wrappers allowing common customizations to be performed with individual command-line options.

There are a lot of options here, and I'd like feedback on whether they provide the right set of functionality and whether they're clearly named.

[This tool](https://gist.github.com/bgilbert/6480e278d4fbc1aee769f8f567358c97) is helpful for visualizing Ignition configs generated by these subcommands.

cc @dustymabe for the networking parts in particular.


src/live.rs
```diff
@@ -736,6 +736,10 @@ struct LiveIgnition {
 impl LiveIgnition {
     pub fn from_common(common: &CommonCustomizeConfig) -> Result<Self> {
         let mut conf = Self::default();
+        // Fail at runtime if the live OS coreos-installer doesn't support
```
jlebon, line 758: Not sure if it's worth it, though at least in the ISO case, we could e.g. extract `coreos-installer.service` and check for `/etc/coreos/installer.d` so that we fail at `coreos-installer iso customize` time.

src/live.rs
```diff
@@ -829,6 +836,30 @@ RequiredBy={install_target}",
...
+        self.installer_config_struct(&Self::filename(path)?, &conf)
+    }
+
+    fn installer_config_struct(&mut self, filename: &str, config: &InstallConfig) -> Result<()> {
+        self.live.add_file(
+            format!(
+                "/etc/coreos/installer.d/{:04}-{}",
+                self.installer_serial, filename
+            ),
+            &serde_yaml::to_vec(config).context("serializing installer config")?,
```
jlebon, line 955: Why reserialize here? From a debugging point of view, it'd probably be better to just drop the config verbatim so e.g. comments are retained. We parsed it, so we know it's at least valid.

src/live.rs
```diff
@@ -782,6 +788,20 @@ RequiredBy=multi-user.target"#.into(),
...
     }
 
+    pub fn dest_device(&mut self, device: &str, overwrite: bool) -> Result<()> {
+        println!(
+            "Warning: boot media will overwrite {} without confirmation.",
+            device
+        );
+        if !overwrite {
+            bail!("specify --overwrite-dest-device to continue");
+        }
```
jlebon, line 838: Any reason for this? Our stance so far has been that you better know what you're doing re. passing the device name.

src/live.rs
```diff
@@ -732,6 +732,11 @@ fn ignition_embed_area(iso: &mut IsoFs) -> Result<Region> {
 struct LiveIgnition {
     /// The Ignition config for the live system
     live: Ignition,
+    /// The Ignition config for the destination system
+    dest: Option<Ignition>,
+    /// A user-supplied Ignition config for the dest system, which might be
+    /// merged into the dest config or might become the dest config
+    deferred_dest: Option<ignition_config::Config>,
```
jlebon, line 746: Would it be cleaner and easier to understand to just keep it as a `Vec<ignition_config::Config>` and then we can decide based on `dest.len()` whether to wrap or not at `into_bytes()` time?

src/cmdline.rs
```diff
@@ -409,6 +409,9 @@ pub struct CommonCustomizeConfig {
     // Undocumented option; --dest-device fails without it
     #[structopt(long, hidden = true)]
     pub overwrite_dest_device: bool,
+    /// NetworkManager keyfile for live & dest
+    #[structopt(long, number_of_values = 1, value_name = "path")]
+    pub network_keyfile: Vec<String>,
```
jlebon, line 420: This makes sense to me. Wanted to mention there's https://github.com/coreos/coreos-installer/issues/545 also related to this. In the future, we could have this knob actually use that so that it takes effect at live Ignition time.

src/cmdline.rs
```diff
@@ -399,7 +399,7 @@ pub struct ListStreamConfig {
 
 #[derive(Debug, StructOpt)]
 pub struct CommonCustomizeConfig {
-    /// Ignition config fragment for dest system
```
jlebon: Looks like this belong in some previous commit.

src/cmdline.rs
```diff
@@ -418,6 +418,9 @@ pub struct CommonCustomizeConfig {
     /// NetworkManager keyfile for live & dest
     #[structopt(long, number_of_values = 1, value_name = "path")]
     pub network_keyfile: Vec<String>,
+    /// PEM cert authority bundle for Ignition
```
jlebon, line 421: For consistency and clarity, should this end with "for live & dest" ? E.g. "Ignition PEM cert authority bundle for live & dest".

jlebon (commented):
A couple of comments, but this looks great overall!

src/cmdline.rs
dustymabe: @bgilbert Does this imply something like `--network-keyfile=./foo.nmconnection --network-keyfile=./boo.nmconnection` i.e. allowing it to be used more than once?

@jlebon, does #545 want the network to be applied in the initramfs of the live boot?

dustymabe:
Naming is hard...

- I'm not particularly in love with the use of `dest` but I can't really think of a much better term. `final` maybe?

- Is `dest` required in `--dest-device`. There's no `--live-device`. 

- `--ignition-ca`. Applies to both live and final system by default? Should we say that somewhere?

bgilbert:
> I'm not particularly in love with the use of `dest` but I can't really think of a much better term. `final` maybe?

I'm not thrilled with it either; better ideas welcome.  `final` doesn't obviously relate to the installation device IMO.

> Is `dest` required in `--dest-device`. There's no `--live-device`.

It applies to the destination system just as the other `--dest-*` options do, so I added the `dest` prefix for clarity/consistency.  Also, this keeps the name consistent with the installer config file option, which is also called `dest-device`.

src/live.rs
bgilbert: Part of that stance has been that we won't autodetect the install device for you, because it's too dangerous; you have to specify it explicitly.  But here we're creating an image that overwrites the target disk without explicit user action, so I thought an additional safeguard would be appropriate.

However, it's reasonable to argue that we're time-shifting the explicit user action rather than eliminating it, and if the user wants a timebomb we should give it to them.  I'll drop the hidden option.

src/cmdline.rs
bgilbert: That won't fit, but we can do `Ignition PEM CA bundle for live & dest`.

src/cmdline.rs
```diff
@@ -397,9 +397,24 @@ pub struct ListStreamConfig {
     pub stream_base_url: Option<Url>,
 }
 
+#[derive(Debug, StructOpt)]
+pub struct CommonCustomizeConfig {
+    /// Script to run before installation
+    #[structopt(long, value_name = "path")]
```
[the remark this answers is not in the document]
bgilbert: xref https://github.com/coreos/coreos-installer/pull/305#discussion_r460055902 :grin: 

Consistency seems good, yes.  I still find SCREAMING CASE less readable, and it seems unnecessary since clap wraps value names in `<>` anyway.  But if you feel strongly about it, I can convert these over.

src/cmdline.rs
jlebon: > @jlebon, does #545 want the network to be applied in the initramfs of the live boot?

Yup exactly. I've talked with @bgilbert about this and we'll try to squeeze it in as part of this.

bgilbert: @dustymabe Yes, almost all of the options can be repeated.

bgilbert:
Rebased, refactored for recent developments, added long help.  Not yet tested.

bgilbert:
In particular, `pxe customize` now takes a CoreOS initramfs and prepends it to the customizations.  This is consistent with the `iso customize` semantics, and allows both subcommands to use the new `features.json` to validate that the OS image supports the customization features we're going to use.

To validate that the input is a CoreOS live initrd, we check for `/etc/coreos-live-initramfs`, which was previously an internal API between coreos-assembler and fedora-coreos-config.  This makes the existence of that file into a stable API.

I originally implemented in-place modification of the input, as `iso customize` does by default, but implementing in-place reversion to a pristine image would be tricky.  To avoid the footgun, I omitted in-place customization for now.  I left the subcommand syntax parallel with `iso customize` (`-o` is required, `-o -` for stdout) so we could add this later.

bgilbert:
Also of note, some options want to inform the install process but not necessarily initiate it.  For example, `--network-keyfile` shouldn't, by itself, cause coreos-installer to run inside the live system, which would happen if we unconditionally set `copy-network` in an installer config.  So, in some cases, we need to defer processing the `dest` portion of an option until all the options have been scanned.

bgilbert:
- Switched to reading `features.json` directly from the ISO to improve performance
  - https://github.com/coreos/coreos-assembler/pull/2617
- Renamed dest karg options for consistency with live karg options
- Added tests

tests/customize.sh
```diff
@@ -0,0 +1,304 @@
+#!/bin/bash
```
jlebon, line 1: These tests look great!

jlebon (approved):
Really nice, LGTM! The split commits are helpful.

Mostly just some tiny nits. But I'd be fine getting this in as is too and fixing things up as follow-ups.

tests/customize.sh
```diff
@@ -0,0 +1,304 @@
...
+    rm -f iso
+    "${prog}" iso customize src-iso -o iso "${opts_iso[@]}" "$@"
+}
+
+pxe_customize() {
+    rm -f initrd
+    "${prog}" pxe customize src-initrd -o initrd "$@"
+}
+
+qemu_common() {
```
jlebon, line 95: Would've been nice to leverage `kola qemuexec` for this, but yeah... it'd require a bunch of changes. Might be worth thinking about long-term doing the enhancements necessary and switching over. E.g. I'd love to be able to do e.g. `kola qemuexec -p pxe`.

src/cmdline.rs
```diff
@@ -484,6 +484,13 @@ pub struct CommonCustomizeConfig {
     /// shell.
     #[structopt(long, value_name = "path")]
     pub post_install: Vec<String>,
+    /// Installer config file
```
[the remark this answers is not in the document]
bgilbert: The user will need to look up (or know) the schema anyway, so "YAML" doesn't add anything here IMO.

PR #708, opened 2021-12-09
live: clean up CPIO handling; support extracting from arbitrarily concatenated archives

bgilbert:
1. The CPIO code was written pretty early and could use some cleanup.
2. Teach coreos-installer to extract files from arbitrarily concatenated CPIO archives, such as an Ignition config that's been concatenated onto the official OS initrd.  This only affects `pxe ignition unwrap` for now, but will also be useful infrastructure for switching `pxe customize` to producing a concatenated initrd.  That's more consistent with the `iso customize` UX, and also allows us to avoid possible version skew between the generated config and the OS version it's intended for.


src/io/initrd.rs
```diff
@@ -39,10 +40,14 @@ pub fn make_initrd(members: &[(&str, &[u8])]) -> Result<Vec<u8>> {
 }
 
 /// Extract a compressed or uncompressed CPIO archive and return the
-/// contents of the specified path.
-pub fn extract_initrd<R: Read>(reader: R, path: &str) -> Result<Option<Vec<u8>>> {
+/// contents of paths matching the specified glob patterns.
```
jlebon, line 43: Since we're filtering out anything that isn't a regular file, should this say e.g. "files with paths matching ..."?

src/io/initrd.rs
```diff
@@ -114,21 +142,19 @@ mod tests {
...
-            assert_eq!(
-                extract_initrd(&*archive, &format!("{}/world", dir))
-                    .unwrap()
-                    .unwrap(),
-                b"WORLD\n"
-            );
-        }
-        assert!(extract_initrd(&*archive, "z").unwrap().is_none());
+        assert_eq!(
+            extract_initrd(&*archive, &["*"]).unwrap(),
```
jlebon, line 146: A natural test to add here would be globby dirs.

jlebon (approved):
Cool stuff, LGTM overall! A bit unfortunate about the xz2 copying. Would be good to file something upstream and link it from the commit message and file header at least.

Last commit message should also expand on the why.

src/io/initrd.rs
```diff
@@ -41,29 +41,63 @@ pub fn make_initrd(members: &[(&str, &[u8])]) -> Result<Vec<u8>> {
...
         }
-        if entry.name() == path {
-            let mut result = Vec::with_capacity(entry.file_size() as usize);
-            reader
-                .read_to_end(&mut result)
-                .context("reading CPIO entry contents")?;
-            return Ok(Some(result));
+
+        // finish decompression and recover reader
+        reader = decompressor.into_inner()?;
```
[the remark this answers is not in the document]
bgilbert: Switched to accepting only padding bytes, which do exist.

bgilbert:
It turns out `xz2::write::XzDecoder` provides the semantics we need for trailing data, so I've dropped the copied code and added an adapter that implements an `xz2::bufread::XzDecoder`-like API on top of `xz2::write::XzDecoder`.  I submitted https://github.com/alexcrichton/xz2-rs/pull/86 upstream, so hopefully we can remove the adapter code eventually.  I've referenced that PR from a comment, but omitted it from the commit message to avoid spamming the PR.

Dropped the last commit for now.  It's better suited to the followup PR where it'll be used.

jlebon (approved):
LGTM!

bgilbert:
CI caught a failure :balloon: :tada: :fireworks: in osmet unpacking because we were calling `read()` twice at EOF of uncompressed streams.  Fixed the double `read()` in `DecompressReader` (it's unnecessary) and fixed EOF handling in `OsmetUnpacker`.

jlebon (approved):

PR #709, opened 2021-12-09
*: replace `ok_or_else` with `context`

bgilbert:
It appears that anyhow has always supported `context()` and `with_context()` on `Option`, and we missed it during the error_chain conversion.


jlebon (approved):
Nice!

PR #712, opened 2021-12-10
POC/RFC: Support reinstalls via `kexec`

jlebon:
This is a proof-of-concept showing how we can use `kexec` to support reinstalling assuming we have access to the rootfs image (this is strongly related to or exactly the same as https://github.com/coreos/fedora-coreos-tracker/issues/399 depending on which flavour of "reset" we're talking about).

Demo:

```
$ coreos-installer reinstall /dev/vda --platform qemu --insecure
Downloading https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/rhcos-4.10/410.84.202112091617-0/x86_64/rhcos-410.84.20.
Downloading https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/rhcos-4.10/410.84.202112091617-0/x86_64/rhcos-410.84.202112091617-0-live-rootfs.x86_64.img...
Loading kernel and initramfs with arguments:  coreos.inst.install_dev=/dev/vda console=ttyS0 coreos.inst.platform_id=qemu coreos.inst.insecure
Pivoting
[ 1259.444402] kexec_core: Starting new kernel
[    0.000000] Linux version 4.18.0-305.28.1.el8_4.x86_64 (mockbuild@x86-vm-08.build.eng.bos.redhat.com) (gcc version 8.4.1 20200928 (Re1
[    0.000000] Command line:  coreos.inst.install_dev=/dev/vda console=ttyS0 coreos.inst.platform_id=qemu coreos.inst.insecure
[   15.495212] coreos-installer-service[1066]: coreos-installer install /dev/vda --platform qemu --insecure --fetch-retries infinite
[   15.660528] coreos-installer-service[1066]: Installing Red Hat Enterprise Linux CoreOS 410.84.202112091617-0 (Ootpa) x86_64 (512-byte)
[   15.951946]  vda: vda1 vda2 vda3 vda4
[   15.960578]  vda: vda1 vda2 vda3 vda4
[   16.955472] coreos-installer-service[1066]: Read disk 43.7 MiB/3.7 GiB (1%)
[  100.914384] coreos-installer-service[1066]: Read disk 3.7 GiB/3.7 GiB (100%)
[  106.201139] coreos-installer-service[1066]: Install complete.
[  107.902354] reboot: Restarting system
[  107.907561] reboot: machine restart
...
Red Hat Enterprise Linux CoreOS 410.84.202112091617-0 (Ootpa) 4.10
Ignition: ran on 2021/12/10 23:10:41 UTC (this boot)
Ignition: user-provided config was applied

[core@cosa-devsh ~]$ journalctl --list-boots --no-pager
 0 5588b683714241839237799c72ca28c6 Fri 2021-12-10 23:10:33 UTC—Fri 2021-12-10 23:12:01 UTC
```

It also supports specifying the target Ignition config, image URL, using a local live initramfs and live rootfs, etc...

FCOS is also supported though for there seems to be a bug in the kexec code there.


jlebon:
We should be able to avoid reserializing to kargs by re-using the bits from #706 and 708.

dustymabe:
Interesting.. Do we have any concerns here about preserving data for filesystems that we want to re-use? i.e. is the kexec safe enough for mounted filesystems that we want to preserve?

jlebon:
> Interesting.. Do we have any concerns here about preserving data for filesystems that we want to re-use? i.e. is the kexec safe enough for mounted filesystems that we want to preserve?

If the filesystems to preserve were created by Ignition and part of the same Ignition passed back to `coreos-installer reinstall`, then data should transparently be preserved. Otherwise, the same support for saving existing partitions as added in https://github.com/coreos/coreos-installer/pull/321 is also present.

travier:
We should probably unmount as much as possible and remount everything else RO before doing the kexec.

travier:
Maybe creating a special systemd target with a similar behavior as shutdown could help here.

jlebon:
> Maybe creating a special systemd target with a similar behavior as shutdown could help here.

We use `systemctl kexec` which is pretty much this. :)
And now I realize that's what @dustymabe was actually asking.

src/install.rs
```diff
@@ -278,6 +287,373 @@ wiping them with `wipefs -a`.\n"
...
+    location: Url,
+}
+
+fn get_fcos_live_urls(
+    stream: &str,
+    version: &str,
+    arch: &str,
+    retries: FetchRetries,
+) -> Result<(Url, Url)> {
+    let release_meta_url = format!(
```
cgwalters, line 570: Isn't this covered by https://github.com/coreos/stream-metadata-rust  ?

Oh, I see the comment above.

But...hum, shouldn't we mainly support re-installing from a well known stream (i.e. usually `stable`)?

The use case of reinstalling from a specific release seems like it can be covered by having the user manually provide a URL to a build.

That would allow them to use a specific release from the FCOS prod builds (until we GC them) and would allow them to use mirrored builds more easily.

src/install.rs
```diff
@@ -278,6 +287,373 @@ wiping them with `wipefs -a`.\n"
...
+        .with_context(|| format!("verifying {}", path))?;
+    drop(v);
+    drop(bf);
+
+    f.seek(SeekFrom::Start(0))
+        .with_context(|| format!("seeking to start of {}", path))?;
+    Ok(f)
+}
+
+// XXX: copied from Zincati; need https://github.com/coreos/rpm-ostree/issues/2389
```
cgwalters, line 478: Yeah...I guess what we should really do is take the hit of splitting the bindings out into a separate git repo, publish as a crate, but have rpm-ostree pull it in from git main.

That would just make it hard to add a new required field, but we can live with that I think.

src/install.rs
```diff
@@ -278,6 +287,373 @@ wiping them with `wipefs -a`.\n"
...
+        "--load",
+        &kernel,
+        "--initrd",
+        fd_path(&final_initrd),
+        "--append",
+        &kargs
+    )?;
+
+    eprintln!("Pivoting");
+    runcmd!("systemctl", "kexec")?;
```
cgwalters, line 404: kexec is great but I think can sometimes have bugs, so it might be useful to think about a mode that instead writes the kernel/initrd to e.g. `/boot/coreos-installer-reprovision/{kernel,initramfs.img}` and updates the bootloader entries to use them manually.  Then we can do the reinstall from the initramfs on the next boot.

The hit is a double reboot, but it avoids kexec.

cgwalters (commented):
Wow cool stuff, I should clearly be watching PRs to coreos-installer more.

I think the code here would be a lot smaller if we only supported stream metadata.

I'd also say I'd be a lot more confident in offering this when it has an e2e test.

lukasbestle:
Can you explain how it works from a high level perspective? I.e. what will the result be after the reinstall: Will the root partition be completely reset as if CoreOS was installed to a blank drive but additional filesystems (like a separate `/var` partition) will be untouched? Or have I misunderstood it?

jlebon:
> Can you explain how it works from a high level perspective? I.e. what will the result be after the reinstall: Will the root partition be completely reset as if CoreOS was installed to a blank drive but additional filesystems (like a separate `/var` partition) will be untouched? Or have I misunderstood it?

Yup exactly. The idea is to be equivalent to `coreos-installer install` except that it's less friction than re-hosting PXE artifacts or downloading and booting from the live ISO. Re. additional filesystems, see https://github.com/coreos/coreos-installer/pull/712#issuecomment-992841838.

src/install.rs
```diff
@@ -278,6 +287,373 @@ wiping them with `wipefs -a`.\n"
...
+    }
+
+    let sig = std::fs::read(format!("{}.sig", path))
+        .with_context(|| format!("opening signature file for {}", path))?;
+
+    let mut bf = BufReader::with_capacity(BUFFER_SIZE, &mut f);
+    let mut v = VerifyReader::new(&mut bf, Some(sig.as_slice()), VerifyKeys::Production)
+        .with_context(|| format!("creating verifier for {}", path))?;
+    copy(&mut v, &mut std::io::sink()).with_context(|| format!("reading {}", path))?;
+    v.verify_without_logging_failure()
```
bgilbert, line 468: This method is intended for `try_` paths where we fall back to another code path on signature mismatch.  Since we do actually fail here, we should use the logging variant.

src/install.rs
```diff
@@ -278,6 +287,373 @@ wiping them with `wipefs -a`.\n"
...
+                    &booted_deployment
+                        .base_metadata
+                        .fedora_coreos_stream
+                        .unwrap(),
+                    &booted_deployment.version,
+                    &booted_deployment.base_metadata.basearch,
+                    config.fetch_retries,
+                )?,
+                "rhcos" => get_rhcos_live_urls(
+                    &booted_deployment.version,
```
bgilbert, line 329: To preserve bootimage semantics, shouldn't we be installing the aleph-version and not the current version?

bgilbert (commented):
Nice!  This looks promising.

src/install.rs
jlebon: I mentioned this when I demo'ed it but not here: the idea with using the same release was that I had a faint hope we could reuse all the artifacts we have on hand to make it a fully local experience. Obviously the rootfs squashfs would've taken some creativity. :)

src/install.rs
jlebon: Yeah, possible. One thing worth mentioning though is that this strategy could be useful not just for reinstalling CoreOS from CoreOS, but any situation in which you're running on top of the same block device you'd like to install to. So I could imagine this being used in the container workflow too. In which case, not having to fiddle with boot configuration is a plus because you don't know what the setup is (and of course `reinstall` doesn't really make sense in that context; maybe would belong better under `install` with a switch?).

jlebon:
Re-opened in https://github.com/coreos/coreos-installer/pull/791.

PR #713, opened 2021-12-12
live: add `iso`/`pxe` subcommands for embedding/wrapping NM keyfiles

### LINKED ISSUE #545: Support injecting NM keyfiles into the live ISO

bgilbert:
An Ignition config for the live system might want to fetch remote resources, and the system might need non-default network configuration to do so.  This is possible in a limited way with kargs, but it's better to expose the full flexibility of NetworkManager keyfiles.

Add `iso` subcommands to `embed`/`extract`/`remove` network settings, and `pxe` subcommands to `wrap`/`unwrap` them.  These are analogous to the Ignition embed mechanism, and in the case of ISO images they share the same initrd and embed area.  NetworkManager keyfiles are deposited into `/etc/coreos-firstboot-network` in the initrd, which is supported by the OS
as of https://github.com/coreos/fedora-coreos-config/pull/1371.

The show/unwrap subcommands are a bit different than the Ignition ones, since there might be multiple keyfiles in an initrd.  By default, print each keyfile to stdout, with a header listing the filename.  Also provide a `-C`/`--directory` option to extract each keyfile to a directory.  Because `show` doesn't completely capture the command's functionality, call the ISO subcommand `iso network extract` instead.

Fixes #545.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -73,6 +73,8 @@ pub enum IsoCmd {
     Remove(IsoRemoveConfig),
     /// Embed an Ignition config in a CoreOS live ISO image
     Ignition(IsoIgnitionCmd),
+    /// Embed network settings in a CoreOS live ISO image
+    Network(IsoNetworkCmd),
     /// Modify kernel args in a CoreOS live ISO image
     Kargs(IsoKargsCmd),
     /// Inspect the CoreOS live ISO image
@@ -91,6 +93,16 @@ pub enum IsoIgnitionCmd {
     Show(IsoIgnitionShowConfig),
     /// Remove an existing embedded Ignition config from an ISO image
     Remove(IsoIgnitionRemoveConfig),
+}
+
+#[derive(Debug, StructOpt)]
+pub enum IsoNetworkCmd {
+    /// Embed network settings in an ISO image
+    Embed(IsoNetworkEmbedConfig),
+    /// Extract embedded network settings from an ISO image
+    Extract(IsoNetworkExtractConfig),
+    /// Remove existing network settings from an ISO image
+    Remove(IsoNetworkRemoveConfig),
 }
 
 #[derive(Debug, StructOpt)]
@@ -130,6 +142,8 @@ pub enum OsmetCmd {
 pub enum PxeCmd {
     /// Commands to manage a live PXE Ignition config
     Ignition(PxeIgnitionCmd),
+    /// Commands to manage live PXE network settings
+    Network(PxeNetworkCmd),
 }
 
 #[derive(Debug, StructOpt)]
@@ -138,6 +152,14 @@ pub enum PxeIgnitionCmd {
     Wrap(PxeIgnitionWrapConfig),
     /// Show the wrapped Ignition config in an initrd image
     Unwrap(PxeIgnitionUnwrapConfig),
+}
+
+#[derive(Debug, StructOpt)]
+pub enum PxeNetworkCmd {
+    /// Wrap network settings in an initrd image
+    Wrap(PxeNetworkWrapConfig),
+    /// Extract wrapped network settings from an initrd image
+    Unwrap(PxeNetworkUnwrapConfig),
 }
 
 // As a special case, this struct supports Serialize and Deserialize for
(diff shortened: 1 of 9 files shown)


bgilbert:
Still need to update integration tests.

jlebon (commented):
Haven't finished reviewing yet.

src/io/initrd.rs
```diff
@@ -252,4 +274,51 @@ mod tests {
...
+        initrd.add("c/a/c/f", vec![]);
+        initrd.add("a/b/c/d/e/f", vec![]);
+
+        let mut cpio = Vec::new();
+        XzDecoder::new(&*initrd.to_bytes().unwrap())
+            .read_to_end(&mut cpio)
+            .unwrap();
+
+        let mut source = &*cpio;
+        let mut dirs = Vec::new();
```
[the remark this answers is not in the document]
bgilbert: Good call!

src/live.rs
```diff
@@ -280,6 +268,19 @@ fn write_live_iso(iso: &IsoConfig, input: &mut File, output_path: Option<&String
     Ok(())
 }
 
+fn write_live_pxe(initrd: &Initrd, output_path: Option<&String>) -> Result<()> {
```
jlebon, line 378: Minor: might be worth a docstring for this mentioning that we assume that in the stdout case, the caller already verified that it's not a TTY.

src/live.rs
```diff
@@ -32,6 +32,7 @@ use crate::iso9660::{self, IsoFs};
 use crate::miniso;
 
 const INITRD_IGNITION_PATH: &str = "config.ign";
+const INITRD_NETWORK_DIR: &str = "etc/NetworkManager/system-connections";
```
jlebon, line 35: Let's change this to `etc/coreos-firstboot-network` as per https://github.com/coreos/fedora-coreos-tracker/issues/841.

jlebon (commented):
I've tested this on top of https://github.com/coreos/fedora-coreos-config/pull/1371#issuecomment-992972324 (though modifying `INITRD_NETWORK_DIR` to `etc/coreos-firstboot-network`).

src/cmdline.rs
```diff
@@ -654,6 +715,29 @@ pub struct PxeIgnitionUnwrapConfig {
...
+    /// Write to a file instead of stdout
+    #[structopt(short, long, value_name = "path")]
+    pub output: Option<String>,
+}
+
+#[derive(Debug, StructOpt)]
+pub struct PxeNetworkUnwrapConfig {
+    /// Destination directory
+    #[structopt(short = "C", long, value_name = "path")]
+    pub directory: Option<String>,
```
jlebon, line 735: It's slightly non-intuitive I think that not specifying a directory is not equivalent to specifying the cwd, but I don't have a better idea.

src/live.rs
```diff
@@ -183,6 +215,81 @@ pub fn pxe_ignition_unwrap(config: PxeIgnitionUnwrapConfig) -> Result<()> {
...
+                .write_all(contents)
+                .with_context(|| format!("writing {}", path.display()))?;
+            println!("{}", path.display());
+        }
+    } else {
+        for (i, (path, contents)) in files.iter().enumerate() {
+            if i > 0 {
+                println!();
+            }
+            println!("========== {} ==========", filename(path)?);
```
jlebon, line 283: We could make this a comment instead (`########## {} ##########`). That way, one can redirect the output to a file and get a syntactically correct NM keyfile in the case that there is only one.

jlebon (approved):
Some minor comments, but LGTM overall. Nice work!

src/cmdline.rs
bgilbert: Yeah, agreed.  Clarified the help text a bit.

bgilbert:
Updated and added test!

jlebon (approved):

PR #720, opened 2021-12-16
miniso: fix table validation failure on empty files

bgilbert:
Empty files can be placed anywhere, even in the middle of another file, which can cause table validation to fail.  But they also don't require us to do anything, so just filter them out of the table.  For bookkeeping, continue counting them as a success.

We don't currently have any empty files in the CoreOS ISO; discovered this while making a test fixture.

### DIFF
--- a/src/miniso.rs
+++ b/src/miniso.rs
@@ -67,13 +67,16 @@ impl Table {
         }
 
         entries.sort_by_key(|e| e.minimal.as_sector());
-        // drop duplicate entries (hardlinks), and calculate how many there were for reporting
+        // drop zero-length files (which can overlap with other files) and
+        // duplicate entries (hardlinks), and calculate how many there were
+        // for reporting
         let size = entries.len();
+        entries = entries.drain(..).filter(|e| e.length > 0).collect();
         entries.dedup();
-        let hardlinks = size - entries.len();
+        let extraneous = size - entries.len();
         let table = Table { entries };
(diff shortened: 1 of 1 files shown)


src/miniso.rs
```diff
@@ -66,7 +66,9 @@ impl Table {
             });
         }
 
-        entries.sort_by_key(|e| e.minimal.as_sector());
+        // empty files can share an address with a nonempty file, so make
+        // sure we see the empty ones first
+        entries.sort_by_key(|e| (e.minimal.as_sector(), e.length));
```
jlebon, line 71: Hmm, let's just delete it from the table instead? It doesn't hurt the algorithm but it has no value for our purposes and takes up space in the table. We should still make sure it's accounted for in the `hardlinks` count though (which should be renamed e.g. `extraneous` or something) for reporting.

jlebon: > but it has no value for our purposes

Well, I guess it tests that we handle the 0 case well... So meh, not sure. Would probably still nuke it though. Sounds more like a job for unit tests.

jlebon (approved):

src/miniso.rs
bgilbert: Sounds good; I've done that now.  While testing that change, I caught genisoimage putting empty files at some random offset into another file (!) so the original fix wouldn't have been sufficient anyway.

jlebon (approved):

PR #722, opened 2021-12-17
Add more ISO back-compat testing; fix resulting bugs; move some tests to GH Actions

bgilbert:
Add an additional test fixture for each major historical ISO metadata change.  These were generated with a hacked-up cosa, but coreos/coreos-assembler#2619 should allow easier fixture generation in the future.

Use the fixtures to add some additional back-compat testing, and restructure some existing ad-hoc back-compat testing to use the fixtures.  Fix some bugs found in the process.

Move image tests to a subdirectory and add a wrapper script that knows how to run them against the relevant fixtures or a build directory.  Move fixture-based tests from CoreOS CI to GitHub Actions for faster PR feedback.


tests/iso-ignition.sh
```diff
@@ -98,24 +98,27 @@ rm "${out_iso}"
...
+    coreos-installer iso network extract "${out_iso}" | grep -q "foo=baz"
+    coreos-installer iso ignition remove "${iso}"
+    coreos-installer iso network extract "${out_iso}" | grep -q "foo=baz"
+    (coreos-installer iso ignition show "${iso}" 2>&1 ||:) | grep -q "No embedded Ignition config"
+    coreos-installer iso network remove "${iso}"
+    # verify we haven't written an empty cpio archive
+    dd if="${iso}" skip="${offset}" count="${length}" bs=1 status=none | cmp -n "${length}" - /dev/zero
+    rm "${out_iso}"
+else
+    echo "Failed to embed network settings; skipping"
```
jlebon, line 122: Hmm, maybe we should pass in to the test scripts the set of features we expect to be supported so we don't rely on feature detection itself not regressing.

tests/images.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+call() {
+    msg "$*"
+    local cmd="$1"
+    shift
+    "$(dirname $0)/images/${cmd}" "$@"
+}
+
+if [ -n "${1:-}" ]; then
```
jlebon, line 24: Hmm, kinda awkward UX. Might be simpler to just always run the fixtures-based tests and only run the cosa based ones if a base dir was given. It'll add a couple of seconds to the CoreOS CI run, but meh...

jlebon (approved):
Minor comments, but LGTM overall!

tests/iso-ignition.sh
jlebon: This isn't a blocker to be clear.

bgilbert:
Cool, I'm going to get this in and we can handle the feedback as a followup.

tests/iso-ignition.sh
bgilbert: Each of `iso-ignition.sh` and `iso-network.sh` test whether they work in the presence of the other feature.  If feature detection returns false negatives, `iso-network.sh` will catch it.  If it returns false positives, `unsupported.sh` should catch it, and anyway it doesn't reduce our test coverage here.

So I think it's okay to leave this as is.  That has the benefit that `iso-ignition.sh` can still be run by hand, without special knowledge.

bgilbert:
Followup in https://github.com/coreos/coreos-installer/pull/729.

PR #727, opened 2021-12-17
Dockerfile: disable LTO to avoid build-time OOM in Quay

bgilbert:


jlebon (approved):
LGTM, though would be good to reach you to the Quay.io folks and see if there's a way to increase resources in Quay.

bgilbert:
Shouldn't affect CoreOS CI.  Merging.

bgilbert:
Fedora armv7hl builders are having the same problem, so https://github.com/coreos/coreos-installer/pull/728 drops LTO entirely.

PR #728, opened 2021-12-17
Disable LTO; disable debug symbols in container build

bgilbert:
LTO saves us 1.5 MB (11%) in the release binary with debug symbols disabled.  On the other hand it's been an enormous pain.  The RHCOS package disables LTO because of problems on s390x.  Quay and the Fedora armv7hl builders apparently OOM building 0.12.0 with LTO.  And it's not a default option in the first place.  Get rid of it.

Meanwhile, our binary size in the container is now 125 MB.  Disable debug symbols there to bring it down to 13 MB.  Symbols can make it easier to debug problems in production, but our bug reports don't generally take advantage of that, and the space requirement is excessive.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ RUN dnf install -y cargo openssl-devel xz-devel
 WORKDIR /build
 COPY Cargo.* ./
 COPY src src/
+# Debug symbols are nice but they're not 100+ MB of nice
+RUN sed -i 's/^debug = true$/debug = false/' Cargo.toml
 RUN cargo build --release
 
 FROM registry.fedoraproject.org/fedora:35
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,6 @@ RUN dnf install -y cargo openssl-devel xz-devel
 WORKDIR /build
 COPY Cargo.* ./
 COPY src src/
-# Disable LTO to avoid OOM in Quay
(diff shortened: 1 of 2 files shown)


dustymabe:
> Meanwhile, our binary size in the container is now 125 MB. Disable debug symbols there to bring it down to 13 MB.

Wow - does that mean real space savings in FCOS?

Dockerfile
```diff
@@ -3,8 +3,8 @@ RUN dnf install -y cargo openssl-devel xz-devel
 WORKDIR /build
 COPY Cargo.* ./
 COPY src src/
-# Disable LTO to avoid OOM in Quay
-RUN sed -i 's/^lto = true$/lto = false/' Cargo.toml
+# Debug symbols are nice but they're not 112 MB of nice
```
dustymabe, line 6: completely optional: Make the comment more approximate since the actual space saving could change over time. 

```suggestion
# Debug symbols are nice but they're not 100+ MB of nice
```

dustymabe:
> Wow - does that mean 112MB space savings in FCOS?

I guess that depends on what our flags are in the RPM spec file there.

dustymabe (approved):
LGTM

bgilbert:
> Wow - does that mean 112MB space savings in FCOS?

Nope, in RPM-land we split those symbols out to debuginfo.  The container build does not.

dustymabe (approved):

jlebon (approved):

PR #732, opened 2021-12-20
live: split into submodules

bgilbert:
`live.rs` is 1600 lines and has several layers in it.  Split things out.


cgwalters (approved):
> live.rs is 1600 lines and has several layers in it. Split things out.

Yeah, I can imagine it'd be tough to live with live.rs as is.

Overall LGTM!

src/live/mod.rs
```diff
@@ -1580,17 +1580,22 @@ fn modify_miniso_kargs(f: &mut File, rootfs_url: Option<&String>) -> Result<()>
...
     // also modify the default kargs because we don't want `coreos-installer iso kargs reset` to
     // re-add `coreos.liveiso`
-    let mut kargs_info = KargEmbedInfo::for_iso(&mut iso)?.context(
+    set_default_kargs(&mut iso, new_default_kargs)
+}
+
+// only for miniso generation
+fn set_default_kargs(iso: &mut IsoFs, default: String) -> Result<()> {
+    // also modify the default kargs because we don't want `coreos-installer iso kargs reset` to
+    // re-add `coreos.liveiso`
```
jlebon: That comment is out of place and a repeat of the new callsite.

jlebon (approved):
Nice. One minor/optional comment, but LGTM as is too!

PR #734, opened 2021-12-22
workflows: run tests in Fedora container; add s390x tests via qemu-user

bgilbert:
Run s390x tests in a Fedora container via a Fedora `qemu-user` static binary (since that one works, unlike the one in `ubuntu-latest`).  We can't use `actions-rs/toolchain` with an s390x container, so download and run `rustup` manually as well.  For consistency, use a Fedora container and manual `rustup` in the x86_64 tests too.

This gets us the s390x build, unit test, and lint coverage that we've been missing since switching away from Travis.

### DIFF
--- a/.github/workflows/env-setup
+++ b/.github/workflows/env-setup
@@ -2,6 +2,30 @@
 
 set -euxo pipefail
 
+# Start the job container
+mkdir ~/ctx
+CID=$(podman create -v ${GITHUB_WORKSPACE}:/workspace -v ~/ctx:/ctx \
+    "${CONTAINER}" tail -f /dev/null)
+podman start ${CID}
+
+# Build the wrapper script that will be used as a shell by subsequent job steps
+sudo tee /usr/bin/wrap-container > /dev/null <<EOF
+#!/bin/bash
+set -euxo pipefail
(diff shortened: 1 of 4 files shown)


jlebon (approved):
Nice! Looks like the longest s390x test took 20 minutes, which is not bad at all.

Can we add more arches using this strategy then?

jlebon:
Related: https://github.com/coreos/coreos-assembler/issues/2623. Maybe down the line we bake in cross-building capabilities directly in cosa, even if it's just for building component software and not FCOS/RHCOS itself. Then CI here could use cosa directly (though there are some other tweaks we need to do before it's GHA-friendly).

bgilbert:
> Nice! Looks like the longest s390x test took 20 minutes, which is not bad at all.

It's about 20 minutes with a warm cache and 40-50 minutes with a cold cache.  PRs should mostly see warm caches, since they'll reuse the cache from the last run on `main`.

> Can we add more arches using this strategy then? (Obviously s390x is the big one because we have a bunch of compile conditionals on it and not others, but for completeness it'd be interesting.)

We could.  I'm not sure it's worth the CPU time, though?  aarch64 and ppc64le are pretty aligned with x86_64 as far as the installer is concerned.

PR #740, opened 2022-01-04
Replace structopt with clap 3.1

bgilbert:
clap 3 drops `...` from the help text for arguments that can be repeated, and expects applications to make repeatability clear from the arg description: https://github.com/clap-rs/clap/issues/1571#issuecomment-541313239.  I'm not sure of a good way to do that, so we're currently leaving repeatable arguments implicit.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -16,6 +16,7 @@
 #![allow(clippy::large_enum_variant)]
 
 use anyhow::{anyhow, Context, Error, Result};
+use clap::{AppSettings, StructOpt};
 use reqwest::Url;
 use serde::{Deserialize, Serialize};
 use serde_with::{
@@ -28,8 +29,6 @@ use std::fs::OpenOptions;
 use std::marker::PhantomData;
 use std::num::NonZeroU32;
 use std::str::FromStr;
-use structopt::clap::AppSettings;
-use structopt::StructOpt;
 
 use crate::io::IgnitionHash;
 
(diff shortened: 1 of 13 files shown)


jlebon (approved):
Very nice!

bgilbert:
CI correctly detected that clap 3.1 isn't in Fedora yet.  :tada:

bgilbert:
Should be ready to go!

src/cmdline.rs
```diff
@@ -315,10 +315,10 @@ pub struct InstallConfig {
     #[structopt(next_line_help(true))]
     pub network_dir: DefaultedString<NetworkDir>,
     /// Save partitions with this label glob
+    #[serde(skip_serializing_if = "is_default")]
     #[structopt(long, value_name = "lx")]
     // Allow argument multiple times, but one value each.  Allow "a,b" in
     // one argument.
-    #[serde(skip_serializing_if = "is_default")]
```
jlebon, line 321: OOC, was this actually causing compile issues?

src/cmdline.rs
```diff
@@ -1072,7 +1072,11 @@ mod serializer {
             .context("reading subcommand help text")?;
 
         let mut serializer = Serializer {
-            help_text: String::from_utf8(help).context("decoding subcommand help text")?,
+            help_text: String::from_utf8(help)
+                .context("decoding subcommand help text")?
+                // add trailing space to each line, so push_field()
+                // option check works consistently for boolean flags
```
jlebon, line 1084: I think I understand this, but feels like the commit message could've elaborated a bit on what broke and why this fixes it.

jlebon (approved):
LGTM overall. Nice work!

src/cmdline.rs
bgilbert: Not at all, it's just a cleanup.

src/cmdline.rs
bgilbert: Good call.  Updated.

jlebon (approved):

PR #741, opened 2022-01-06
Consolidate build/debug commands/options into separate command hierarchies

bgilbert:
We've gradually accumulated hidden commands and options throughout the subcommand tree.  Some of these are deprecated, which is legitimate, but the rest are build-time/debug commands/options which developers might want to know about.  Since they're individually hidden, there's no discoverability other than by reading the source code.  And in the case of the debug options, the implementation is commingled with user-facing code.

Add a new `pack` command hierarchy for osmet and miniso packing (which form a private API with cosa) and a `dev` hierarchy for development commands (which have no stability at all).  Move everything relevant into them.  For the `pack` commands, leave the original commands as aliases until cosa can be updated.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -62,6 +62,10 @@ pub enum Cmd {
     // users shouldn't be interacting with this command normally
     #[structopt(setting(AppSettings::Hidden))]
     Pack(PackCmd),
+    /// Development commands (unstable)
+    // users shouldn't be interacting with this command normally
+    #[structopt(setting(AppSettings::Hidden))]
+    Dev(DevCmd),
 }
 
 #[derive(Debug, StructOpt)]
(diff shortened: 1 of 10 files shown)


src/cmdline.rs
```diff
@@ -783,7 +813,9 @@ pub struct IsoResetConfig {
 }
 
 #[derive(Debug, StructOpt)]
-pub struct OsmetPackConfig {
+// default usage line lists all mandatory options and so exceeds 80 characters
+#[structopt(usage = "coreos-installer pack osmet [OPTIONS]")]
```
jlebon, line 817: Seems like overkill to enforce the 80 character rule even for these commands. Though meh, I guess it's good to be consistent.

jlebon (approved):
Neat. Idea makes sense to me!

src/cmdline.rs
bgilbert: Yeah, part of the point of improving discoverability is having readable help, so I think it's worth the hassle.

PR #742, opened 2022-01-06
live: add `dev show/extract initrd` commands

bgilbert:
`cpio -t` and `cpio -i` can't be easily used to read an initrd image composed of several concatenated archives, as CoreOS initrds are.  Sadly, `lsinitrd` can't either.  Add developer commands to list and extract initrd contents.


src/cmdline.rs
```diff
@@ -912,6 +916,32 @@ pub struct PxeNetworkUnwrapConfig {
...
+}
+
+#[derive(Debug, StructOpt)]
+pub struct DevExtractInitrdConfig {
+    /// Output directory
+    #[structopt(short = "C", long, value_name = "path", default_value = ".")]
+    pub directory: String,
+    /// List extracted contents
+    #[structopt(short, long)]
+    pub verbose: bool,
```
jlebon, line 936: Instead of naming this `--verbose`, should it be e.g. `--print-paths`? I can imagine it being useful for scripting. Whereas `verbose` sounds more like adding more info on stderr.

jlebon (commented):
Any plans to add some tests for this? One way is to use it to cross-check the output from `pxe customize`.

Commit message should probably also mention `lsinitrd` which is more canonical than `cpio` to handle initrds, but also suffers from the same limitation.

src/cmdline.rs
bgilbert: I used `-v`/`--verbose` to match the name/behavior of the corresponding options in `tar` and `cpio`.

bgilbert:
Updated commit message and added tests.

jlebon (approved):

Issue #746, opened 2022-01-10
`coreos-installer install` waits 2 seconds for every busy disk on the system


bgilbert:
# Bug #

## Host Operating System Version ##

any

## Target Operating System Version ##

any

## coreos-installer Version ##

&ge; 0.11.0

## Expected Behavior ##

`coreos-installer install` exits quickly after installation is complete.

## Actual Behavior ##

`coreos-installer install` waits two seconds for each disk in use on the system.

## Reproduction Steps ##

Run as root:

```bash
#!/bin/bash

set -euxo pipefail
shopt -s nullglob

dest=

cleanup() {
    if [ -n "$dest" ]; then
        losetup -d "$dest"
    fi
    for d in "$tmpd"/mount-*; do
        umount "$d"
    done
    rm -rf "$tmpd"
}

tmpd=$(mktemp -d)
trap cleanup EXIT
cd "${tmpd}"

for i in $(seq 0 10); do
    truncate -s 1G img-$i
    mkfs.ext4 img-$i
    mkdir mount-$i
    mount -o loop img-$i mount-$i
done

truncate -s 8G target
dest=$(losetup -P -f --show target)
coreos-installer install "$dest" "$@"
```

## Other Information ##

Starting in ace4c7255d767eed4ee9c3404e070672f8775454, we attempt to reread the partition table of every disk on the system.  If the disk is busy, we retry for two seconds before giving up.  Since the rereads are done serially, installation can take a long time to finish on systems with many busy disks, such as VM hosts in some configurations.

PR #749, opened 2022-01-11
blockdev: add 'retry' flag to reread_partition_table 

nikita-dubrovskii:
Avoid unnecessary retry delays while checking for extra boot partitions after install.

fixes https://github.com/coreos/coreos-installer/issues/746

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -136,7 +136,7 @@ impl Disk {
                 .write(true)
                 .open(&self.path)
                 .with_context(|| format!("opening {}", &self.path))?;
-            reread_partition_table(&mut f).map(|_| Vec::new())
+            reread_partition_table(&mut f, false).map(|_| Vec::new())
         };
         if rereadpt_result.is_ok() {
             return rereadpt_result;
@@ -212,7 +212,7 @@ impl PartTableKernel {
 
 impl PartTable for PartTableKernel {
     fn reread(&mut self) -> Result<()> {
-        reread_partition_table(&mut self.file)?;
+        reread_partition_table(&mut self.file, true)?;
         udev_settle()
     }
 }
@@ -798,7 +798,8 @@ fn lsblk_all(rereadpt: bool) -> Result<Vec<HashMap<String, String>>> {
     if rereadpt {
         for dev in output.lines() {
             if let Ok(mut fd) = std::fs::File::open(dev) {
-                let _ = reread_partition_table(&mut fd);
+                // best-effort reread of disk that may have busy partitions; don't retry
+                let _ = reread_partition_table(&mut fd, false);
             }
         }
         udev_settle()?;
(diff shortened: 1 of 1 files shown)


src/blockdev.rs
```diff
@@ -212,7 +212,7 @@ impl PartTableKernel {
 
 impl PartTable for PartTableKernel {
     fn reread(&mut self) -> Result<()> {
-        reread_partition_table(&mut self.file)?;
+        reread_partition_table(&mut self.file, NonZeroU32::new(20).unwrap())?;
```
bgilbert, line 215: Since this is the usual case, maybe we can define a `DEFAULT_PARTITION_TABLE_RETRIES`?  Or, to simplify, it might make sense to just have the function take a `retry` boolean.

bgilbert (commented):
Let's expand the commit message to explain the problem we're solving, not how we're solving it.  The user doesn't specify the retry count &mdash; the function's caller does &mdash; and the goal is to avoid unnecessary retry delays while checking for extra `boot` partitions after install.

src/blockdev.rs
```diff
@@ -136,7 +136,7 @@ impl Disk {
                 .write(true)
                 .open(&self.path)
                 .with_context(|| format!("opening {}", &self.path))?;
-            reread_partition_table(&mut f).map(|_| Vec::new())
+            reread_partition_table(&mut f, true).map(|_| Vec::new())
```
bgilbert, line 139: I think we shouldn't be retrying here.  We're not calling this to update anything after a change; we're calling it exactly because we want to know whether it fails on a busy partition.

src/blockdev.rs
```diff
@@ -798,7 +798,7 @@ fn lsblk_all(rereadpt: bool) -> Result<Vec<HashMap<String, String>>> {
     if rereadpt {
         for dev in output.lines() {
             if let Ok(mut fd) = std::fs::File::open(dev) {
-                let _ = reread_partition_table(&mut fd);
+                let _ = reread_partition_table(&mut fd, false);
```
bgilbert, line 801: Maybe add a comment here?  Something like: "Best-effort reread of disk that may have busy partitions; don't retry"

src/blockdev.rs
```diff
@@ -993,11 +993,12 @@ pub fn get_blkdev_deps_recursing(device: &Path) -> Result<Vec<PathBuf>> {
     Ok(ret)
 }
 
-fn reread_partition_table(file: &mut File) -> Result<()> {
+fn reread_partition_table(file: &mut File, retry: bool) -> Result<()> {
     let fd = file.as_raw_fd();
     // Reread sometimes fails inexplicably.  Retry several times before
     // giving up.
-    for retries in (0..20).rev() {
+    let max_retries = if retry { 20 } else { 1 };
```
bgilbert, line 1000: This is actually `max_tries`.  (But the name of the `retries` variable is correct, because it contains the number of retries remaining.)

bgilbert (approved):

PR #764, opened 2022-01-30
Handle trailing NUL bytes in cpio member filenames

bgilbert:
[dracut-cpio](https://github.com/dracutdevs/dracut/blob/master/src/dracut-cpio/src/main.rs) sometimes adds extra NUL bytes to a member name (and includes them in `namesize`) to pad the header out to a block boundary, allowing the file's data blocks to be reflinked into the archive.  Update the `cpio` crate to a version that handles this, and add a test.


bgilbert:
Upstream PR in https://github.com/jcreekmore/cpio-rs/pull/5.

bgilbert:
Updated to switch to the new `cpio` version.

jlebon (approved):

PR #776, opened 2022-02-13
docs/customizing-install: update for `iso/pxe customize`

bgilbert:
Deemphasize customizing via kernel arguments and manual Ignition embedding, since `iso/pxe customize` has better UX.

Drop documentation of manual pre/post hooks, since it's very clunky and `iso/pxe customize` can handle this.

### DIFF
--- a/docs/customizing-install.md
+++ b/docs/customizing-install.md
@@ -110,6 +110,8 @@ architecture: name
 preserve-on-error: true
 # Fetch retries, or string "infinite"
 fetch-retries: N
+# Destination device
+dest-device: path
 ```
 
 ## Hooking coreos-installer at boot time
--- a/docs/customizing-install.md
+++ b/docs/customizing-install.md
@@ -103,46 +103,6 @@ All config files in the `installer.d` directory are evaluated in
 alphabetical order, and any `coreos.inst` kernel command line arguments are
 evaluated afterward.
 
-### Example manual customization via `installer.d`
-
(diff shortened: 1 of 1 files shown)


docs/customizing-install.md
```diff
@@ -8,10 +8,77 @@ nav_order: 5
...
+  `--live-karg-replace`, `--live-karg-delete`).  There are no `--live-karg`
+  options for the PXE image; modify the PXE boot configuration instead.
+- Running scripts before or after installation (`--pre-install`,
+  `--post-install`).  For example, a pre-install script might run a
+  container that performs hardware validation, or a post-install script
+  might use IPMI to configure the machine to boot from the local disk
+  instead of the network.  Pre-install scripts can change the options
+  processed by coreos-installer, including the choice of destination device,
+  by writing an installer config file to `/etc/coreos/installer.d` (see
+  below).
```
jlebon, line 71: In a follow-up, it'd probably be worth fleshing this out with some examples. E.g. a script which uses the MAC address to auto-generate an NM keyfile with static IP to copy in? Or automatically select the destination device based on disk model?

jlebon (approved):
Nice!

docs/customizing-install.md
bgilbert: https://github.com/coreos/coreos-installer/issues/784

bgilbert:
Moved some additional language here that was dropped from https://github.com/coreos/fedora-coreos-docs/pull/355.

PR #781, opened 2022-02-15
Drop use of `openat-ext`, Rust `std` now has efficient `std::io::copy`

cgwalters:
See https://github.com/rust-lang/rust/pull/75272


sohankunkerkar (approved):

src/live/util.rs
```diff
@@ -62,8 +61,7 @@ pub(super) fn write_live_iso(
                 .tempfile_in(output_dir)
                 .context("creating temporary file")?;
             input.seek(SeekFrom::Start(0)).context("seeking input")?;
-            input
-                .copy_to(output.as_file_mut())
+            std::io::copy(input, output.as_file_mut())
```
bgilbert, line 64: We're already importing `std::io::copy`.

bgilbert (commented):
Awesome, thanks!

The upstream changes landed in 1.50.0, so this applies to every Rust version back to our 1.51.0 MSRV.

cgwalters:
Also xref https://github.com/coreos/openat-ext/pull/43

src/live/util.rs
cgwalters: Ah, copy that!

jlebon (approved):

bgilbert (approved):

PR #782, opened 2022-02-15
Add COPR integration Makefile and add CI test for RPM building 

jlebon:
```
commit d768073f4531ee3666a5206235d49fa83609c3f0
Date:   Tue Feb 15 12:23:37 2022 -0500

    Add COPR integration Makefile

    I'd like to enable auto-builds of this repo to
    https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/ so it could
    eventually feed into
    https://github.com/coreos/fedora-coreos-tracker/issues/910.
```
```
commit b9e7ec2b3bbe79b9dcc79d70c69b5459b92d8957
Date:   Tue Feb 15 12:31:03 2022 -0500

    workflows: make TOOLCHAIN optional

    For a future test, I'd like to ensure that we're using the Rust tooling
    from Fedora. So let's make the rustup setup optional so we don't have to
    worry about which `rustc` is actually being used.

```
```
commit 6ef95ab3a648a225a0f3874e69d611feb6ce47ec
Date:   Tue Feb 15 12:37:30 2022 -0500

    workflows: add "RPM build" test

    Verify that we can build using purely Fedora packages. This will catch
    earlier any Rust packages we need to update or add in Fedora rather than
    find out at release time.

    It also ensures that the new continuous COPR builds will not break since
    we don't vendor there either.
```

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -1,0 +1,34 @@
+---
+name: RPM build
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+permissions:
+  contents: read
+
+jobs:
+  test-rpm-build:
+    name: "RPM build"
+    runs-on: ubuntu-latest
+    container:
+      image: registry.fedoraproject.org/fedora:latest
+      options: --privileged
+    steps:
+      # need to install git before checkout to get a git repo
+      - name: Install packages
+        run: dnf install -y git make mock
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Build RPM
+        run: |
+          mkdir rpms
+          make -f .copr/Makefile srpm outdir=rpms
+          mock --rebuild rpms/*.src.rpm
+          find /var/lib/mock -wholename '*/result/*.rpm' | xargs mv -t rpms
+      - name: Archive RPMs
+        uses: actions/upload-artifact@v2
+        with:
+          name: rpms
+          path: rpms/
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -203,3 +203,27 @@ jobs:
             echo "Rerun 'make docs'."
             exit 1
           fi
+  test-rpm-build:
```
bgilbert, line 206: Here too, I'd be inclined to move this to a different workflow.

.github/workflows/rust.yml
```diff
@@ -203,3 +203,27 @@ jobs:
...
+    name: "RPM build"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Set up environment
+        run: .github/workflows/env-setup
```
bgilbert, line 213: Are we using any features of this script, or are we just doing it this way for consistency?  If the latter, I'd be inclined to just use `options: --privileged`; it'd make the job easier to understand.

bgilbert (commented):
We should require the job in branch protection settings after this merges.

jlebon:
Updated! (Though didn't test it on my fork first, let's see what happens in CI here.)

.github/workflows/rpm.yml
```diff
@@ -0,0 +1,34 @@
+---
+name: Rust
```
bgilbert, line 2: Copypaste error.

bgilbert (approved):

PR #785, opened 2022-02-17
blockdev: if possible use 'partprobe' for rereading partition table

nikita-dubrovskii:
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1105

After rereading partition table  there could be a race between systemd and udev (often happens during CI), that's why 
during execution of `coreos-boot-edit.service` systemd can signal `rdcore` with `SIGTERM`. 
To mitigate that try rereading partition table with `partprobe`, which uses newer `BLKPG` ioctl and updates partition table in safier way.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -796,13 +796,18 @@ fn lsblk_all(rereadpt: bool) -> Result<Vec<HashMap<String, String>>> {
         .arg("NAME");
     let output = cmd_output(&mut cmd)?;
     if rereadpt {
-        for dev in output.lines() {
-            if let Ok(mut fd) = std::fs::File::open(dev) {
-                // best-effort reread of disk that may have busy partitions; don't retry
-                let _ = reread_partition_table(&mut fd, false);
+        // First try using 'partprobe', cause it uses newer BLKPG ioctl and updates partition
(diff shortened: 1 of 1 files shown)


bgilbert:
Hmm, I don't follow.  Are we doing something to cause rdcore to be terminated?  If so, how does this fix it?

In general we want to properly wait for events, since sleeping causes boot delays and also may not be sufficient on heavily-loaded systems.  My understanding is that `udevadm settle` doesn't initiate any new actions, just waits for existing ones, so removing the settle shouldn't fix anything.  Is that incorrect?

bgilbert:
I think there's an argument that `BLKPG` is actually less safe, since it allows changing the bounds of partition device nodes while something is using the device.

If the root cause is the process being killed with SIGTERM, it seems like the right approach is to figure out why that's happening and address it directly.

nikita-dubrovskii:
> I think there's an argument that `BLKPG` is actually less safe, since it allows changing the bounds of partition device nodes while something is using the device.
> 
> If the root cause is the process being killed with SIGTERM, it seems like the right approach is to figure out why that's happening and address it directly.

trying, hard to reproduce with systemd's debug messages. catching/blocking SIGTERM doesn't help - systemd sends SIGABRT

PR #786, opened 2022-02-20
fixtures/iso: add new fixture without `coreiso+`/`coreKarg` headers

bgilbert:
For https://github.com/coreos/coreos-assembler/pull/2455.

This fixture is slightly larger than previous ones because this is the first one built with the upstreamed `cosa buildextend-live --fixture` mechanism, which runs `isohybrid` rather than skipping it.


jlebon (approved):
Nice!

PR #802, opened 2022-03-09
blockdev: rework EFI vendor dir checking

jlebon:
On some Dell machines at least, something (UEFI firmware?) creates a
`Dell` directory in the `EFI` dir. This throws off our logic here which
expects only a single vendor dir.

Let's tweak the logic so that we only consider a "vendor dir" a
directory which has a `grub.cfg`.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1116

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -930,13 +930,13 @@ pub fn find_efi_vendor_dir(efi_mount: &Mount) -> Result<PathBuf> {
     let mut vendor_dir: Vec<PathBuf> = Vec::new();
     for ent in p.read_dir()? {
         let ent = ent.with_context(|| format!("reading directory entry in {}", p.display()))?;
-        if ent.file_name() == "BOOT" {
-            continue;
-        }
         if !ent.file_type()?.is_dir() {
             continue;
         }
-        vendor_dir.push(ent.path());
+        let path = ent.path();
+        if path.join("grub.cfg").is_file() {
+            vendor_dir.push(path);
+        }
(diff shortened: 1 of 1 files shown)


jlebon:
Keeping as draft for now. Debugging some issues with cosa preventing me from testing this locally.

src/blockdev.rs
```diff
@@ -930,13 +930,13 @@ pub fn find_efi_vendor_dir(efi_mount: &Mount) -> Result<PathBuf> {
...
         let ent = ent.with_context(|| format!("reading directory entry in {}", p.display()))?;
-        if ent.file_name() == "BOOT" {
-            continue;
-        }
         if !ent.file_type()?.is_dir() {
             continue;
         }
-        vendor_dir.push(ent.path());
+        let path = ent.path();
+        if path.join("grub.cfg").is_file() {
```
cgwalters, line 937: I'm a little uncertain about us saying we'll always use grub.  I guess it's OK for now, if we ever do something else we can start looking for sd-boot or whatever too.

jlebon: Yeah, I wasn't sure if there was a better heuristic. Doing this somewhat makes sense though because this code is concerned with adding the `bootuuid.cfg` dropins which are currently sourced by `grub.cfg`.

jlebon:
This is tested working now!

cgwalters (approved):

bgilbert (approved):

PR #811, opened 2022-03-23
ci: surface image tests in Blue Ocean pipeline view

bgilbert:


jlebon (approved):

.cci.jenkinsfile
```diff
@@ -48,6 +48,8 @@ cosaPod(buildroot: true, runAsUser: 0) {
             skipMultipath: true,
             skipUEFI: true
         )
-        shwrap("tests/images.sh /srv/fcos/builds/latest/x86_64")
+        stage("Image tests") {
```
jlebon, line 51: I think you might have to put this at the same level as `stage("Test ISO")`.

PR #813, opened 2022-03-25
blockdev: use blkid instead of lsblk for querying all filesystems

nikita-dubrovskii:
`blkid` usage may help with https://github.com/coreos/fedora-coreos-tracker/issues/1105 , as it returns proper `LABEL` when `lsblk` doesn't

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -786,16 +786,16 @@ pub fn lsblk_single(dev: &Path) -> Result<HashMap<String, String>> {
 /// rereadpt mitigates possible issue with outdated UUIDs on different
 /// paths to the same disk: after 'ignition-ostree-firstboot-uuid'
 /// '/dev/sdaX' path gets new UUID, but '/dev/sdbX/' path has an old one
-fn lsblk_all(rereadpt: bool) -> Result<Vec<HashMap<String, String>>> {
-    let mut cmd = Command::new("lsblk");
-    cmd.arg("--noheadings")
-        .arg("--nodeps")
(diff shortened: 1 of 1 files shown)


src/blockdev.rs
```diff
@@ -854,6 +849,30 @@ pub fn lsblk(dev: &Path, with_deps: bool) -> Result<Vec<HashMap<String, String>>
     Ok(result)
 }
 
+/// Parse key-value pairs from blkid.
+fn split_blkid_line(line: &str) -> HashMap<String, String> {
+    let re = Regex::new(r#"([\w/]+): (.+)"#).unwrap();
```
jlebon, line 854: Do we need a regex here or would it work fine if we just split on `:`?

jlebon (commented):
One optional comment, but LGTM overall! For the commit message, can you add the rationale in the body, and for the commit title, use something like `blockdev: use blkid instead of lsblk for querying all filesystems`?

src/blockdev.rs
jlebon: (Fine with a regex too to be clear, but I tend to avoid them if a really simple string manipulation will do.)

src/blockdev.rs
```diff
@@ -1189,6 +1212,50 @@ mod tests {
         );
     }
 
+    #[test]
```
jlebon, line 1215: Awesome, thanks for adding these tests!

jlebon (approved):
LGTM! I just added the same rationale in the first comment to the commit message.

PR #824, opened 2022-04-14
Add man page generation

bgilbert:
Have `make docs` generate subcommand man pages using a `pack` subcommand, and store them in the repo alongside the Markdown docs.  Since we don't need the generation code at runtime, gate it (and its deps) behind an optional Cargo feature, and have `make` enable the feature when building with the debug profile.

These are mostly useful for distribution packages; the container doesn't have a good way to expose them.  Also, the output isn't completely clean, but at least we get man pages.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -201,6 +201,7 @@ jobs:
         run: tests/docs-config-file.sh
       - name: Command-line docs
         run: |
+          rm man/*.8
           make docs
           if [ -n "$(git status --porcelain docs)" ]; then
             echo "Found local changes after regenerating docs:"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -134,6 +134,16 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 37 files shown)


jlebon (commented):
Cool, LGTM overall. I think you mentioned this earlier when we discussed it OOB, but can you expand again on why we're doing it via a conditional CLI command vs via `build.rs` as described in the docs for `clap-mangen`?

bgilbert:
If we run `clap_mangen` in `build.rs`, downstream packagers will need to package it and `roff`, and we'll regenerate the man pages on every build.  And if the `clap_mangen` version changes (either the upstream one or the one packaged downstream), we won't notice any changes to its output.

We already have a `make docs` flow for updating the Markdown docs when the `--help` text changes, and a CI job for checking that those docs aren't stale.  Since the man pages need to be updated under the exact same circumstances, it makes sense to generate the man pages at the same time, and keep `clap_mangen` out of the downstream-visible deps.

jlebon (approved):

PR #839, opened 2022-04-26
cargo: coreos-installer release 0.14.0

jlebon:
Major changes:

- Add aarch64 support to container image
- Add generated manpages

Minor changes:

- Silence error from reporting commands like `iso ignition show` and `pxe ignition unwrap` if output pipe is closed
- docs: Document `dest-device` config file field
- docs: Document `iso/pxe customize` commands
- docs: Drop documentation of pre/post installation hooks

Internal changes:

- bind-boot: Fix EFI vendor directory detection
- verify-unique-fs-label: use `blkid` instead of `lsblk` to make filesystem label querying more reliable
- Delete legacy aliases for `osmet` and `minimal-iso` pack commands used by coreos-assembler

Packaging changes:

- Require Rust ≥ 1.56.0
- Migrate from `structopt` to `clap` 3
- Add `clap_mangen` dependency
- Drop `openat-ext` dependency
- Remove Windows binaries from vendor archive

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,7 +161,7 @@ checksum = "ea221b5284a47e40033bf9b66f35f984ec0ea2931eb03505246cd27a963f981b"
 
 [[package]]
 name = "coreos-installer"
-version = "0.13.1"
+version = "0.14.0"
 dependencies = [
  "anyhow",
  "base64",
@@ -180,7 +180,7 @@ dependencies = [
  "libc",
  "maplit",
  "mbrman",
- "nix 0.23.1",
+ "nix",
  "openssl",
  "pipe",
  "rand",
@@ -424,7 +424,7 @@ checksum = "f5a6ef4b3fdbdf3d312de4fb312ea65d2e58cc046a7554742dbf8eccc49eb0c5"
 dependencies = [
  "bincode",
  "crc",
- "nix 0.22.3",
+ "nix",
  "serde",
  "thiserror",
 ]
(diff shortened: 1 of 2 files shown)


bgilbert:
> Major changes:

"Add aarch64 support to container image"?

> * Automatically generate and install manpages

"Add generated manpages"?

> * Don't ignore SIGPIPE in reporting commands like `iso ignition show`, `pxe ignition unwrap`, and `iso kargs show`

Maybe "Silence error from reporting commands like `iso ignition show` and `pxe ignition unwrap` if output pipe is closed"?  If we're not going to list all the commands, we can probably prune the list a bit.

> * docs: Replace pre/post installation hooks with `iso/pxe customize` invocations

This could imply that we're only documenting hooks via `customize`, but in fact we added an entire docs section on `customize` in general.

> * verify-unique-fs-label: use `blkid` instead of `lsblk` to query filesystem labels

Can we phrase this in terms of its consequences?

> * Delete legacy aliases for `osmet` and `minimal-iso` pack commands

Should we try to emphasize that these are solely cosa integration points?

> * Move to `clap` 3.1

In particular, we're dropping the `structopt` dependency.  And since we don't say what we're moving from, the magnitude of this change is getting buried a bit.

jlebon:
Updated! Nice catch on the aarch64 bit.

> > * docs: Replace pre/post installation hooks with `iso/pxe customize` invocations
> 
> This could imply that we're only documenting hooks via `customize`, but in fact we added an entire docs section on `customize` in general.

I've ended up just making this two separate bullet items to make it clearer.

> > * Move to `clap` 3.1
> 
> In particular, we're dropping the `structopt` dependency. And since we don't say what we're moving from, the magnitude of this change is getting buried a bit.

Reworked this to say "`clap` 3" since it's the major version that's significant here. Also changed to using the word "migrate" to be more explicit. I made the addition of `clap_mangen` separate also.

prestist:
From my perspective LGTM.

bgilbert (approved):
Commit contents LGTM; will circle back to the release notes.

bgilbert:
Notes LGTM generally.

> * docs: Drop documentation of pre/post installation hooks

"...in favor of `customize` commands"?

> * Add `clap_mangen` dependency

Well, sorta.  Packagers probably won't notice or need to care.

PR #851, opened 2022-05-05
s390x: add Secure Execution support

nikita-dubrovskii:


src/bin/rdcore/kargs.rs
```diff
@@ -19,9 +19,21 @@ use libcoreinst::io::*;
...
+#[cfg(target_arch = "s390x")]
+use libcoreinst::blockdev::*;
+
 use crate::cmdline::*;
 use crate::rootmap::get_boot_mount_from_cmdline_args;
 
+#[cfg(target_arch = "s390x")]
+pub fn zipl(config: ZiplConfig) -> Result<()> {
+    if config.boot_mount.is_none() {
+        bail!("--boot-mount required");
```
jlebon, line 31: If it's required, then we could just make the field a `String` instead of an `Option<String>`. The reason it's an `Option` in other `rdcore` commands is because we support specifying it as a mount or a block device.

src/bin/rdcore/kargs.rs
```diff
@@ -19,9 +19,21 @@ use libcoreinst::io::*;
...
 use libcoreinst::s390x;
 
+#[cfg(target_arch = "s390x")]
+use libcoreinst::blockdev::*;
+
 use crate::cmdline::*;
 use crate::rootmap::get_boot_mount_from_cmdline_args;
 
+#[cfg(target_arch = "s390x")]
+pub fn zipl(config: ZiplConfig) -> Result<()> {
```
jlebon, line 29: I'd prefer we put this at the end of the file rather than the start since it's a "guest" here.

src/s390x/zipl.rs
```diff
@@ -29,22 +40,196 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
     Ok(())
 }
 
-/// Runs `zipl` based on Ignition and BLS configuration in `boot`.
-pub fn zipl<P: AsRef<Path>>(boot: P) -> Result<()> {
-    let boot = boot.as_ref();
+fn is_secure_execution() -> bool {
```
jlebon, line 43: The same libostree conversation around error-handling applies here. See e.g. [this code](https://github.com/coreos/coreos-installer/blob/8ab20393c388daecab08c05ed9c71a8bf40cf7e9/src/blockdev.rs#L1000) for how to gracefully handle `ENOENT`.

jlebon, line 43: Also, maybe let's rename to `secure_execution_is_enabled()`?

src/s390x/zipl.rs
```diff
@@ -86,9 +271,37 @@ pub fn zipl<P: AsRef<Path>>(boot: P) -> Result<()> {
         boot.join("loader/entries")
     };
 
-    runcmd!("zipl", "--blsdir", blsdir, "--config", conffile.path())?;
-
-    Ok(())
+    match is_secure_execution() {
```
jlebon, line 274: Isn't this `match` just `if`/`else`? I think clippy has a warning about this.

src/s390x/zipl.rs
```diff
@@ -29,22 +40,196 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        .with_context(|| format!("Found no BLS entries in {}", bls_path.display()))?;
+    let contents =
+        std::fs::read_to_string(&path).with_context(|| format!("reading {}", path.display()))?;
+    let name = format!("{} ", name.trim());
+    let line = contents
+        .lines()
+        .find(|l| l.starts_with(&name))
+        .with_context(|| format!("'{}' doesn't have '{}'", path.display(), name))?;
+    Ok(line[name.len()..].trim().to_owned())
+}
```
jlebon, line 81: We have `visit_bls_entry` which I think could replace all this.

src/s390x/zipl.rs
```diff
@@ -29,22 +40,196 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+
+fn regenerate_initrd<P: AsRef<Path>>(path: P, files: &[PathBuf]) -> Result<NamedTempFile> {
```
jlebon, line 83: Similarly to the cosa PR, I think here rather than decompressing the initrd and appending, we can just append the little bit we need to a copy of the initrd.

Does the initrd in `/boot` need to have the LUKS keys? IIUC, it only actually needs to be in the initrd that becomes part of `sdboot`, right? So we could just leave the canonical initrd alone. That way, we don't need to think about possibly appending the same thing over and over.

jlebon: > i've tried to do so. but that didn't work.

OK, let's debug that bit in the cosa PR first and then circle back here.

> what do you mean? `rdcore` is called just once on firstboot-complete, so luks keys are appended just once

Yes, you're right. Though in the spirit of `rdcore` having re-usable pieces, I think we should leave the possibility open for `rdcore zipl` being called multiple times. Anyway, it sounds like this won't be an issue since we don't have to modify the initrd in `/boot`.

nikita-dubrovskii: done. looks like my issue was lack of `-mindepth 1` and `/etc/luks` in `cpio`

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+
+    let hostkeys = find_files("/etc/se-hostkeys", |p: &PathBuf| {
+        p.file_name()
+            .unwrap_or_default()
+            .to_string_lossy()
+            .starts_with("ibm-z-hostkey-")
+    })?;
+
+    let crypttab = PathBuf::from("/etc/crypttab");
+    let seinitrd = if crypttab.exists() {
```
jlebon, line 160: I don't understand this logic. Don't we always expect the `/etc/crypttab` to exist if Secure Execution is turned on?

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        Ok(mut f) => {
+            let mut buffer = String::new();
+            f.read_to_string(&mut buffer)?;
+            Ok(buffer.trim() == "1")
+        }
+    }
+}
+
+fn find_files<P: AsRef<Path> + ?Sized>(
+    config_path: &P,
```
jlebon, line 56: The variable name here should just be `path` to keep this function generic.

jlebon, line 56: Also should be `P` instead of `&P`. That should allow you to drop the `?Sized` marker.

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+    f: impl Fn(&PathBuf) -> bool,
+) -> Result<Vec<PathBuf>> {
+    let mut entries = read_dir(&config_path)
+        .with_context(|| format!("reading directory {}", config_path.as_ref().display()))?
+        .filter_map(Result::ok)
```
jlebon, line 61: We shouldn't ignore errors here. See the last example of https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect for how one could convert an iterator of `Result<T>` into a `Result<Vec<T>>`

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        .filter_map(|e| if f(&e.path()) { Some(e.path()) } else { None })
```
jlebon, line 62: Instead of passing `e.path()` to the closure, it'd be cleaner I think to pass `&e` directly. Then for the `hostkeys` case, you can do `e.file_name()` and for the LUKS keys, you can do `e.metadata()?.is_file()`.

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        .collect::<Vec<_>>();
+    entries.sort();
```
jlebon, line 64: Is sorting important? From the callers, it doesn't seem like it. So then we could drop this and collapse the whole body of this function to `read_dir(...)...collect()`.

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+    Ok(entries)
+}
+
+fn regenerate_initrd<P: AsRef<Path>>(source: P, files: &[PathBuf]) -> Result<NamedTempFile> {
```
jlebon, line 68: Now that we're just concatenating, we could use the `Initrd` API from `crate::io` to build our mini-initrd. That should simplify this function a lot.

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        p.file_name()
+            .unwrap_or_default()
+            .to_string_lossy()
+            .starts_with("ibm-z-hostkey-")
+    })?;
+
+    let crypttab = PathBuf::from("/etc/crypttab");
+    let seinitrd = if crypttab.exists() {
+        let mut extra = find_files("/etc/luks/", |p: &PathBuf| p.is_file())?;
+        extra.push(PathBuf::from("/etc/luks")); //add directory to make kernel happy
```
jlebon, line 162: Related to the previous comment for example, the `Initrd` API knows to take care of things like this.

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        .arg(&sdboot);
+    for k in hostkeys {
+        cmd.arg("-k").arg(k);
+    }
+    if let Some(seinitrd) = seinitrd.as_ref() {
+        cmd.arg("-r").arg(seinitrd.path());
+    } else {
+        cmd.arg("-r").arg(initrd);
+    }
+    let _ = cmd_output(&mut cmd)?;
```
jlebon, line 188: This would be another good target for factoring out.

src/s390x/zipl.rs
```diff
@@ -29,22 +39,159 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+    initrd.finish()?.flush()?;
+
+    Ok(dest)
+}
+
+fn generate_sdboot<P: AsRef<Path>>(mountpoint: P, boot: P) -> Result<PathBuf> {
     let boot = boot.as_ref();
+    let kernel = Cell::new(String::new());
+    let initrd = Cell::new(String::new());
+    let options = Cell::new(String::new());
```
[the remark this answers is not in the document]
nikita-dubrovskii: `FnMut` requires some other places to be fixed,  i'm not sure `visit_bls_entry` should be changed just because of single usage on `s390x` . but can do it

src/s390x/zipl.rs
nikita-dubrovskii: in general - yes. but i was also using this code without LUKSed rootfs

src/s390x/zipl.rs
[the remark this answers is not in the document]
jlebon: Here's what I've got working locally:

```diff
diff --git a/src/io/bls.rs b/src/io/bls.rs
index 0b7b1e4..481e1ed 100644
--- a/src/io/bls.rs
+++ b/src/io/bls.rs
@@ -30,7 +30,7 @@ use std::path::{Path, PathBuf};
 /// Returns `true` if BLS content was modified.
 pub fn visit_bls_entry(
     mountpoint: &Path,
-    f: impl Fn(&str) -> Result<Option<String>>,
+    mut f: impl FnMut(&str) -> Result<Option<String>>,
 ) -> Result<bool> {
     // walk /boot/loader/entries/*.conf
     let mut config_path = mountpoint.to_path_buf();
diff --git a/src/s390x/zipl.rs b/src/s390x/zipl.rs
index f812e62..d5b6a8a 100644
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -23,7 +23,6 @@ use flate2::write::GzEncoder;
 use flate2::Compression;
 use nix::mount::MsFlags;
 use regex::Regex;
-use std::cell::Cell;
 use std::fs::{copy, create_dir_all, read_dir, File};
 use std::io::{BufReader, Read, Write};
 use std::os::unix::prelude::MetadataExt;
@@ -114,30 +113,25 @@ fn regenerate_initrd<P: AsRef<Path>>(source: P, files: &[PathBuf]) -> Result<Nam

 fn generate_sdboot<P: AsRef<Path>>(mountpoint: P, boot: P) -> Result<PathBuf> {
     let boot = boot.as_ref();
-    let kernel = Cell::new(String::new());
-    let initrd = Cell::new(String::new());
-    let options = Cell::new(String::new());
+    let mut kernel: Option<String> = None;
+    let mut initrd: Option<String> = None;
+    let mut options: Option<String> = None;

-    let read_opts = |contents: &str| {
+    visit_bls_entry(boot, |contents: &str| {
         for l in contents.lines() {
-            if let Some(s) = l.strip_prefix("linux ") {
-                kernel.set(s.trim().to_owned());
-            } else if let Some(s) = l.strip_prefix("initrd ") {
-                initrd.set(s.trim().to_owned());
-            } else if let Some(s) = l.strip_prefix("options ") {
-                options.set(s.trim().to_owned());
+            match l.split_once(' ') {
+                Some(("linux", v)) => kernel = Some(v.trim().to_owned()),
+                Some(("initrd", v)) => initrd = Some(v.trim().to_owned()),
+                Some(("options", v)) => options = Some(v.trim().to_owned()),
+                _ => {}
             }
         }
         Ok(None)
-    };
-    visit_bls_entry(boot, read_opts)?;
+    })?;

-    let kernel = kernel.into_inner();
-    let initrd = initrd.into_inner();
-    let options = options.into_inner();
-    if kernel.is_empty() || initrd.is_empty() || options.is_empty() {
-        bail!("bls config doesn't have kernel/initrd/options");
-    }
+    let kernel = kernel.ok_or_else(|| anyhow!("missing 'linux' key in default BLS config"))?;
+    let initrd = initrd.ok_or_else(|| anyhow!("missing 'initrd' key in default BLS config"))?;
+    let options = options.ok_or_else(|| anyhow!("missing 'options' key in default BLS config"))?;
     let kernel = format!("{}{}", boot.to_string_lossy(), kernel);
     let initrd = format!("{}{}", boot.to_string_lossy(), initrd);

```

src/s390x/zipl.rs
jlebon: If it's always expected to be present, then I'd suggest we assume that here and fail otherwise. That should help simplify the code too.

src/s390x/zipl.rs
```diff
@@ -86,8 +233,32 @@ pub fn zipl<P: AsRef<Path>>(boot: P) -> Result<()> {
         boot.join("loader/entries")
     };
 
-    runcmd!("zipl", "--blsdir", blsdir, "--config", conffile.path())?;
-
+    if secure_execution_is_enabled()? {
```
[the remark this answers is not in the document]
nikita-dubrovskii: > coreos-installer install --append-karg foobar still works in SE mode, right?

Not really, it'd be only possible to have SE with `qemu-secex` image

src/s390x/zipl.rs
jlebon: Hmm, not sure I follow. We're not actually adding symlinks here, no? But also, the bit you're referring to is in `Initrd::from_reader_filtered`, which we shouldn't have to invoke here since we're not reading in any initrd, just creating a new one.

To clarify, I was thinking something like this:

```rust
let mut initrd = Initrd::default();
for file in files {
    initrd.add(&file, std::fs::read(&file)?);
}
// std::io:copy from initrd.to_bytes() to tempfile
```

We could add an `Initrd` API so that it can directly read from the target files when streaming out to be more efficient, but meh doesn't seem worth it given that the files we're dealing here shouldn't be very large.

src/s390x/zipl.rs
```diff
@@ -29,66 +38,232 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        Ok(None)
+    };
+    visit_bls_entry(boot, read_opts)?;
+
+    let kernel = kernel.ok_or_else(|| anyhow!("missing 'linux' key in default BLS config"))?;
+    let initrd = initrd.ok_or_else(|| anyhow!("missing 'initrd' key in default BLS config"))?;
+    let options = options.ok_or_else(|| anyhow!("missing 'options' key in default BLS config"))?;
+
+    let kernel = format!("{}{}", boot.to_string_lossy(), kernel);
+    let initrd = format!("{}{}", boot.to_string_lossy(), initrd);
```
jlebon, line 136: Let's avoid using `to_string_lossy`. Those paths should really never be non-UTF8 but it's easy to do the right thing: we can do `boot.join(kernel)` and `boot.join(initrd)` and return `PathBuf`s. I think all the spots where `kernel` and `initrd` are used in the caller support `Path`s already.

src/s390x/zipl.rs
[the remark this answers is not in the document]
jlebon: Ahh right, good point. That made we realize something related to this: `secure_execution_is_enabled()` will return `true` even in the install case if the user happens to be running `coreos-installer install` from a Secure Execution-enabled host.

Maybe let's add an argument to this `zipl` function so that we only do the Secure Execution check in the "in-host" case (i.e. from `rdcore zipl`)?

src/s390x/zipl.rs
nikita-dubrovskii: ahh, ok. now got what you mean! yep, that makes sense

src/s390x/zipl.rs
```diff
@@ -29,66 +38,232 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+    }
+}
+
+fn find_files<P: AsRef<Path>>(
+    path: P,
+    f: impl Fn(&DirEntry) -> Option<PathBuf>,
+) -> Result<Vec<PathBuf>> {
+    Ok(read_dir(&path)
+        .with_context(|| format!("reading directory {}", path.as_ref().display()))?
+        .collect::<std::io::Result<Vec<_>>>()?
```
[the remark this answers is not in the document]
nikita-dubrovskii: hmm, from the docs: `iterator of [Result<T, E>] items can be collected into Result<Collection<T>, E>` . maybe just crash on error with `unwrap` ?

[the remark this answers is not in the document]
nikita-dubrovskii: i saw, but too much code just for handling that `unlikely` error (anyhow `rdcore` will exit with failure)

src/s390x/zipl.rs
nikita-dubrovskii: > we can do boot.join(kernel)

actually not:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2e566c75273b264d50927b6102ef2644

we'd have to remove `/` , so imo it's easier to use `format!`

src/s390x/zipl.rs
```diff
@@ -29,66 +32,259 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+    )?;
+    luks_files.push(PathBuf::from(crypttab_path));
+    let initrd = generate_initrd(&initrd, &luks_files, &rootfs)?;
+
+    // during cosa-build we override hostkey(s) with a universal one
+    let hostkeys = if let Some(hostkey) = hostkey {
+        vec![PathBuf::from(hostkey)]
     } else {
-        boot.join("loader/entries")
+        let hostkeys_path = format!("{}/etc/se-hostkeys/", rootfs);
```
jlebon, line 167: Then here:

```suggestion
        let hostkeys_path = rootfs.join("etc/se-hostkeys");
```

src/s390x/zipl.rs
```diff
@@ -29,66 +32,259 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        .with_context(|| format!("reading directory {}", path.as_ref().display()))?
+        .map(Result::unwrap)
+        .filter_map(|e| if f(&e) { Some(e.path()) } else { None })
+        .collect::<Vec<_>>())
+}
+
+fn generate_initrd<P: AsRef<Path>>(
+    source: P,
+    files: &[PathBuf],
+    rootfs: &str,
```
jlebon, line 59: Then we change this to a `&Path`.

src/s390x/zipl.rs
```diff
@@ -29,66 +32,259 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+        let path = path
+            .to_str()
+            .with_context(|| format!("converting '{}' to str", path.display()))?;
+
+        // during cosa-build rootfs includes ostree commit path
+        if !rootfs.is_empty() {
+            initrd.add(path.trim_start_matches(rootfs), contents);
+        } else {
+            initrd.add(path, contents);
+        };
```
jlebon, line 88: Then here:

```suggestion
        let contents =
            std::fs::read(path).with_context(|| format!("reading {}", path.display()))?;
        let initrd_member = path
            .strip_prefix(rootfs)
            .unwrap() // safety: `files` are found in the rootfs, so it *must* be a prefix
            .to_str()
            .with_context(|| format!("invalid non-UTF8 path '{}'", path.display()))?;
        initrd.add(initrd_member, contents);
```

src/s390x/zipl.rs
```diff
@@ -29,66 +32,259 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+    }
+}
 
-    // create dummy config for zipl
-    let mut conffile = Builder::new()
-        .prefix("coreos-installer-zipl.")
+fn find_files<P: AsRef<Path>>(path: P, f: impl Fn(&DirEntry) -> bool) -> Result<Vec<PathBuf>> {
+    Ok(read_dir(&path)
+        .with_context(|| format!("reading directory {}", path.as_ref().display()))?
+        .map(Result::unwrap)
```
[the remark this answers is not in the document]
nikita-dubrovskii: got you, will fix

src/bin/rdcore/cmdline.rs
```diff
@@ -119,6 +122,30 @@ pub struct VerifyUniqueFsLabelConfig {
...
+    #[clap(long, value_name = "BOOT_MOUNT")]
+    pub boot_mount: String,
+
+    /// Path to rootfs
+    #[clap(long, value_name = "ROOTFS")]
+    pub rootfs: Option<String>,
+
+    /// Path to hostkey
+    #[clap(long, value_name = "HOSTKEY")]
+    pub hostkey: Option<String>,
```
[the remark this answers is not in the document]
jschintag: No, we are not allowed to include the universal key inside the image, which is why we need to "exchange" it during firstboot. The universal hostkey should at no point be part of the image.
As far as i understand it is not needed, as long as we re-secure the image during firstboot with the hostkey provided by the user.

src/s390x/zipl.rs
```diff
@@ -29,66 +32,259 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
-            extra.extend_from_slice(
-                &firstboot_kargs
-                    .split_whitespace()
-                    .map(|s| s.to_string())
-                    .collect::<Vec<String>>(),
-            );
+        .context("creating cpio for extras")?;
+
+    // copying from original initrd, on s390x SE it's not concatenated
+    std::fs::copy(source, dest.path())
```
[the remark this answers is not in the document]
nikita-dubrovskii: i'm not a fan of `/proc/self/fd/N` . maybe without `tempfile` ?

src/bin/rdcore/cmdline.rs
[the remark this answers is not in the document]
nikita-dubrovskii: than we need that `hostkey` override option

src/s390x/zipl.rs
nikita-dubrovskii: did a bit different

src/s390x/zipl.rs
nikita-dubrovskii: it's an Option now

src/s390x/mod.rs
```diff
@@ -20,3 +22,10 @@ pub use zipl::chreipl;
 pub use zipl::zipl;
 mod eckd;
 mod fba;
+
+#[derive(ArgEnum, Clone, Debug)]
+pub enum ZiplMode {
+    Auto,
+    EnforceSecex,
+    DisableSecex,
+}
```
jlebon, line 31: And then we'd rename this to `ZiplSecexMode` and the members would be `Auto`, `Enforce`, and `Disable`.

src/s390x/zipl.rs
```diff
@@ -29,66 +33,273 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+
+    let mut initrd = Initrd::default();
+    for path in files {
+        let mut f = File::open(path).with_context(|| format!("opening {}", path.display()))?;
+        let mut contents = Vec::new();
+        f.read_to_end(&mut contents)
+            .with_context(|| format!("reading {}", path.display()))?;
+        let mut path = path
+            .to_str()
+            .with_context(|| format!("converting '{}' to str", path.display()))?;
```
jlebon, line 89: The convention is to mention UTF-8 in these errors, e.g. `path {} is not UTF-8`.

jlebon (commented):
This is getting close!

There are a few places where we do path manipulations as strings where I think it would've been better to use the proper `Path` APIs (as suggested in https://github.com/coreos/coreos-installer/pull/851#discussion_r887300582), but not a blocker.

src/bin/rdcore/cmdline.rs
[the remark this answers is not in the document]
nikita-dubrovskii: it's done by `ignition`. we can try symlinking `/dev/disk/by-label/hostkey` to that predefined path, but that looks awkward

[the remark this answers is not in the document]
nikita-dubrovskii: >we're missing that firstboot code which injects the host-specific keys

this universal key is a special one, best option would be to include it into `secex-qemu` luksed-rootfs and later during firstboot to replace it with host-specific. maybe that'd be possible some day..

jlebon (approved):
LGTM! :tada:

jlebon:
Are you ready for this to merge or are you still testing it?

nikita-dubrovskii:
> Are you ready for this to merge or are you still testing it?

ready to merge )

nikita-dubrovskii:
> (I've rebased this to pick up #873.)

thx!

PR #855, opened 2022-05-09
ci: mark git checkout as safe

jlebon:
Recent git became more strict wrt git repos in parent dirs owned by
other users. This broke us likely due to the container workdir being a
volume mount. We need to explicitly mark the repo as safe.

Interestingly, the checkout action also does this but only temporarily
for itself.

For more information, see:
https://github.com/actions/checkout/issues/760


bgilbert (approved):
Thanks!

PR #859, opened 2022-05-13
cargo: regenerate docs when releasing

bgilbert:
The man pages embed the coreos-installer version number.  `cargo-release` has a mechanism for doing string replacement on files, but it's easier to just regenerate the docs.  Run `cargo clean` afterward to preserve the clean checkout contemplated by the release checklist.


bgilbert:
CI didn't catch this because of https://github.com/coreos/coreos-installer/pull/860.

PR #864, opened 2022-05-26
customize: Add --network-nmstate option

qinqon:
It's possible to generate NetworkManager keyfiles from a Nmstate
configuration. This change add a new option to use Nmstate format
instead of multiple NetworkManager keyfiles to specify network
configuration.

The output from nmstate `gen_conf` is a HashMap of backend name to nmconnection vectors with the following format 

```yaml
NetworkManager:
  - "[connection]\nautoconnect=true\nautoconnect-slaves=-1\nid=dummyfoo\ninterface-name=dummyfoo\ntype=dummy\nuuid=157d8614-c3bd-4a10-9330-310fcf692468\n\n[ipv4]\nmethod=disabled\n\n[ipv6]\nmethod=disabled\n"
```

One call to `gen_conf` can dump multiple keyfiles from the "NetworkManager" backend.


src/cmdline/mod.rs
```diff
@@ -281,6 +281,16 @@ pub struct CommonCustomizeConfig {
...
     pub network_keyfile: Vec<String>,
+    /// Nmstate file to generate NetworkManager keyfile for live & dest
+    ///
+    /// Configure networking using the generated NetworkManager keyfile.
+    /// Network settings will be applied in the live environment, including
+    /// when Ignition is run.  If installer is enabled via additional options,
+    /// network settings will also be applied in the destination system,
+    /// including when Ignition is run.
+    #[clap(long, value_name = "path")]
+    pub network_nmstate: Option<String>,
```
jlebon, line 292: For consistency with `--network-keyfile`, let's make this be a `Vec<String>` too instead?

src/cmdline/mod.rs
```diff
@@ -281,6 +281,16 @@ pub struct CommonCustomizeConfig {
     /// including when Ignition is run.
     #[clap(long, value_name = "path")]
     pub network_keyfile: Vec<String>,
+    /// Nmstate file to generate NetworkManager keyfile for live & dest
```
jlebon, line 284: It looks like it's typed "NMState".

src/live/customize.rs
```diff
@@ -169,6 +175,28 @@ impl LiveInitrd {
         Ok(())
     }
 
+    pub fn network_nmstate(&mut self, path: &str) -> Result<()> {
+        if !self.features.live_initrd_network {
+            bail!("This OS image does not support customizing network settings.");
+        }
+        let net_state_str = read_to_string(path)?;
+        match NetworkState::new_from_json(net_state_str.as_str()) {
+            Ok(net_state) => match net_state.gen_conf() {
```
jlebon, line 184: These two `match`es look like they could be collapsed to two `.context(...)?` instead to handle errors. That'll help unindent things.

Cargo.toml
```diff
@@ -71,6 +71,8 @@ url = ">= 2.1, < 3.0"
 uuid = { version = ">= 0.8, < 2.0", features = ["v4"] }
 walkdir = "^2.3"
 xz2 = "^0.1"
+#nmstate = "2.1.0"
+nmstate = { path = "../nmstate/rust/src/lib" }
```
jlebon, line 75: This looks like leftover from testing.

fixtures/customize/installer-test.json
```diff
@@ -0,0 +1,11 @@
+{
```
jlebon, line 1: Aren't NMState files YAML? Or does it support both YAML and JSON? (Or is this actually JSON being parsed as YAML?)

Would be good to add a YAML fixture too/instead.

jlebon (commented):
Cool, thanks! Assuming getting the deps right in Fedora isn't too much work, this SGTM. I think even if we end up doing https://github.com/coreos/fedora-coreos-tracker/issues/1175, this still has value because it allows configuring the network in the initramfs of the live system too using NMState files.

src/live/customize.rs
qinqon: Tried that first but the NmstateError is not implemented the expected trait.

Cargo.toml
qinqon: The nmstate rust lib is not published yet at cargo.io they are going to do so during this week I think. Meanwhile I point it there to my local clone, this is draft PR.

src/cmdline/mod.rs
qinqon: Official name is with just capital N, https://github.com/nmstate/nmstate/pull/224.

jlebon: Ahh interesting. Looks like the official website needs to be fixed then: https://nmstate.io/.

tests/images/customize.sh
```diff
@@ -58,6 +58,8 @@ opts_common=(
     --ignition-ca "${fixtures}/cert.pem"
     # Condition of @applied-live-ign@ and @applied-dest-ign@
     --network-keyfile "${fixtures}/installer-test.nmconnection"
+    # Condition of @applied-live-ign@ and @applied-dest-ign@
```
bgilbert, line 61: This is not true currently, so you'll need to update the `live.ign` and `dest.ign` fixtures.

src/cmdline/mod.rs
bgilbert: Once CI is working for this PR, it'll reject this line as being too long.  You can check manually with `tests/help.sh`.

src/cmdline/mod.rs
```diff
@@ -281,6 +281,16 @@ pub struct CommonCustomizeConfig {
...
+    ///
+    /// Configure networking using the generated NetworkManager keyfile.
```
bgilbert, line 286: Maybe let's say something like "using a NetworkManager keyfile generated from the specified Nmstate config"?

src/cmdline/mod.rs
```diff
@@ -281,6 +281,16 @@ pub struct CommonCustomizeConfig {
...
+    /// Network settings will be applied in the live environment, including
+    /// when Ignition is run.  If installer is enabled via additional options,
+    /// network settings will also be applied in the destination system,
+    /// including when Ignition is run.
+    #[clap(long, value_name = "path")]
+    pub network_nmstate: Option<String>,
+
```
bgilbert, line 293: Nit: stray blank line.

src/live/customize.rs
```diff
@@ -25,6 +25,9 @@ use crate::iso9660::{self, IsoFs};
 use super::embed::{INITRD_IGNITION_PATH, INITRD_NETWORK_DIR};
 use super::util::filename;
 
+use nmstate::NetworkState;
+use serde_json;
```
bgilbert, line 29: Let's put these in the first block with the other crate imports.

src/live/customize.rs
```diff
@@ -169,6 +175,28 @@ impl LiveInitrd {
         Ok(())
     }
 
+    pub fn network_nmstate(&mut self, path: &str) -> Result<()> {
+        if !self.features.live_initrd_network {
+            bail!("This OS image does not support customizing network settings.");
+        }
+        let net_state_str = read_to_string(path)?;
```
bgilbert, line 182: We generally avoid bare `?` operators when calling external libraries.  Let's add `context()` or `with_context()` calls that explain what we were doing at the time of the failure.

src/live/customize.rs
bgilbert: Yup, it [looks like](https://github.com/nmstate/nmstate/blob/base/rust/src/lib/error.rs) `NmstateError` doesn't implement `Error`.  Ideally it would do so, but if necessary you can wrap the error in `anyhow!()` here.

src/live/customize.rs
```diff
@@ -169,6 +175,28 @@ impl LiveInitrd {
...
+        match NetworkState::new_from_json(net_state_str.as_str()) {
+            Ok(net_state) => match net_state.gen_conf() {
+                Ok(nm_conf) => {
+                    let nm_conf_str = serde_json::to_string(&nm_conf)?;
+                    let name = filename(path)?;
+                    let path = format!("{}/{}.nmconnection", INITRD_NETWORK_DIR, name);
```
bgilbert, line 188: Hmm, won't this create something like `my-config.yaml.nmconnection`?

src/live/customize.rs
```diff
@@ -169,6 +175,28 @@ impl LiveInitrd {
...
+                    self.initrd.add(&path, nm_conf_str.as_bytes().to_vec());
```
bgilbert, line 189: Looks like we lost the overwrite check that `network_keyfile()` performs here.

bgilbert (commented):
This is great, thank you!

src/cmdline/mod.rs
qinqon: Can we enforce this with `cargo fmt` ?

bgilbert: The issue is that the short help text needs to be a single line of a certain maximum length, which varies depending on the width of the flags.  I'm not sure `cargo fmt` is configurable to that degree.

src/live/customize.rs
qinqon: The final PR will just use base name so the extension is skip and we get `my-config.nmconnection`.

src/cmdline/mod.rs
qinqon: Done

src/cmdline/mod.rs
qinqon: Done

src/live/customize.rs
qinqon: Going to try to make the right thing and put a PR at Nmstate so we have the proper NmstateError type there.

src/cmdline/mod.rs
qinqon: Done

src/cmdline/mod.rs
qinqon: Done

src/live/customize.rs
qinqon: Done

src/live/customize.rs
qinqon: Done

src/live/customize.rs
qinqon: Done

src/live/customize.rs
qinqon: Created PR at nmstate implementing the trait https://github.com/nmstate/nmstate/pull/1912

qinqon:
Testing part is not fixed yet, but code is reviewable now.

tests/images/customize.sh
qinqon: After changing them to add the condition for generate NetworkManager keyfile I have try to run the tests with 
```bash
./tests/images.sh ~/Documents/cnv/sandbox/coreos-live/
```
And the output is the following, looks like the checksums does not match.
```bash
Boot media will automatically install to /dev/vda without confirmation.
1189: digest inplace-iso
222: sha256sum inplace-iso
222: awk '{print $1}'
189: '[' e41247f55606742bd0083c20ffc53bd92b8001d015165614f5858c37efb7b015 = fb49271eda68f2fa7ce2c4727411e3bfd30f099a93a3784d7469d4dedd63e54d ']'
1: rm -rf /tmp/tmp.sNbyWtpoKh
```

Do you know what's going on here ?

src/cmdline/mod.rs
```diff
@@ -281,6 +281,15 @@ pub struct CommonCustomizeConfig {
     /// including when Ignition is run.
     #[clap(long, value_name = "path")]
     pub network_keyfile: Vec<String>,
+    /// Nmstate file for live & dest
```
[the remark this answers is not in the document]
qinqon: I would say `nmstate_file` since it's important to state that it's a file to the state not the state embedded at the argument.

src/live/customize.rs
```diff
@@ -169,6 +175,37 @@ impl LiveInitrd {
...
+            NetworkState::new_from_json(net_state_str.as_str()).context("parsing nmstate")?;
+        let nm_conf = net_state
+            .gen_conf()
+            .context("generating NM configuration from nmstate")?;
+        let nm_conf_str =
+            serde_json::to_string(&nm_conf).context("converting generated NM conf to string")?;
+        let net_state_path_basename = Path::new(path)
+            .file_stem()
+            .context("extracting nmstate path basename")
+            .unwrap();
```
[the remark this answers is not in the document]
qinqon: Adding "context" is not enough to have a proper return that we can pin point ?

src/live/customize.rs
```diff
@@ -169,6 +175,37 @@ impl LiveInitrd {
...
+        let path = format!(
+            "{}/{:?}.nmconnection",
```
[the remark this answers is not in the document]
qinqon: Converting the "file_stem" to str.

src/cmdline/mod.rs
```diff
@@ -281,6 +281,15 @@ pub struct CommonCustomizeConfig {
...
+    ///
+    /// Configure networking using a NetworkManager keyfile generated from the
+    /// specified Nmstate config Network settings will be applied in the live
```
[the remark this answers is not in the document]
qinqon: Done

src/cmdline/mod.rs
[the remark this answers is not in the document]
qinqon: Done

src/live/customize.rs
```diff
@@ -169,6 +175,37 @@ impl LiveInitrd {
...
+            INITRD_NETWORK_DIR, net_state_path_basename
+        );
+        if self.initrd.get(&path).is_some() {
+            bail!(
+                "nmstate config already specifies keyfile {:?}",
```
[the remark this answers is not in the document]
qinqon: Done

tests/images/customize.sh
```diff
@@ -265,6 +270,10 @@ coreos-installer pxe customize src-initrd -o initrd
     --network-keyfile "${fixtures}/installer-test.nmconnection" \
     --network-keyfile "${fixtures}/installer-test.nmconnection" 2>&1 ||:) |
     grepq "already specifies keyfile"
+(iso_customize \
+    --network-nmstate "${fixtures}/installer-test-nmstate.yaml" \
+    --network-nmstate "${fixtures}/installer-test-nmstate.yaml" 2>&1 ||:) |
```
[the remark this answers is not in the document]
qinqon: Done

tests/images/customize.sh
qinqon: I am going to proposa a PR a change some we can instruct it to generate the keyfile but without the UUID.

src/live/customize.rs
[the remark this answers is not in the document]
cgwalters: You can avoid the `to_str()` and `unwrap()` here by using [camino](https://crates.io/crates/camino) - we use it extensively in the ostree stack which hard requires UTF-8 pathnames.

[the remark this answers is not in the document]
qinqon: Done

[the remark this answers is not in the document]
bgilbert: @cgwalters Let's not add an additional dependency just to handle one call site.  Pathname conversion is a recurring pattern throughout the codebase.

@qinqon Apologies for the noise, but please revert the camino addition.  If we want to change the way we handle pathnames, we should do it comprehensively as a separate PR.

As to your question, adding `context` is the correct approach, but then the `unwrap()` panics if the conversion fails.  Please use `?` to return an error instead.

src/cmdline/mod.rs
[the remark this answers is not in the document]
bgilbert: I was suggesting that we use consistent terminology in the help text.  We're currently saying "Nmstate file" in the short help and "Nmstate config" in the long help.

As to the argument name, I think the original `--network-nmstate` (or maybe `--network-nmstate-file` if it fits) is clearer than the updated `--nmstate-file`, since the first one helpfully groups the network configuration arguments together.  However, I'd be interested in other people's thoughts here.

src/live/customize.rs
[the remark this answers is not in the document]
qinqon: No problem, at least I have learn about `camino` now, reverted and removed unwrap.

src/cmdline/mod.rs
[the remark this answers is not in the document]
qinqon: What about removing `nmstate` and keeping `--network-config-file` ? is similar to how it's done at openshift installer at the baremetal host [1].

[1] https://github.com/openshift/installer/blob/969378dc498336d1ac29baa89f75324a44467d5d/pkg/types/baremetal/platform.go#L43

qinqon:
Cargo.toml has now a pin to https://github.com/nmstate/nmstate/pull/1915, test should pass, but this cannot be merged until the PR is merge and nmstate published to cargo.io.

qinqon:
@bgilbert can you help me find why CI is failing ? I think all the bits are in place now (except the TODOs).

bgilbert:
@qinqon `live.ign` (and `dest.ign`) is checking for the existence of `installer-test-nmstate.nmconnection`, but the tests are no longer creating that file.

qinqon:
@bgilbert fixed `pr-merge` job, do we have to do something for `x390x` to work or this is a WIP job ?

qinqon:
closed/re-open by mistake.

bgilbert:
The s390x jobs broke recently and we're not sure why.  This change doesn't affect any s390x-specific code, so I think it's reasonable to ignore those jobs.

src/cmdline/mod.rs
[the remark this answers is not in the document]
qinqon: @bgilbert what is the best name for this ?

bgilbert:
FYI, the s390x jobs are working again, but you'll have to rebase to pick up the fix.

qinqon:
@bgilbert every issue is addressed now, it's ready for review.

qinqon:
/hold
Looks like nmstate gen_conf rust should return the keyfiles name as it was like that at python version, let's wait for that to consume it.

bgilbert:
@qinqon Is the nmstate Rust crate packaged in Fedora?  The `RPMs` test thinks it isn't, and I can't find it in the `nmstate` subpackages either.  To keep coreos-installer shippable downstream, we generally avoid landing PRs here until their dependencies are available in Fedora.

qinqon:
> @qinqon Is the nmstate Rust crate packaged in Fedora? The `RPMs` test thinks it isn't, and I can't find it in the `nmstate` subpackages either. To keep coreos-installer shippable downstream, we generally avoid landing PRs here until their dependencies are available in Fedora.

Not yet, they are working on it, ist just a matter of time.

bgilbert:
For the record, this PR increases the size of the release binary without debug symbols from 12.1 MiB to 15.0 MiB.

qinqon:
> For the record, this PR increases the size of the release binary without debug symbols from 12.1 MiB to 15.0 MiB.

Is it worth to investigate the nmstate side before merging this PR or we can do a follow up ?

bgilbert:
I was concerned about the number of additional dependencies we were pulling in, which is why I checked, but I'm personally not too worried about 3 MiB.  (Maybe I'm just used to Go binaries.)  If there are ways to reduce the size, that'd be great, but it can be done as a followup.

qinqon:
> I was concerned about the number of additional dependencies we were pulling in, which is why I checked, but I'm personally not too worried about 3 MiB. (Maybe I'm just used to Go binaries.) If there are ways to reduce the size, that'd be great, but it can be done as a followup.

I will prepare a follow up PR at nmstate using "features" from rust so we have "default" with "show" and "apply" and "gc" for "gen_conf" this we only stuff needed for a feature get compiled.

bgilbert (commented):
This looks great!  Just some nits.

Docs need to be regenerated with `make docs`.

src/cmdline/mod.rs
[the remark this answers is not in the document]
bgilbert: `--network-nmstate` sounds good, thanks. :+1: `--network-config-file` implies that a particular file format is the canonical one, which isn't the case here.

fixtures/customize/installer-test-nmstate-json-0.nmconnection
```diff
@@ -0,0 +1,6 @@
+[connection]
```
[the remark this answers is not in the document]
qinqon: We are not testing "nmstatectl gc" we are testing it's integration, but it's true that a bug at the integration can dump the wrong file, I can add a check for content too if needed.

Cargo.toml
```diff
@@ -71,6 +71,7 @@ url = ">= 2.1, < 3.0"
 uuid = { version = ">= 0.8, < 2.0", features = ["v4"] }
 walkdir = "^2.3"
 xz2 = "^0.1"
+nmstate = "2.1.1"
```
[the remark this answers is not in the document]
bgilbert: `nix` < `nmstate` :slightly_smiling_face:

fixtures/customize/installer-test-nmstate-json-0.nmconnection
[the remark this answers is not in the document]
bgilbert: Yeah, I don't think it's worth a lot of effort, but maybe the test could do `coreos-installer iso network extract -C` and then grep the configs for `nmstate-eth<n>`?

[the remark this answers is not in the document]
qinqon: At the end the nmstate team is going back to adding the generated nmconection file name to the gen_conf output (it was included on the python nmstate version) so using the path also verify partially the content, I think that's good enough.

qinqon:
@bgilbert addressed comments and adapted to latest nmstate so the keyfile name is generated by it, we have to wait for the new crate/rpm.

src/live/customize.rs
```diff
@@ -169,6 +175,35 @@ impl LiveInitrd {
...
+        let nm_connections = generated_conf
+            .get("NetworkManager")
+            .context("extracting NetworkManager generated config")?;
+        for (nm_con_file_name, nm_con_content) in nm_connections {
+            let nm_con_path = Path::new(INITRD_NETWORK_DIR).join(nm_con_file_name);
+            let nm_con_path_str = nm_con_path
+                .to_str()
+                .context("converting generated NetworkManager keyfile path to UTF-8")?;
+            if self.initrd.get(&nm_con_path_str).is_some() {
+                bail!("config already specifies keyfile {}", path);
```
bgilbert, line 198:
```suggestion
                bail!("config already specifies keyfile {}", nm_con_path_str);
```

bgilbert:
LGTM!  I think the remaining things are:

- [ ] Fix clippy lints
- [ ] nmstate 2.1.2 crate released and `Cargo.toml` updated
- [ ] nmstate Rust crate available in Fedora updates so RPMs CI is green

qinqon:
@bgilbert fixed clippy issues.

qinqon:
Hey @qinqon Looks like fedora 36 update is ready at https://bodhi.fedoraproject.org/updates/FEDORA-2022-b78be96657
But the coreos-install build RPM need to add requirements ` (crate(nmstate/default) >= 2.1.2 with crate(nmstate/default) < 3.0)`

bgilbert:
Rust package requirements are automatically added by `rust-srpm-macros`.

qinqon:
@bgilbert Do you know what happening with CI ?

bgilbert:
The MSRV tests should be fixed by https://github.com/coreos/coreos-installer/pull/926.  The RPM test is awaiting [FEDORA-2022-b78be96657](https://bodhi.fedoraproject.org/updates/FEDORA-2022-b78be96657) going stable.

bgilbert:
I'm not seeing that ENOSPC error in other PRs.

bgilbert:
Could you try rebasing?  That should fix the GitHub Actions checks, and we can see whether it fixes the kola failure too.  If it doesn't, there may be an actual problem with the PR.

jlebon:
I think @bgilbert and I came to the same realization in parallel. :) I suspect CI is hitting `ENOSPC` because `rdcore` includes debuginfo and we're not using LTO in debug mode so it's including nmstate. But even without this change, I think we should probably strip the binary when installing it in debug mode like CI does. Did that in https://github.com/coreos/coreos-installer/pull/929.

bgilbert:
We're testing with 1 GB of RAM, and this PR increases the `rdcore` size from 132 to 178 MiB.

(Actually we're not using LTO at all anymore; it was too buggy.)

bgilbert:
Confirmed that https://github.com/coreos/coreos-installer/pull/929 fixes the `ENOSPC`.

qinqon:
> We're testing with 1 GB of RAM, and this PR increases the `rdcore` size from 132 to 178 MiB.
> 
> (Actually we're not using LTO at all anymore; it was too buggy.)

Do we have to do something at nmstate to reduce `rdcore` size ? or 178 is ok ?

bgilbert:
No, the rdcore binary is only about 2 MiB with all symbols stripped, or 6 MiB with debug symbols stripped.  That's what we actually ship in the OS, so this is only a CI problem.  We'll just land #929 and then you can rebase on top of it.

bgilbert:
Okay, if you rebase this on `main` it should fix CI.

bgilbert (approved):
Thanks for all your work on this! :tada: :balloon: :fireworks:

PR #870, opened 2022-05-30
Cargo.toml: move s390x-specific deps near others

jlebon:
It's odd to have s390x deps be listed after `dev-dependencies`. Have it
be just after `dependencies` instead since it's the same kind.


jlebon:
This is a tiny cleanup, but my reason for doing this is that then my hack to compile test s390x-specific code as x86_64 on my machine Just Works (`git ls-files | xargs sed -i -e '/target_arch.*s390x/ d'`).

bgilbert (approved):

Issue #872, opened 2022-06-01
s390x CI failing


bgilbert:
CI is failing early in `cargo build`:

```
+ podman exec -w /workspace -e PATH=/root/.cargo/bin:/usr/sbin:/usr/bin:/sbin:/bin d5e8071cc867186dbbad89a5c4b1a2dc09b77656b970894e58eb5c9971b7ed26 bash --noprofile --norc -eo pipefail /ctx/script-coIDbw
    Updating crates.io index
/ctx/script-coIDbw: line 1:   301 Killed                  cargo build
Error: Process completed with exit code 137.
```

It's not immediately clear why.  It happens on the pinned toolchains, so it's not a Rust 1.61 issue, and it happens if we downgrade CI to the Fedora 35 container, so it's not a Fedora 36 issue.  The podman version in the `ubuntu-latest` image has not changed.

PR #873, opened 2022-06-03
workflows: disable fail-fast on Rust jobs

bgilbert:
Keep CI usable even though the s390x jobs are failing (https://github.com/coreos/coreos-installer/issues/872).

If this lands, I'll also mark the s390x jobs as non-required.


jlebon (approved):

jlebon:
I'm going to merge this because I want to merge https://github.com/coreos/coreos-installer/pull/851 but the "Rust / Tests, stable toolchain (x86_64)" test keeps being cancelled because the s390x one failed and fail-fast is on.

PR #878, opened 2022-06-07
workflows: disable cargo bundled libgit to avoid s390x OOM

bgilbert:
For now, shell out to git commands instead.

Fixes https://github.com/coreos/coreos-installer/issues/872.  Works around https://github.com/rust-lang/cargo/issues/10583.


bgilbert:
Re-enabled marking s390x tests as required.

jlebon (approved):
Nice find!

PR #888, opened 2022-06-16
templates/release-checklist: delete all `.a` files from vendor dir

bgilbert:
We've gained some additional Windows-specific dependencies, and `vcpkg` also has compiled libraries for Darwin and iOS.  Our vendor tarball should never need to include compiled static libs, so delete them all.


jlebon (approved):
Makes sense to me!

PR #904, opened 2022-06-30
workflows: fix installation of `qemu-s390x-static`

bgilbert:
We install `qemu-s390x-static` by downloading and extracting a single Fedora binary package into an Ubuntu system, without respecting package dependencies.  As a consequence, when `qemu-user-static` was converted to a metapackage for https://github.com/coreos/fedora-coreos-tracker/issues/1088, we stopped actually installing the emulator.  Use the correct subpackage instead.

I've opted not to revert c3b3d49d4195c6016d3bececbf1c8012f21bc020 for now.  By leaving fail-fast disabled, we avoid wasting resources for successful jobs when we rerun failed ones after a flake, and make it easier to get CI signal if the s390x jobs break again in the future.  The downside is some CI resource waste when a PR has correlated job failures.


prestist (approved):
LGTM!

PR #906, opened 2022-06-30
rdcore: Use collect instead of manual loop

cgwalters:
I happened to be have this code open for an unrelated reason
and noticed that this open-coded loop can be replaced by a `collect`.
We can already generate hashmap entries from a `(key, value)` tuple,
and `collect()` also knows how to generate a `Result<T>` which
it will know to do from type inference.


bgilbert:
@cgwalters Rebasing should fix CI.

PR #907, opened 2022-06-30
rootmap: Use `split_once`

cgwalters:
This combinator style avoids an unnecessary `Vec` allocation and
also the implicit panic possible on array indexing.

### DIFF
--- a/src/bin/rdcore/rootmap.rs
+++ b/src/bin/rdcore/rootmap.rs
@@ -135,11 +135,9 @@ fn mdadm_detail(device: &Path) -> Result<HashMap<String, String>> {
 }
 
 fn split_mdadm_line(line: &str) -> Result<(String, String)> {
-    let v: Vec<&str> = line.splitn(2, '=').collect();
-    if v.len() != 2 {
-        bail!("invalid mdadm line: {}", line);
-    }
-    Ok((v[0].into(), v[1].into()))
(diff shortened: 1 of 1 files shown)


bgilbert:
@cgwalters Rebasing should fix CI.

prestist (approved):
TIL "combinator && mdadm". 
Very clean. LGTM!

bgilbert:
@cgwalters There's not actually an implicit panic in the old code, right?  We're checking the length before making assumptions about it.

cgwalters:
> @cgwalters There's not actually an implicit panic in the old code, right? We're checking the length before making assumptions about it.

Yes, LLVM probably figures out how to omit the check here, but the implicit ability to panic is still there.  (Today AIUI there's no MIR level optimizations for this, so the panics are only omitted in release builds, which is why things like https://lib.rs/crates/dont_panic exist)

bgilbert:
This is a nice cleanup, thanks.  :+1:

> LLVM probably figures out how to omit the check here, but the implicit ability to panic is still there.

If I understand correctly, you're saying the compiler may emit unreachable panic code that isn't optimized away?  Even if so, we're neither size- nor performance-critical here, so that seems harmless.  (Put a different way, I don't think we should avoid array indexing just because it might panic if we misuse it.)

Issue #869, opened 2022-05-30
Give clearer error message when ISO file is not complete


sshnaidm:
# Bug #

Misleading error message when using not fully downloaded rhcos ISO image.
I haven't download the image fully, it's only 512M size. When I'm running `coreos-installer iso kargs modify -a "ip=dhcp"` command on this ISO, I get the message:
```
Error: reading kargs embed area

Caused by:
    0: reading 1132 bytes at 1082311624
    1: failed to fill whole buffer
```

## Host Operating System Version ##
RHEL 8.2
## coreos-installer Version ##
0.14, 0.11
## Expected Behavior ##
Message about `Error: reading kargs embed area`

## Actual Behavior ##

Probably would be better to detect that the image is wrong and can't be modified.

## Reproduction Steps ##
Start download ISO of RHCOS, stop it in the middle and run the command:
`$ coreos-installer iso kargs modify -a "ip=dhcp" rhcos-411.85.202205101201-0-live.x86_64.iso`

jlebon:
I think one of the base issues is that `seek` doesn't error out by default if seeking past the end of the file. There's `SEEK_DATA` which would (and then we could catch `ENXIO` and give a nicer error), but we'd need to look beyond `std::io` for it (it also doesn't seem supported on every filesystem, though all the popular ones at least). We could also wrap seeks intended for subsequent reads with a method that also does a follow-up `read` of one byte.

jlebon:
Hmm, actually another simpler check we can do upfront is to just get the offset of the last file in the ISO and verify that `file offset + file length <= ISO length`.

travier:
Yes, I was thinking we could do a basic "is this file large enough?" check to quickly root out such issues.

jlebon:
Related: https://github.com/coreos/coreos-installer/issues/644

If `coreos-installer download` were supported for OCP, we'd correctly verify checksums.

travier:
@bgilbert is this fixed by https://github.com/coreos/coreos-installer/pull/908?

bgilbert:
Yes, it is.

PR #908, opened 2022-06-30
embed.rs: Added error check when ISO file incomplete

RishabhSaini:
Follows on [Issue#869](https://github.com/coreos/coreos-installer/issues/869) and Jira [cos-1556](https://issues.redhat.com/browse/COS-1556) 
Giving clearer error message when ISO file is not complete
Implements jlebon's [suggestion](https://github.com/coreos/coreos-installer/issues/869#issuecomment-1141265911)
`last file's offset + file length <= ISO length`
The new error message displayed when incomplete ISO file detected is: `Error: ISO Image is wrong and can't be modified`


bgilbert:
Thanks for pursusing this!

I don't think the karg embed area parser is the right place to put this check, because any other code that reads data from the ISO could encounter the same problem.  Instead, let's do this validation when initially parsing the ISO.

Jonathan's suggestion requires reading all of the ISO image's directories, and per the above we'd do it every time we parsed an ISO.  Instead, I'd suggest comparing the Volume Space Size field in the [Primary Volume Descriptor](https://wiki.osdev.org/ISO_9660#The_Primary_Volume_Descriptor) to the file size, which should save a bunch of I/O.  That wouldn't detect the case where an invalid ISO image claims to have files beyond the Volume Space Size, but we're only trying to catch a truncated download, not arbitrary data corruption, so that should be okay.  Note that the actual file size will likely be larger than the Volume Space Size, because `isohybrid` adds an EFI partition table after the end of the filesystem.

CI is failing because you need to run `rustfmt`.

RishabhSaini:
Currently this is what the Primary Volume Descriptor returns
`PrimaryVolumeDescriptor { system_id: "LINUX", volume_id: "rhcos-412.86.202206301758-0", root: Directory { name: ".", address: Address(29), length: 2048 } }`
To access the field `VolumeSpaceSize`, I have tried to read the specific bytes by using the address(29) as offset and adding 80 which is the offset of the VolumeSpaceSize as mentioned in the ISO9660 wiki. Then using read_exact to read the length 8 bytes from the offset. 
Is this the correct method to get the `VolumeSpaceSize` field from PrimaryVolumeDescriptor?

bgilbert:
The address is measured in sectors, not bytes.  (Filesystem data structures often do this, since they point to things at sector boundaries.  The lower bits of the address would be wasted otherwise.)

coreos-installer has an existing parse function that creates the `PrimaryVolumeDescriptor` struct.  The in-memory struct doesn't include every field in the on-disk structure, since we don't use most of them.  However, since we're now going to care about an additional field, we should add it to the in-memory struct and to the parse function.

src/iso9660.rs
```diff
@@ -298,13 +313,22 @@ impl PrimaryVolumeDescriptor {
             parse_iso9660_string(eat(buf, 1), 32, IsoString::StrA).context("parsing system id")?;
         let volume_id = // technically should be StrD, but non-compliance is common
             parse_iso9660_string(buf, 32, IsoString::StrA).context("parsing volume id")?;
-        let root = match get_next_directory_record(eat(buf, 156 - 72), 34, true)? {
+        eat(buf, 8); // Unused field always 0x00
+        let volume_space_size = // Data type: int32_LSB-MSB
```
bgilbert, line 317: The comment isn't really necessary, since it's clear from context.  I'm okay with keeping it, but let's put it at the end of the statement, not after the `=`.

src/iso9660.rs
```diff
@@ -298,13 +313,22 @@ impl PrimaryVolumeDescriptor {
...
+            buf.get_i32_le();
```
bgilbert, line 318: In [ISO9660 nomenclature](https://wiki.osdev.org/ISO_9660#Numerical_formats), `int32` is unsigned.

src/iso9660.rs
```diff
@@ -49,6 +49,19 @@ pub struct IsoFs {
 impl IsoFs {
     pub fn from_file(mut file: fs::File) -> Result<Self> {
         let descriptors = get_volume_descriptors(&mut file)?;
+        match &descriptors[0] {
+            VolumeDescriptor::Boot(_boot_volume_descriptor) => {}
+            VolumeDescriptor::Primary(primary_volume_descriptor) => {
+                if i64::from(primary_volume_descriptor.volume_space_size)
+                    * i64::from(primary_volume_descriptor.logical_block_size)
+                    > file.metadata()?.len().try_into().unwrap()
```
bgilbert, line 57: Since the `volume_space_size` should be unsigned, these will be `u64` and the `try_into().unwrap()` goes away.

src/iso9660.rs
```diff
@@ -49,6 +49,19 @@ pub struct IsoFs {
...
+                {
+                    bail!("ISO Image is wrong and can't be modified");
```
bgilbert, line 59: "wrong" isn't very specific; let's say what the problem actually is.  And we don't know that the caller's intent is to modify the ISO; they might just be reading it.

bgilbert:
Also, let's update `test_primary_volume_descriptor`, and maybe add a test based on `open_iso()` that verifies that we correctly detect a truncated image.

bgilbert:
Oh, also, we've switched to updating `docs/release-notes.md` from individual PRs, so let's add a note there.

src/iso9660.rs
```diff
@@ -298,13 +306,16 @@ impl PrimaryVolumeDescriptor {
             parse_iso9660_string(eat(buf, 1), 32, IsoString::StrA).context("parsing system id")?;
         let volume_id = // technically should be StrD, but non-compliance is common
             parse_iso9660_string(buf, 32, IsoString::StrA).context("parsing volume id")?;
-        let root = match get_next_directory_record(eat(buf, 156 - 72), 34, true)? {
+        eat(buf, 8); // Unused field always 0x00
+        let volume_space_size = u64::from(buf.get_u32_le());
```
bgilbert, line 310: A bit simpler:
```suggestion
        let volume_space_size = buf.get_u32_le() as u64;
```

src/iso9660.rs
```diff
@@ -567,8 +579,23 @@ mod tests {
         IsoFs::from_file(iso_file).unwrap()
     }
 
+    fn open_truncated_iso() -> Result<IsoFs, Error> {
```
bgilbert, line 582: Unlike `open_iso()`, this doesn't need to be a helper function, since it's only needed by one test.  So let's make this function a `#[test]` and do the `unwrap_err()` and comparison here, and drop the `assert_eq!` from `test_primary_volume_descriptor`.

src/iso9660.rs
```diff
@@ -567,8 +579,23 @@ mod tests {
...
+
     #[test]
     fn test_primary_volume_descriptor() {
+        assert_eq!(
+            open_truncated_iso().unwrap_err().to_string(),
+            "ISO image is incomplete"
+        );
         let iso = open_iso();
         let desc = iso.get_primary_volume_descriptor().unwrap();
         assert_eq!(desc.system_id, "system-ID-string");
```
bgilbert, line 601: We should assert the value of `volume_space_size` also.  `isoinfo -d` says it should be 338.  I think it's okay to hardcode assumptions about the test fixture; we can always update the test if it changes.

docs/release-notes.md
```diff
@@ -12,6 +12,7 @@ Major changes:
 Minor changes:
 
 - Add release notes to documentation
+- Added an error check for detecting an incomplete ISO image
```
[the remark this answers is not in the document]
bgilbert: Maybe "iso: Improve error for incomplete ISO files"? We were detecting it before, just with an obscure message.

src/iso9660.rs
```diff
@@ -48,8 +48,15 @@ pub struct IsoFs {
 
 impl IsoFs {
     pub fn from_file(mut file: fs::File) -> Result<Self> {
+        let length = file.metadata()?.len();
         let descriptors = get_volume_descriptors(&mut file)?;
-        Ok(Self { descriptors, file })
+        let iso_fs = Self { descriptors, file };
+        let primary = iso_fs.get_primary_volume_descriptor()?;
+        if primary.volume_space_size * ISO9660_SECTOR_SIZE as u64 > length {
```
cgwalters, line 55: I think in the case of a corrupted ISO this multiplication could overflow?  Probably worth something like:

```suggestion
    if primary.volume_space_size.checked_mul(ISO9660_SECTOR_SIZE as u64).ok_or_else(|| anyhow!("Invalid ISO volume size"))? > length 
```
or so?

bgilbert: `volume_space_size` is 32 bits on disk, so this can't overflow.

bgilbert (approved):
Looks great!

PR #909, opened 2022-07-01
rootmap: handle filesystems with LUKS integrity

nikita-dubrovskii:
Systems where some filesystems were luks2-encrypted with 'integrity' option have special
crypt-integrity devices:
```
NAME                 FSTYPE      FSVER LABEL        UUID
vda
|-vda3               crypto_LUKS 2     crypt_bootfs f55add4a-1351-4699-9e61-32f33a68a031
| `-crypt_bootfs_dif
|   `-crypt_bootfs   ext4        1.0   boot         3d4ed0e4-40ba-4930-99ff-35d2be907ec3
|-vda4               crypto_LUKS 2     crypt_rootfs 7c5341bf-9616-4ce1-90dc-111b03379799
| `-crypt_rootfs_dif
|   `-crypt_rootfs   xfs               root         d0bc5f93-f2ac-42f3-8b16-27b0794266c5
```

So to get 'uuid' of such devices, we have to recursively go to the underlying block device.
crypt-integrity devices on the other hand should be skipped during 'rdcore rootmap'.

### DIFF
--- a/src/bin/rdcore/rootmap.rs
+++ b/src/bin/rdcore/rootmap.rs
@@ -108,7 +108,11 @@ fn device_to_kargs(root: &Mount, device: PathBuf) -> Result<Option<Vec<String>>>
     if blktype.starts_with("raid") || blktype == "linear" {
         Ok(Some(get_raid_kargs(&device)?))
     } else if blktype == "crypt" {
-        Ok(Some(get_luks_kargs(root, &device)?))
+        if Disk::new(&device)?.is_luks_integrity()? {
+            Ok(None)
+        } else {
+            Ok(Some(get_luks_kargs(root, &device)?))
+        }
     } else if blktype == "part" || blktype == "disk" || blktype == "mpath" {
         Ok(None)
     } else {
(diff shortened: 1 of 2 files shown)


src/bin/rdcore/rootmap.rs
```diff
@@ -216,9 +207,26 @@ fn get_luks_name(device: &Path) -> Result<String> {
...
-        .trim()
-        .into())
+    // The LUKS UUID is a property of the backing block device of *this* block device, so we have
+    // to get its parent. In case of crypt-integrity device we have to recursively go to the
+    // underlying block device
+    let deps = get_blkdev_deps(device)?;
+    match deps.len() {
+        0 => bail!("missing parent device for {}", device.display()),
+        1 => {
+            let device = deps[0].as_path();
```
cgwalters, line 217: I know this isn't a new issue, but we should be able to match like this instead which is a bit cleaner IMO:

```
match deps.as_slice() {
  [] => bail!("missing parent device for {}", device.display()),
  [device] => { ... }
  _ => bail!("found multiple...")
}
```

src/blockdev.rs
```diff
@@ -166,22 +167,32 @@ impl Disk {
...
 
-    fn is_dm_device(&self) -> Result<bool> {
-        let canon_path = Path::new(&self.path)
-            .canonicalize()
-            .with_context(|| format!("canonicalizing {}", &self.path))?;
-        // convert back to &str because .starts_with on a Path doesn't mean the same thing
-        let canon_path = canon_path
-            .to_str()
-            .with_context(|| format!("path {} is not UTF-8", canon_path.display()))?;
-        Ok(canon_path.starts_with("/dev/dm-"))
```
jlebon, line 184: We still need this part since I think the comment there still applies (see https://doc.rust-lang.org/std/path/struct.Path.html#method.starts_with).

src/bin/rdcore/rootmap.rs
```diff
@@ -35,9 +35,13 @@ pub fn rootmap(config: RootmapConfig) -> Result<()> {
     let mut backing_devices = get_blkdev_deps_recursing(&device)?;
     backing_devices.push(device);
 
-    // for each of those, convert them to kargs
+    // for each of those (except crypt-integrity), convert them to kargs
     let mut kargs = Vec::new();
     for backing_device in backing_devices {
+        if Disk::new(&backing_device)?.is_luks_integrity()? {
+            continue;
+        }
```
jlebon, line 43: I think this would fit better as part of `device_to_kargs` where we would then return an `Ok(None)` like we do for some other no-op cases already. E.g. in the `else if blktype == "crypt" {` block, we can check if the device is `CRYPT-INTEGRITY`.

src/bin/rdcore/rootmap.rs
```diff
@@ -143,27 +147,14 @@ fn split_mdadm_line(line: &str) -> Result<(String, String)> {
 }
 
 fn get_luks_kargs(root: &Mount, device: &Path) -> Result<Vec<String>> {
-    // The LUKS UUID is a property of the backing block device of *this* block device, so we have
-    // to get its parent. This is a bit awkward because we're already iterating through parents, so
-    // theoretically we could re-use the same state here. But meh... this is easier to understand.
```
jlebon, line 148: Can we keep the rest of this comment in the new location?

src/blockdev.rs
nikita-dubrovskii: No, i've made `path` non public, so it's canonicalized in `new()`

src/bin/rdcore/rootmap.rs
```diff
@@ -35,7 +35,7 @@ pub fn rootmap(config: RootmapConfig) -> Result<()> {
     let mut backing_devices = get_blkdev_deps_recursing(&device)?;
     backing_devices.push(device);
 
-    // for each of those, convert them to kargs
+    // for each of those (except crypt-integrity), convert them to kargs
```
jlebon, line 38: I don't think we need that comment edit now.

jlebon (approved):
Minor nits, but LGTM as is too!

jlebon (approved):
Thanks!

PR #920, opened 2022-07-14
Support reading zstd-compressed initrds

bgilbert:


src/io/zstd.rs
```diff
@@ -0,0 +1,130 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Implementation of an API similar to zstd::stream::read::Decoder using
+// zstd::stream::raw::Decoder.  We need this because read::Decoder returns
+// io::ErrorKind::Other if there's trailing data after an xz stream, which
```
lucab, line 17: `after a zstd stream` perhaps?

lucab (approved):

src/io/zstd.rs
```diff
@@ -0,0 +1,130 @@
...
+// can't be disambiguated from an actual error.  By using the low-level API,
+// we can check zstd::stream::raw::Status.remaining to see whether the
+// decoder thinks it's at the end of a frame, check the upcoming bytes for
+// the magic number of another frame, and decide whether we're done.  The
+// raw decoder always stops at frame boundaries, so this is reliable.  If
+// done, return Ok(0) and allow the caller to decide what it wants to do
+// about trailing data.
```
jlebon, line 24: Have you considered opening a PR upstream to support this in the higher-level API?

jlebon (approved):
Nice!

src/io/zstd.rs
bgilbert: Hmm.  In zstd there are two parts to EOF detection: noticing we're at the end of a frame and checking whether another frame is coming.  Upstream could do the first part, but I don't think the second part can be done robustly without something like `PeekReader`, which the upstream API can't require.  And it seems awkward to provide a high-level API that requires the caller to manually read ahead and check magic numbers.  Allowing trailing data is a pretty obscure requirement, really, so sticking to the low-level API seems fine.

bgilbert:
Minor updates for comments and to avoid hardcoding the magic numbers.

PR #929, opened 2022-07-20
Makefile: strip `rdcore` DWARF in debug mode

jlebon:
When testing in debug mode, we don't want to copy the full ~180M
`rdcore` binary in the initramfs to avoid taking too much space. In
release mode, we do want it because it gets split into the debuginfo
package anyway.

Tweak things so that by default in debug mode, we strip the binary. Do
it in a separate step instead of using `install --strip` because we only
want to strip DWARF info, but by default the full symbol table is
stripped and there's no way to customize it.

This will have an effect on backtraces in case of crashes/panics.
Function names will still be printed, but filenames and line numbers
will be missing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,16 @@ RDCORE ?= 1
 ifeq ($(RELEASE),1)
 	PROFILE ?= release
 	CARGO_ARGS = --release
+	STRIP_RDCORE ?= 0
 else
 	PROFILE ?= debug
 	CARGO_ARGS = --features mangen
+	# In debug mode (most often used by devs/CI), we default to stripping
+	# `rdcore` because it's otherwise huge and in kola's default 1G VMs can
+	# cause ENOSPC. In release mode (most often used by Koji/Brew), we don't do
+	# this because the debuginfo gets split out anyway. In either profile,
+	# we allow overriding the default.
+	STRIP_RDCORE ?= 1
 endif
 ifeq ($(RDCORE),1)
 	CARGO_ARGS := $(CARGO_ARGS) --features rdcore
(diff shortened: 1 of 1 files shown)


bgilbert:
Does retaining the function names with `-g` affect any error output?  I don't think we enable `RUST_BACKTRACE` in the initrd.

Also, it feels like the Makefile is the wrong place to handle this.  The semantics would be: `make install` doesn't strip debug symbols in release mode but does strip them in debug mode, for rdcore and not coreos-installer.  That seems highly non-obvious.

The reason we're not stripping symbols in release mode is that splitting/removing them is downstream's responsibility.  I think that should be true for debug mode as well.  There's nothing fundamentally wrong with putting debug symbols in a development initrd; it's just that our specific CI setup doesn't give the VMs enough RAM for that.  So to me it makes more sense to leave the Makefile as is, and have `.cci.jenkinsfile` perform the stripping needed by CoreOS CI.

jlebon:
> Does retaining the function names with `-g` affect any error output?

It doesn't.

> I don't think we enable `RUST_BACKTRACE` in the initrd.

We don't, but a developer might do so to debug something. Given that the symbol table doesn't really take much space and it's the debug build, it doesn't seem worth being overly aggressive.

> Also, it feels like the Makefile is the wrong place to handle this. The semantics would be: `make install` doesn't strip debug symbols in release mode but does strip them in debug mode, for rdcore and not coreos-installer. That seems highly non-obvious.
> 
> The reason we're not stripping symbols in release mode is that splitting/removing them is downstream's responsibility. I think that should be true for debug mode as well. There's nothing fundamentally wrong with putting debug symbols in a development initrd; it's just that our specific CI setup doesn't give the VMs enough RAM for that. So to me it makes more sense to leave the Makefile as is, and have `.cci.jenkinsfile` perform the stripping needed by CoreOS CI.

It's not just CI though. Kola defaults to 1G and so local developers will hit this as well. I'd prefer to not have devs have to remember to either strip the binary by hand and keep the default `make && make install DESTDIR=/srv/fcos/overrides/rootfs` working like in other projects. I guess we could document requiring to pass `--qemu-memory` to bump up the default when spawning VMs/running tests? Though that's awkward and error-prone too.

I agree the behaviour is non-obvious, though it seems like an acceptable tradeoff.

If we do decide to offload the decision to strip to a higher level, I'd still prefer to have the logic live in the Makefile and have it be opt-in.

jlebon:
We could also have `make install` print a notice if `DESTDIR` ends with `/overrides/rootfs` and `RDCORE_STRIP` wasn't set.

bgilbert:
Yeah, point taken.  Can we add a comment to the Makefile explaining why we're doing this?  It seems highly non-obvious for someone who's reading the Makefile without that context.

jlebon:
Done! Also tweaked it so that the default can be overridden from the CLI either way.

bgilbert (approved):

PR #932, opened 2022-07-27
Fix corrupt `iso extract minimal-iso` output to stdout; add tests

bgilbert:


src/live/mod.rs
```diff
@@ -634,13 +634,13 @@ pub fn iso_extract_minimal_iso(config: IsoExtractMinimalIsoConfig) -> Result<()>
         .context("creating temporary file")?;
     data.unxzpack(full_iso.as_file()?, &mut outf)
         .context("unpacking miniso")?;
-    outf.seek(SeekFrom::Start(0))
-        .context("seeking back to start of miniso tempfile")?;
 
     modify_miniso_kargs(outf.as_file_mut(), config.rootfs_url.as_ref())
         .context("modifying miniso kernel args")?;
 
     if &config.output == "-" {
```
prestist, line 641: What does the presence of a "-" or absence of one mean ?

bgilbert: Specifying `-` as an input or output file is a common Unix command-line convention for stdin or stdout.

bgilbert: (But it's not magic; the program has to implement it explicitly.)

docs/release-notes.md
```diff
@@ -13,6 +13,7 @@ Major changes:
 Minor changes:
 
 - Detect truncated xz archives
+- iso extract minimal-iso: Fix corrupt output when writing to stdout
```
jlebon, line 16: Minor: WDYT about

```suggestion
- iso: Fix corrupt `extract minimal-iso` output when writing to stdout
```

for consistency?

(Came up while reviewing this PR with @prestist.)

jlebon (approved):
LGTM! Thanks a lot for adding tests.

prestist (approved):
Could not find anything, poked @jlebon to look at it with him. Def a LGTM! good job!

bgilbert:
Rebased, updated changelog, and added tests for running `iso extract minimal-iso` on a customized ISO.

jlebon (approved):

Issue #816, opened 2022-04-06
Partition saving doesn't notice MBR partitions


bgilbert:
# Bug #

## Host Operating System Version ##

Any

## Target Operating System Version ##

Any

## coreos-installer Version ##

All releases through 0.13.1

## Expected Behavior ##

`coreos-installer install --save-partindex` correctly saves partitions on disks that have an MBR partition table and no GPT.

## Actual Behavior ##

`coreos-installer install --save-partindex` clobbers such partitions.

## Reproduction Steps ##

1. Use `fdisk` to create a partition on a blank target disk
2. Install to the disk, passing `--save-partindex` to save the partition

## Other Information ##

`--save-partlabel` isn't affected because MBR doesn't support partition labels.

We could support this via mbrman.  We'd need to detect and ignore protective MBRs, and to fail if the MBR otherwise doesn't match a valid GPT.

prestist:
In speaking with @bgilbert we are going to be taking this issue with a 2 pronged approach

1. Mitigation reduce risk of data loss from unexpected clobbering of saved partitions. 
2. Migrate MBR saved partitions to GPT format so the installer can effectively save them.

PR #934, opened 2022-07-29
blockdev: fail `install --save-partindex` on MBR-formatted disks 

prestist:
`install --save-partindex` and `install --save-partlabel` both search the GPT for the specified partitions. While MBR doesn't have partition labels, it does have partition indexes. Users might try to use `--save-partindex` with an MBR disk, resulting in data loss when coreos-installer doesn't notice the specified partitions.

To address this, we'll need to detect saved MBR partitions and convert them to GPT partitions in the output. As a first step, if `--save-partindex` is specified and the disk has an MBR but no GPT, fail install to avoid data loss.

Fixes https://github.com/coreos/coreos-installer/issues/816.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -21,6 +21,7 @@ Minor changes:
 - Add release notes to documentation
 - iso: Detect incomplete ISO files
 - Warn if console kargs could have used `--console`/`--dest-console` instead
+- install: fail if `--save-partindex` is specified on an MBR disk
 
 Internal changes:
 
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -570,6 +570,16 @@ impl SavedPartitions {
         let gpt = match GPT::find_from(disk) {
             Ok(gpt) => gpt,
             Err(gptman::Error::InvalidSignature) => {
+                // ensure no indexes are listed to be saved from a MBR disk
+                // we don't need to check for labels since MBR does not support them
+                if filters
+                    .iter()
+                    .any(|f| matches!(f, PartitionFilter::Index(_, _)))
+                    && disk_has_mbr(disk).context("checking if disk has an MBR")?
+                {
+                    bail!("saving partitions from an MBR disk is not yet supported");
+                }
+
                 // no GPT on this disk, so no partitions to save
                 return Ok(Self {
                     sector_size,
@@ -1103,6 +1113,14 @@ pub fn get_gpt_size(file: &mut (impl Read + Seek)) -> Result<u64> {
     Ok(gpt.header.first_usable_lba * gpt.sector_size)
 }
 
+fn disk_has_mbr(file: &mut (impl Read + Seek)) -> Result<bool> {
+    let mut sig = [0u8; 2];
+    file.seek(SeekFrom::Start(510))
+        .context("seeking to MBR signature")?;
+    file.read_exact(&mut sig).context("reading MBR signature")?;
+    Ok(sig == [0x55, 0xaa])
+}
+
 pub fn udev_settle() -> Result<()> {
     // "udevadm settle" silently no-ops if the udev socket is missing, and
     // then lsblk can't find partition labels.  Catch this early.
@@ -1469,7 +1487,7 @@ mod tests {
             let saved = SavedPartitions::new_from_file(&mut base, 512, filter).unwrap();
             let mut disk = make_unformatted_disk();
             saved.overwrite(&mut disk).unwrap();
-            assert!(disk_has_mbr(&mut disk), "test {}", testnum);
+            assert!(disk_has_mbr(&mut disk).unwrap(), "test {}", testnum);
             let result = GPT::find_from(&mut disk).unwrap();
             assert_eq!(
                 get_gpt_size(&mut disk).unwrap(),
@@ -1481,7 +1499,7 @@ mod tests {
             let mut disk = make_disk(512, &merge_base_parts);
             saved.merge(&mut image, &mut disk).unwrap();
             assert!(
-                disk_has_mbr(&mut disk) == !expected_blank.is_empty(),
+                disk_has_mbr(&mut disk).unwrap() == !expected_blank.is_empty(),
                 "test {}",
                 testnum
             );
@@ -1515,7 +1533,7 @@ mod tests {
             let saved =
                 SavedPartitions::new_from_file(&mut disk, *sector_size as u64, &vec![]).unwrap();
             saved.overwrite(&mut disk).unwrap();
-            assert!(disk_has_mbr(&mut disk), "{}", *sector_size);
+            assert!(disk_has_mbr(&mut disk).unwrap(), "{}", *sector_size);
             disk.seek(SeekFrom::Start(0)).unwrap();
             let mut buf = vec![0u8; *sector_size + 1];
             disk.read_exact(&mut buf).unwrap();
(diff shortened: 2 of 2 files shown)


src/blockdev.rs
```diff
@@ -1452,12 +1460,13 @@ mod tests {
 
         // test merging with unformatted initial disk
         let mut disk = make_unformatted_disk();
-        let saved = SavedPartitions::new_from_file(&mut disk, 512, &vec![label("z")]).unwrap();
-        let mut disk = make_disk(512, &merge_base_parts);
-        saved.merge(&mut image, &mut disk).unwrap();
-        let result = GPT::find_from(&mut disk).unwrap();
-        assert_partitions_eq(&merge_base_parts, &result, "unformatted disk");
-
+        let err = SavedPartitions::new_from_file(&mut disk, 512, &vec![label("z")]).unwrap_err();
```
prestist, line 1463: This is an expectation change, in the past, we would say there is no GPT, and then forget the list of saved partitions. 

Now that we are reacting to no GPT or MBR, we get in a state where we dont want to proceed. 

The user thinks there are partitions to save, but the filesystem is not known, as a result we cannot continue, and keep the promise of saving the partitions. so we bail, does that makes sense?

src/blockdev.rs
```diff
@@ -1531,12 +1539,13 @@ mod tests {
...
+        assert!(
+            format!("{:#}", err).contains(&gptman::Error::InvalidSignature.to_string()),
+            "incorrect error: {:#}",
+            err
+        );
+
```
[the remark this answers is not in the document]
prestist: Since we are not including mbrman console, I think we can get away with just the protective mbr for this step.

docs/release-notes.md
```diff
@@ -21,6 +21,7 @@ Minor changes:
 - Add release notes to documentation
 - iso: Detect incomplete ISO files
 - Warn if console kargs could have used `--console`/`--dest-console` instead
+- save-partindex: bail when listed partindex's are located on a MBR disk
```
bgilbert, line 24: The prefix should be the subcommand, `install`.  And this implies that we continue if the MBR doesn't contain any of the listed partindexes, which isn't true.  If any partition saving is requested, and the disk is MBR, we bail.

src/blockdev.rs
```diff
@@ -519,6 +519,13 @@ pub struct SavedPartitions {
     partitions: Vec<(u32, GPTPartitionEntry)>,
 }
 
+fn disk_has_mbr(disk: &mut File) -> bool {
+    let mut sig = [0u8; 2];
+    disk.seek(SeekFrom::Start(510)).unwrap();
+    disk.read_exact(&mut sig).unwrap();
```
bgilbert, line 525: `unwrap` is only suitable if a failure is a programmer error, or in tests.  These should use `context()?` and return `Result<bool>`.

src/blockdev.rs
```diff
@@ -519,6 +519,13 @@ pub struct SavedPartitions {
     partitions: Vec<(u32, GPTPartitionEntry)>,
 }
 
+fn disk_has_mbr(disk: &mut File) -> bool {
```
bgilbert, line 522: Nit: I'd probably move this next to `get_gpt_size` and make the argument signature match the one in that function.  Strictly speaking we don't need a `File`.

src/blockdev.rs
```diff
@@ -571,10 +578,14 @@ impl SavedPartitions {
             Ok(gpt) => gpt,
             Err(gptman::Error::InvalidSignature) => {
                 // no GPT on this disk, so no partitions to save
-                return Ok(Self {
-                    sector_size,
-                    partitions: Vec::new(),
-                });
+                if disk_has_mbr(disk) {
+                    bail!("Unsupported: Cannot save partitions from a MBR disk.")
+                } else {
```
bgilbert, line 583: We've just bailed in the `true` case, so the `else` is unnecessary and adds indentation.

src/blockdev.rs
```diff
@@ -571,10 +578,14 @@ impl SavedPartitions {
             Ok(gpt) => gpt,
             Err(gptman::Error::InvalidSignature) => {
                 // no GPT on this disk, so no partitions to save
-                return Ok(Self {
-                    sector_size,
-                    partitions: Vec::new(),
-                });
+                if disk_has_mbr(disk) {
+                    bail!("Unsupported: Cannot save partitions from a MBR disk.")
```
bgilbert, line 582: Nit:
```suggestion
                    bail!("saving partitions from an MBR disk is not yet supported")
```

src/blockdev.rs
```diff
@@ -1320,6 +1331,15 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_gpt_disk_not_flagged_as_mbr() {
```
bgilbert, line 1335: Hmm, can you elaborate on what this is testing for?  It doesn't write a protective MBR, which we would do in real life (in `overwrite`).  And it doesn't actually matter what the function does in isolation, since we only call it if we've found no GPT.

I think instead `test_saved_partitions` should have a section that verifies that we bail if there's an MBR but no GPT.  `GPT::write_protective_mbr_into` might be a good way to construct this.

src/blockdev.rs
prestist: Thats super cool, did not know I could do that!

src/blockdev.rs
prestist: Yeah you are right, It was a misunderstanding on what the new_from did, I assumed(sigh i need to not do that) that the new_from was doing more then just creating a gpt header

prestist: And yeah that makes sense, I will.

docs/release-notes.md
prestist: Ah, ok. thank oyu

docs/release-notes.md
```diff
@@ -21,6 +21,7 @@ Minor changes:
 - Add release notes to documentation
 - iso: Detect incomplete ISO files
 - Warn if console kargs could have used `--console`/`--dest-console` instead
+- install: bail when `--save-partindex` is specified on a MBR disk
```
bgilbert, line 24: The release notes and commit message both refer to `--save-partindex` specifically, but the code also activates if `--save-partlabel` is specified.  We should decide which approach we want to take.

bgilbert (commented):
Some nits and one open question.

The commit message is better, but I think we could still improve it a bit.  While issue links should always be included, the message should be written as though the reader hasn't read the issue.  Let's include a brief description of the problem we're trying to solve.

docs/release-notes.md
bgilbert, line 24:
```suggestion
- install: fail if `--save-partindex` is specified on an MBR disk
```

src/blockdev.rs
```diff
@@ -1549,6 +1567,31 @@ mod tests {
             err
         );
 
+        // test trying to save partitions from a MBR disk
+        let mut disk = make_unformatted_disk();
+        gptman::GPT::write_protective_mbr_into(&mut disk, 512).unwrap();
+        // label only
+        SavedPartitions::new(&mut disk, 512, &vec![label("*i*")]).unwrap();
+
```
bgilbert, line 1575: Stray blank line.

bgilbert (commented):
LGTM generally.

bgilbert (approved):

bgilbert:
:tada:

PR #950, opened 2022-08-11
rdcore: zipl: add support for multiple '--append-karg'

nikita-dubrovskii:
This allows such invocation to be possible:
```
rdcore zipl --append-karg=ignition.firstboot --append-karg=lockdown=confidentiality --secex-mode=enforce --boot-mount=/boot
```

Required for:
https://github.com/coreos/coreos-assembler/pull/3002

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -19,6 +19,7 @@ Minor changes:
 - iso: Detect incomplete ISO files
 
 Internal changes:
+- rdcore: Rename `--kargs` to `--append-karg` in `zipl` subcommand and support passing it multiple times
 
 
 Packaging changes:
@@ -243,18 +244,18 @@ Major changes:
 Minor changes:
 
 - blockdev: Fix RHEL `lsblk` ordering [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1916502) by using `--nodeps` option
-- blockdev: Strengthen device mapper path detection 
+- blockdev: Strengthen device mapper path detection
 
 Internal changes:
 
-- osmet: Drop support for `--real-rootdev` option 
+- osmet: Drop support for `--real-rootdev` option
 - Add `--override-options` to `rdcore kargs` to make it easier to test kernel argument changes
 - Optionally create a file if kernel arguments are modified
 - Add declarative semantics for kernel argument modification
 
 Packaging changes:
 
-- Switch from `error-chain` to `anyhow` library 
+- Switch from `error-chain` to `anyhow` library
 
 
 ## coreos-installer 0.8.0 (2021-01-12)
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -145,9 +145,9 @@ pub struct ZiplConfig {
     #[clap(long, value_name = "HOSTKEY")]
     pub hostkey: Option<String>,
 
-    /// Extra kargs
-    #[clap(long, value_name = "KARGS")]
-    pub kargs: Option<String>,
+    /// Append kernel argument
+    #[clap(long, value_name = "KARG")]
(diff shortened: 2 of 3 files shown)


jlebon:
Bikeshed: should we instead switch it to be more consistent with `rdcore kargs` and `coreos-installer install`? I.e. have the switch be `--append-karg` (singular, with `append` in its name). We should be free to make this change given that `rdcore zipl` isn't yet used anywhere in production.

If it helps ratcheting it in, we can keep the old `--kargs` as a hidden alias for now and remove it in a future release.

docs/release-notes.md
```diff
@@ -17,6 +17,7 @@ Minor changes:
 - Fix unlikely decompression error reading initrd
 - Add release notes to documentation
 - iso: Detect incomplete ISO files
+- rdcore: Support multiple `--append-karg` option for `zipl` command
```
jlebon, line 20: I think `rdcore` changes are usually listed under "Internal changes".

jlebon: Also, maybe something like:

> rdcore: Rename `--kargs` to `--append-karg` in `zipl` subcommand and support passing it multiple times

?

src/bin/rdcore/kargs.rs
```diff
@@ -99,7 +99,7 @@ pub fn zipl(config: ZiplConfig) -> Result<()> {
         boot.mountpoint(),
         config.hostkey,
         config.rootfs,
-        config.kargs,
+        config.append_karg.as_ref().map(|v| v.join(" ")),
```
jlebon, line 102: This is not new, but for a follow-up: note that this function uses the kargs passed here in two ways: the first is in the SE path to create the sdboot blob in which it is appended to the regular kargs, the second is in the non-SE path to edit a temporary BLS entry in which we use `KargsEditor::append_if_missing`, which is not quite the same as the declared semantics of `--append-karg` (notice how for example `rdcore kargs` has both `--append` and `--append-if-missing`).

My suggestion is to:
1. pass in the `config.kargs` as a slice here
2. in the SE path, use `KargsEditor::append` when generating the blob
3. in the non-SE path, use `KargsEditor::append` for these kargs, and `append_if_missing` for `ignition.firstboot` only

src/bin/rdcore/cmdline.rs
```diff
@@ -146,8 +146,8 @@ pub struct ZiplConfig {
     pub hostkey: Option<String>,
 
     /// Extra kargs
```
jlebon, line 148: Should we change this to e.g. `Append kernel argument` for clarity and consistency with `rdcore kargs`?

src/bin/rdcore/kargs.rs
nikita-dubrovskii: i'm not sure it's worth doing in this small PR, cause all of those look like a rework. Also in SE-path usage of `KargsEditor` would look like an overhead:
1) mirror `/boot/loader/entries` to `/tmp/smth`
2) append `kargs` to `tmp *.conf` 
3) read `kargs` for `genprotimg`

jlebon: `KargsEditor` operates directly on the string so doesn't require doing the `/tmp` hack.

jlebon (approved):

src/bin/rdcore/kargs.rs
nikita-dubrovskii: Ok, will re-work a bit `zipl` code in ongoing patches

jlebon:
> If it helps ratcheting it in, we can keep the old `--kargs` as a hidden alias for now and remove it in a future release.

We do need this because current pipelines building non-SE images rely on it. Done in https://github.com/coreos/coreos-installer/pull/968.

PR #964, opened 2022-08-23
s390x: don't fail if luks isn't used

nikita-dubrovskii:
This brings support for recent move from `LUKS2` to `dm-verity` introduced by https://github.com/coreos/coreos-assembler/pull/3039.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -21,7 +21,7 @@ Minor changes:
 
 Internal changes:
 - rdcore: Rename `--kargs` to `--append-karg` in `zipl` subcommand and support passing it multiple times
-
+- Only open BLS configs in write mode when modifying
 
 Packaging changes:
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -22,6 +22,7 @@ Minor changes:
 Internal changes:
 - rdcore: Rename `--kargs` to `--append-karg` in `zipl` subcommand and support passing it multiple times
 - Only open BLS configs in write mode when modifying
+- s390x: Don't fail and skip initrd update when there is no LUKS2 keys and config
 
 Packaging changes:
 
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -150,20 +150,29 @@ fn generate_sdboot(
         .write_all(options.as_bytes())
         .context("writing zipl se cmdline")?;
 
-    let mut lukskeys_path = "/etc/luks/".to_string();
-    let mut crypttab_path = "/etc/crypttab".to_string();
-    let mut hostkeys_path = "/etc/se-hostkeys/".to_string();
-
     // during cosa-build rootfs includes ostree commit path
-    if let Some(ref rootfs) = rootfs {
-        lukskeys_path.insert_str(0, rootfs);
-        crypttab_path.insert_str(0, rootfs);
-        hostkeys_path.insert_str(0, rootfs);
-    }
+    let ostree_commit = if let Some(ref rootfs) = rootfs {
+        Path::new(rootfs)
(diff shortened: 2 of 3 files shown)


docs/release-notes.md
```diff
@@ -21,7 +21,7 @@ Minor changes:
 
 Internal changes:
 - rdcore: Rename `--kargs` to `--append-karg` in `zipl` subcommand and support passing it multiple times
-
+- io: open bls configs 'ro' by default
```
jlebon, line 24: `io` is only meaningful if one know the codebase well. Maybe something like:

> Only open BLS configs in write mode when modifying

?

jlebon: Also, can you add some background for this change in the commit message, maybe with a link to the dm-verity cosa PR?

src/s390x/zipl.rs
```diff
@@ -150,20 +150,29 @@ fn generate_sdboot(
...
-    let mut crypttab_path = "/etc/crypttab".to_string();
-    let mut hostkeys_path = "/etc/se-hostkeys/".to_string();
-
     // during cosa-build rootfs includes ostree commit path
-    if let Some(ref rootfs) = rootfs {
-        lukskeys_path.insert_str(0, rootfs);
-        crypttab_path.insert_str(0, rootfs);
-        hostkeys_path.insert_str(0, rootfs);
-    }
+    let ostree_commit = if let Some(ref rootfs) = rootfs {
```
jlebon, line 154: Aside: can't we drop the `--rootfs` arg from the `rdcore zipl` command now? I think all instances where this code is called, it's done from within the deployment root.

But maybe better to keep it for now to make the switch to the new dm-verity approach easier.

src/s390x/zipl.rs
```diff
@@ -150,20 +150,29 @@ fn generate_sdboot(
...
+    };
+    let lukskeys_path = ostree_commit.join("etc/luks");
+    let crypttab_path = ostree_commit.join("etc/crypttab");
+    let hostkeys_path = ostree_commit.join("etc/se-hostkeys");
+
     // new initrd with LUKS keys & config
-    let mut luks_files = find_files(&lukskeys_path, |e: &DirEntry| Ok(e.metadata()?.is_file()))?;
-    luks_files.push(PathBuf::from(crypttab_path));
-    let initrd = generate_initrd(&initrd, &luks_files, rootfs)?;
+    let new_initrd = if lukskeys_path.exists() {
```
jlebon, line 164: Also would be good to add a commit message with rationale in the commit message introducing this hunk.

jlebon (approved):
Thanks! Some minor comments, but overall LGTM.

src/s390x/zipl.rs
nikita-dubrovskii: i'd wait with dropping til dm-verity replaces LUKS  in both FCOS & RHCOS

src/s390x/zipl.rs
```diff
@@ -150,20 +150,29 @@ fn generate_sdboot(
...
+        let mut luks = find_files(&lukskeys_path, |e: &DirEntry| Ok(e.metadata()?.is_file()))?;
+        luks.push(crypttab_path);
+        Some(generate_initrd(&initrd, &luks, rootfs)?)
+    } else {
+        None
+    };
+
+    let initrd = match new_initrd {
```
jlebon, line 172: Minor: this is `.map().unwrap_or()` I think.

src/s390x/zipl.rs
jlebon, line 154: And here we can use `.as_deref().unwrap_or()`.

jlebon (approved):

PR #968, opened 2022-08-26
rdcore/zipl: alias old `--kargs` to new `--append-karg`

jlebon:
We need to keep providing the old alias here to make the transition
easier in cosa.

Note the semantics are a little different (before, this was an
`Option<String>`, now it's an `Option<Vec<String>>`), but it still works
for what we need here.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -21,7 +21,7 @@ Minor changes:
 
 Internal changes:
 
-- rdcore: Rename `--kargs` to `--append-karg` in `zipl` subcommand and support passing it multiple times
+- zipl: Deprecate `--kargs` in favor of new `--append-karg`, which supports multiple instances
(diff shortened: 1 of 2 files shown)


jlebon:
I'm marking this as `skip-notes` because I think the notes added in https://github.com/coreos/coreos-installer/pull/950 are still sufficient, but happy to tweak it to mention that the old switch is still valid.

bgilbert:
That release note item could probably use some cleanup anyway.  Maybe something like:

```
- zipl: Deprecate `--kargs` in favor of new `--append-karg`, which supports multiple instances
```

(Yeah, it's not technically correct as written, but I guess that's how we want people to think of it.)

jlebon:
Updated!

bgilbert (approved):

PR #969, opened 2022-08-30
rdcore: add '--append-file' for 'zipl' subcommand

nikita-dubrovskii:
In SE case we append `luks2` keys to original `initrd`. It may be also useful to append some more user-defined files to `sdboot` image.


bgilbert (approved):
LGTM

PR #970, opened 2022-09-01
Bump clippy to 1.63.0; fix lints

bgilbert:


prestist (approved):
LGTM, 

Do we have a issue # we can link? 

Also why are we skipping notes?

bgilbert:
clippy lints don't affect functionality; they don't even show up during a normal compile.  Since they're pretty trivial and not user-facing, we don't normally file issues for them and we don't normally release-note them.

PR #976, opened 2022-09-07
Generate and install example install config file

bgilbert:
If the `docgen` feature is enabled, add a subcommand to walk the `InstallConfig` struct and generate an example config file.  Use it to regenerate `customizing-install.md` and a new `example-config.yaml` when rebuilding docs.

Install `example-config.yaml` in `/usr/share/coreos-installer` for use by cosa, which can automatically add feature flags to the live image for each config option supported by the installer.

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -169,6 +169,9 @@ pub enum PackCmd {
     /// Generate man pages for coreos-installer
     #[cfg(feature = "docgen")]
     Man(PackManConfig),
+    /// Generate example config file for install subcommand
+    #[cfg(feature = "docgen")]
+    ExampleConfig(PackExampleConfigConfig),
 }
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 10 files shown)


src/cmdline/mod.rs
```diff
@@ -167,8 +167,11 @@ pub enum PackCmd {
     /// Pack a minimal ISO into a CoreOS live ISO image
     MinimalIso(PackMinimalIsoConfig),
     /// Generate man pages for coreos-installer
-    #[cfg(feature = "mangen")]
+    #[cfg(feature = "docgen")]
     Man(PackManConfig),
+    /// Generate example config file for install subcommand
+    #[cfg(feature = "docgen")]
+    ExampleConfig(PackExampleConfigConfig),
```
lucab, line 174: Looking at the CI failure, was this enum variant supposed to be `ConfigFormat` instead?

bgilbert: Whoops, that was an incomplete refactor.  Should be fixed now.

lucab (approved):

PR #977, opened 2022-09-08
Add `install --console` and `... customize --dest-console` to configure GRUB and kernel console

bgilbert:
Once CoreOS stops enabling the serial console by default, users will need a way to enable it themselves.  This really only affects bare metal, since other platforms have a consistent console configuration, so coreos-installer is a good place to handle it.  Users can already pass `console=` kargs, but we need a way to specify GRUB commands, ideally without having to actually use GRUB syntax.

Add a `--console` option that accepts kernel `console=` syntax, since that's already familiar.  Parse the option and convert it to both a console karg and GRUB commands, using the existing mechanism for GRUB command substitution.  If `--platform` is also specified, have `--console` override that platform's default console settings.  If `--console` is specified more than once, aggregate the console settings, following the multiple-console semantics for GRUB (use both consoles) and the kernel (use all consoles, with the last console primary).

There are serial console features only supported by GRUB (stop bits) or by the kernel (RTS/CTS flow control).  Support the common subset and reject the rest.  Also, round-trip console kargs to a less-compact but semantically equivalent format, since explicit is good and it saves some code complexity.

Document this as "bootloader console" rather than "GRUB console" to allow for future bootloaders other than GRUB.

Add the corresponding `--dest-console` option to the `customize` commands.  Use the `installer-config-directives` `features.json` object introduced by https://github.com/coreos/coreos-assembler/pull/3083 to check whether the image's coreos-installer is new enough to support `--console`.  In the future we can potentially support a `--live-console` or combined `--console` customize option that also affects the live kernel and bootloader (#979), but defer that for now.


bgilbert:
Ready for review!

bgilbert:
CI is failing because `registry.ci.openshift.org/coreos/coreos-assembler:latest` doesn't have https://github.com/coreos/coreos-assembler/pull/3083 yet.

dustymabe:
Overall strategy seems sane to me without reviewing the code directly.

One question I have.. Should we consider giving the user a warning if they try to set the kernel argument for console directly so that existing users get pointed towards this new (better) feature?

man/coreos-installer-install.8
```diff
@@ -56,6 +56,11 @@ Override the Ignition platform ID
 
 Install a system that will run on the specified cloud or virtualization platform, such as "vmware".
 .TP
+\fB\-\-console\fR=\fIspec\fR
+Kernel and bootloader console
+
+Set the kernel and bootloader console, using the same syntax as the parameter to the "console=" kernel argument.
```
lucab, line 62: Silly doubt: is the user-specified value expected to include the `console=` part or not (I think the answer is "no")? Should we try to handle that explicitly?

lucab (commented):
Code LGTM.
I  am both scared by the regex monster and by your newly acquired knowledge on serial oddities across Linux and GRUB :smile: 
On a more serious note, I really only have a general question about UX.

dustymabe:
> One question I have.. Should we consider giving the user a warning if they try to set the kernel argument for console directly so that existing users get pointed towards this new (better) feature?

And by "set the kernel argument for console directly" I mean either in the provided Ignition config (won't work for remote Ignition references, but that's OK), or through `coreos-installer` features like `--append-karg`.

man/coreos-installer-install.8
bgilbert: It isn't.  I've added a specific error message for this case.

bgilbert:
Updated!

> Should we consider giving the user a warning if they try to set the kernel argument for console directly so that existing users get pointed towards this new (better) feature?
>
> And by "set the kernel argument for console directly" I mean either in the provided Ignition config (won't work for remote Ignition references, but that's OK), or through coreos-installer features like --append-karg.

I've added a warning for `install --append-karg` and `iso/pxe customize --dest-karg-append`.  Checking Ignition kargs doesn't seem worth it, though.  We couldn't check merged/replaced configs as you say, couldn't detect kernel arguments specified via OCP MachineConfig, and the user may not have the flexibility to switch from Ignition kargs to `--console` (for example, because the Ignition config was generated by a tool).

lucab (approved):

src/cmdline/console.rs
```diff
@@ -0,0 +1,347 @@
+// Copyright 2022 Red Hat, Inc.
```
jlebon, line 1: This module is both terrifying and great, which is about what we expected I think. :laughing:

Really nice work!

PR #506, opened 2021-03-25
ci: adapt to new buildroot image

jlebon:
Similar to https://github.com/coreos/ignition/pull/1182 as a result of
https://github.com/coreos/fedora-coreos-config/pull/740.

PR #999, opened 2022-09-21
*: ShellCheck fixes

travier:
They look benign but likely need careful review.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1989551

### DIFF
--- a/scripts/coreos-installer-service
+++ b/scripts/coreos-installer-service
@@ -18,7 +18,7 @@ INSTALLER_CONFIG_DIR=/etc/coreos/installer.d
 
 args=("install")
 
-cmdline=( $(</proc/cmdline) )
+IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
 karg() {
     local name="$1" value="$2"
     for arg in "${cmdline[@]}"; do
@@ -30,7 +30,8 @@ karg() {
 }
 
 karg_bool() {
-    local value=$(karg "$@")
+    local value
+    value=$(karg "$@")
     case "$value" in
         ""|0|no|off) return 1;;
         *) return 0;;
@@ -41,8 +42,9 @@ copy_arg() {
     local arg="$1"; shift
     local opt="$1"; shift
 
-    local value="$(karg "${arg}")"
-    if [ ! -z "${value}" ]; then
+    local value
+    value="$(karg "${arg}")"
+    if [ -n "${value}" ]; then
         args+=("${opt}" "${value}")
     fi
 }
@@ -54,7 +56,7 @@ fi
 
 # Config files
 have_config_file=
-for file in ${INSTALLER_CONFIG_DIR}/*; do
+for file in "${INSTALLER_CONFIG_DIR}/"*; do
     if [ -f "${file}" ]; then
         args+=("--config-file" "${file}")
         have_config_file=1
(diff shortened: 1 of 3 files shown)


tests/images.sh
```diff
@@ -50,7 +51,14 @@ call_for_fixtures() {
 if [ -n "${1:-}" ]; then
     # test with artifacts in cosa build dir
     basedir="$1"
-    if ! [ -e "${basedir}"/*.iso ] ;then
+    found=false
+    for f in "${basedir}"/*.iso; do
+        if [[ -e "${f}" ]]; then
```
bgilbert, line 56: This could be reversed to "if doesn't exist" so `found` can be dropped.

bgilbert (commented):
LGTM generally.

tests/images.sh
travier: How would that work? If there is not file then we don't enter the for loop.

scripts/coreos-installer-service
```diff
@@ -126,7 +128,7 @@ args+=("--fetch-retries" "infinite")
 udevadm settle
 
 # Install
-echo "coreos-installer ${args[@]}"
+echo "coreos-installer " "${args[@]}"
```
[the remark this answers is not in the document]
travier: That was an incorrect fix. Updated.

tests/images.sh
```diff
@@ -2,8 +2,9 @@
 
 set -euo pipefail
 
-export PATH="$(realpath $(dirname $0)/../target/${PROFILE:-debug}):$PATH"
-fixturesdir="$(realpath $(dirname $0)/../fixtures)"
+export PATH
+PATH="$(realpath "$(dirname "$0")/../target/${PROFILE:-debug}"):$PATH"
```
bgilbert, line 6: It looks like this line is failing in CI and I don't understand why.

jlebon: I found at least [one other CI](https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/blue/rest/organizations/jenkins/pipelines/coreos-installer/branches/PR-1028/runs/4/nodes/223/steps/226/log/?start=0) run for a different PR where this failed. But I think it was ignored exactly because of what ShellCheck wants us to fix here (splitting `export` from setting).

I think the issue is that `target/` isn't present because since https://github.com/coreos/coreos-installer/pull/506, we build in a separate pod from where we build FCOS. We don't actually need to do this though for the CoreOS CI case because `coreos-installer` is [now installed](https://github.com/coreos/coreos-installer/blob/2c08dd141a4dfbc3306733bc6b7845fcc439d73a/.cci.jenkinsfile#L17-L18) in the cosa pod.

It's useful for local development though. So one solution is to change these lines to be conditional on `target/` existing and otherwise assume we want to test the installed binary. Or alternatively if we want to be more strict, add a knob (e.g. env var) to explicitly request testing installed binaries and use it from `.cci.jenkinsfile`.

bgilbert: Ohh, good catch.  Yeah, the env var approach seems best for preventing surprises, along with a message saying that we're testing the installed copy.

tests/images.sh
travier: Hum indeed!

tests/images.sh
bgilbert: @travier, looks like this CI failure wasn't addressed in your last push.  Is this PR still on your plate?

travier: I did not address it indeed

travier: Will try to get to it soon

travier: Should be fixed now

bgilbert: Merging as-is, and will address this in a follow-up.

bgilbert (approved):

tests/images.sh
bgilbert: Followup in https://github.com/coreos/coreos-installer/pull/1125.

PR #1012, opened 2022-10-06
osmet: avoid install performance regression in debug mode on Rust 1.64+

bgilbert:
Debug builds on Rust 1.64+ were seeing a substantial performance hit in osmet unpack, causing the `kola testiso` `pxe-offline-install` scenario to time out.  This is due to an apparent Rust performance regression when `BufReader`'s buffer is larger than the blocks returned by the inner reader, if the code is built without optimization.  In our case, osmet's `write_unpacked_image()` produces data in 8 KiB chunks, which are passed verbatim through the pipe to the 256 KiB `PeekReader` in `write_image()`.

For now, work around this by having osmet consolidate writes through a 256 KiB `BufWriter`.  The extra buffer is harmless, but once the affected Rust versions are no longer being used for development, it probably makes sense to drop this code.

xref https://github.com/rust-lang/rust/issues/102727

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -10,6 +10,8 @@ Major changes:
 
 Minor changes:
 
+- install: Avoid osmet performance regression in debug builds on Rust 1.64+
+
 Internal changes:
 
 Packaging changes:
--- a/src/osmet/unpacker.rs
+++ b/src/osmet/unpacker.rs
@@ -14,7 +14,7 @@
 
 use std::ffi::OsStr;
 use std::fs::{File, OpenOptions};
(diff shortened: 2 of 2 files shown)


cgwalters:
FWIW it's possible to do "debug mode for app, optimize dependencies" which is often a sane tradeoff.

https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#overrides

cgwalters (approved):

jlebon (approved):
Nice investigation!

bgilbert:
> FWIW it's possible to do "debug mode for app, optimize dependencies" which is often a sane tradeoff.

Yup, we're already doing that for the decompressors.  I tried it here too, but in this case it appears we'd need to optimize the app itself.

PR #1030, opened 2022-10-24
workflows: cancel superseded Rust and RPM jobs

bgilbert:
s390x tests are expensive, and when we're merging a lot of Dependabot PRs, CI can get backed up running redundant tests on main.  We don't need to continue running tests that have been superseded by new code, either in main or in a PR branch, so add concurrency configuration to cancel them.

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -8,6 +8,11 @@ on:
 permissions:
   contents: read
 
+# don't waste job slots on superseded code
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test-rpm-build:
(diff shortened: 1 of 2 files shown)


prestist (approved):
Great add! LGTM

travier:
I'm not sure if it's truly an issue, but I've added `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` to my workflows that uses the same config for push & pr jobs: https://github.com/travier/quay-containerfiles/blob/main/.github/workflows/consul.yml#L30

bgilbert:
In this case we primarily need the cancellation for `main`, and the cancellation in PRs is a nice side benefit.  Without it, merging half a dozen Dependabot PRs can cause CI to be backed up for hours.

travier:
OK so in this case your config should be correct.

Issue #431, opened 2020-12-17
Network timeout while low-level formatting ECKD DASD


bgilbert:
`low-level format` takes time, for 20Gb disk about 2 minutes on my machine, so after those 2 minutes all opened network connections are closed, and `installer` exits with failure:

```
Caused by: Broken pipe (os error 32)
```

_Originally posted by @nikita-dubrovskii in https://github.com/coreos/coreos-installer/pull/430#discussion_r544948811_

PR #1031, opened 2022-10-27
install: low-level format ECKD DASD before initiating fetch

bgilbert:
Normally we want to check the install image before starting to do I/O to the destination disk.  However, low-level formatting an ECKD DASD can take a long time, and if we initiate download of the install image first, the TCP connection will sit without any bytes being read and may eventually time out.  Avoid this by doing the low-level format before starting the fetch.

Fixes https://github.com/coreos/coreos-installer/issues/431.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -11,6 +11,7 @@ Major changes:
 Minor changes:
 
 - install: Avoid osmet performance regression in debug builds on Rust 1.64+
+- install: Avoid network fetch timeout when low-level formatting ECKD DASD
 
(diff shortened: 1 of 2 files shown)


bgilbert:
@nikita-dubrovskii, could you test this when you get a chance?  It's not urgent.

nikita-dubrovskii:
LGTM! 

thx for fixing it.

nikita-dubrovskii:
Did test on zVM + DASD

prestist (approved):
LGTM

PR #1033, opened 2022-10-29
WIP: io: use Sequoia for signature verification

bgilbert:
Shelling out to GPG is awkward, brittle, subject to changes in GPG's output, and requires GPG to be installed locally.  Sequoia was originally GPLv2-licensed and wasn't mature enough, but it's now LGPLv2 and further along.  Try switching to it.

- This currently uses the Nettle backend, which requires `clang-devel` and `nettle-devel` at build time and libnettle at runtime.  We probably can't require that in OCP where users just download the binary.  Sequoia has a [WIP OpenSSL backend](https://gitlab.com/sequoia-pgp/sequoia/-/merge_requests/1361) which will hopefully work for us.

- Stderr reporting needs improvement.  We should match the content (not necessarily the format) of GPG.

- This will need a careful audit.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -52,7 +52,7 @@ jobs:
           path: target/debug
           key: deps-${{ runner.os }}-${{ matrix.arch }}-${{ env.INSTALLED_TOOLCHAIN }}
       - name: Install dependencies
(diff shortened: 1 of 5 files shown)


bgilbert:
This isn't a priority at the moment, so I'll close for now.

PR #1038, opened 2022-10-31
cargo: bump to nmstate 2.2.3 with only `gen_conf` feature

bgilbert:
nmstate 2.2.0 split its functionality into multiple features, which allows us to reduce the set of dependencies we pull in.  nmstate 2.2.3 restored the MSRV to 1.58.0 so we can use it.

xref https://github.com/coreos/coreos-installer/issues/1032#issuecomment-1296979952

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -36,7 +36,7 @@ jobs:
         run: |
           mkdir rpms
           make -f .copr/Makefile srpm outdir=rpms
-          mock --rebuild rpms/*.src.rpm
+          mock --enablerepo updates-testing --rebuild rpms/*.src.rpm
           find /var/lib/mock -wholename '*/result/*.rpm' | xargs mv -t rpms
       - name: Archive RPMs
         uses: actions/upload-artifact@v3
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,25 +24,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2cb2f989d18dd141ab8ae82f64d1a8cdd37e0840f73a406896cf5e99502fab61"
 
 [[package]]
-name = "async-io"
-version = "1.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5e18f61464ae81cde0a23e713ae8fd299580c54d697a35820cfd0625b8b0e07"
-dependencies = [
- "concurrent-queue",
- "futures-lite",
- "libc",
- "log",
- "once_cell",
- "parking",
- "polling",
- "slab",
- "socket2",
- "waker-fn",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


bgilbert:
nmstate 2.2.0 [currently requires](https://github.com/coreos/coreos-installer/issues/1032#issuecomment-1297586357) Rust 1.60.

cathay4t:
Is rust 1.60 a problem? I through it is widely available(even RHEL 8.7 will ship 1.62 very soon).

bgilbert:
coreos-installer currently needs to support Rust 1.58 because that's what ships in RHEL 8.6 and 9.0.  If nmstate wants to bump to 1.60, I think that's fine; we just won't be able to land this PR for a while.  (Also, you may want to consider setting `package.rust-version` in `Cargo.toml`, as documentation.)

cathay4t:
https://github.com/nmstate/nmstate/pull/2085 crated to add `rust-version` to `Cargo.toml`. Thanks!

cathay4t:
@bgilbert The RHEL 8.7/9.1 has released with 1.62/1.64. Both EPEL8 and EPEL9 build target has it now. Can we try again?

bgilbert:
@cathay4t No, because the RHCOS development branches are currently based on RHEL 8.6 and 9.0.

cgwalters:
I don't think we need to block getting releases into Fedora onto *also* getting them into OCP 4.13/9.0.  We can also do the builds for C9S and ensure our code lands and is tested in 9.2 now.

bgilbert:
We keep coreos-installer buildable in RHCOS so that if we discover that we need to ship new features into the upcoming OCP release, we don't need to back out dependency changes to land a new installer release.

cgwalters:
In theory, it's more preferred for any work that is targeted for earlier RHEL (like 9.0) should actually be structured in terms of backported patches, not new upstream releases.  We are already creating branch structure and build configurations to support this.

Also, for OCP specifically we absolutely can take newer Rust (and Go) toolchains into the buildroot for things targeted for that stream.  In fact, we already have a build of coreos-installer queued there, so it's just a matter of getting the buildroot override in place, which is pretty easy.

bgilbert:
While we do backport fixes, I don't agree that entire features should be backported into older OCP packages.  Large divergences from upstream produce code paths that were never tested upstream and that remain a maintenance burden for the support lifetime of the OCP release.

Historically I'm not aware that we've pulled in newer Rust than is available in the underlying RHEL release.  I'm open to that, though.  Want to put in a ticket?

cathay4t:
Both fedora and c9s has shipped the update of nmstate 2.2.2 which should have this problem fixed. Can we re-run the CI?

bgilbert:
Rebased.  The RPM build succeeded, but the MSRV checks still fail, as expected.

cgwalters:
FYI filed https://issues.redhat.com/browse/CLOUDBLD-12755 since it turns out it's needed for newer rpm-ostree too.

bgilbert:
I've added a commit that temporarily enables `updates-testing` in the RPM CI job, unbreaking CI.  :tada:  Let's get this in?

cathay4t:
The nmstate-2.2.3-2.fc37 and nmstate-2.2.3-2.fc36 fixed the installing problem.

Issue #1039, opened 2022-11-02
RPM CI is broken because nmstate uses Rust 1.60 dependency syntax


bgilbert:
Well, unfortunately #1032 wasn't the end of it.  We hand-tested [FEDORA-2022-b499698493](https://bodhi.fedoraproject.org/updates/FEDORA-2022-b499698493) by manually installing all the coreos-installer build dependencies in a container, and indeed the nmstate package is installable and produces a successful build.

However, automatic depsolving doesn't work, because rust2rpm [doesn't yet support](https://pagure.io/fedora-rust/rust2rpm/issue/186) the new 1.60 dependency syntax.  As a result, rust-coreos-installer [currently FTBFS](https://kojipkgs.fedoraproject.org//work/tasks/6222/93696222/build.log) in Fedora:

```
+ /usr/bin/env CARGO_HOME=.cargo RUSTC_BOOTSTRAP=1 'RUSTFLAGS=-Copt-level=3 -Cdebuginfo=2 -Ccodegen-units=1 -Clink-arg=-Wl,-z,relro -Clink-arg=-Wl,-z,now -Clink-arg=-Wl,-dT,/builddir/build/BUILD/coreos-installer-0.16.1/.package_note-rust-coreos-installer-0.16.1-2.fc36.x86_64.ld --cap-lints=warn' /usr/bin/cargo build -j6 -Z avoid-dev-deps --release --features rdcore
error: no matching package named `nispor` found
location searched: registry `crates-io`
required by package `nmstate v2.2.0`
    ... which satisfies dependency `nmstate = "^2.1.2"` of package `coreos-installer v0.16.1 (/builddir/build/BUILD/coreos-installer-0.16.1)`
error: Bad exit status from /var/tmp/rpm-tmp.6RIcM4 (%build)
```

@cathay4t, maybe `nmstate.spec` could manually add the subpackage dependencies that rust2rpm currently can't produce on its own?  Another option might be to avoid the new dependency syntax upstream for now.

bgilbert:
For now, I'm going to mark the RPM test non-required so we can merge things in this repo.  The fix will need to happen in the nmstate packaging.

also cc @qinqon for visibility.

cathay4t:
@bgilbert You mean add something like these to `rust-nmstate+default-devel` and `rust-nmstate+gen_conf-devel`?

```
BuildRequires:  (crate(nispor/default) >= 1.2.8 with crate(nispor/default) < 2.0)
```

[...]

PR #1040, opened 2022-11-04
Use inline `format!` in a few places

cgwalters:
Just a drive by commit; I was looking at the code and I think inline format is nicer.


bgilbert (approved):

bgilbert:
RPM CI failure is expected due to https://github.com/coreos/coreos-installer/issues/1039.

PR #1047, opened 2022-11-16
blockdev.rs: add support for saving mbr partitions

### LINKED ISSUE #957: Parititon Saving should handel converting MBR partitions to GPT
# Feature Request #

## Desired Feature ##
[Now](https://github.com/coreos/coreos-installer/issues/816) that partition saving bails when its not able to save partitions that are found on a MBR device, we need to have a way to convert the listed save partitions that are MBR partitions to GPT partitions and save them. 

## Other Information ##
This request is expecting[ issue 816](https://github.com/coreos/coreos-installer/issues/816) to be resolved first.

prestist:
Fixes #957. If a MBR partition is marked to be saved, it will be translated using best effort to GPT. To elaberate on "best effort"; there is metadata which is present in GPT but not MBR, so it might not translate 1:1.


docs/release-notes.md
```diff
@@ -7,6 +7,7 @@ nav_order: 8
 ## Upcoming coreos-installer 0.17.0 (unreleased)
 
 Major changes:
+- install: if using `--save-partindex` add support for MBR device
```
lucab, line 10: Wording suggestion: Add support for MBR partition tables to `--save-partindex`

src/blockdev.rs
```diff
@@ -519,6 +520,32 @@ pub struct SavedPartitions {
     partitions: Vec<(u32, GPTPartitionEntry)>,
 }
 
+fn translate_mbr_types_to_gpt(sys: u8) -> Uuid {
+    // non inclusive best effort mapping of MBR types to GPT types
+    match sys {
```
lucab, line 525: If you source this table from somewhere on the web, it would be good to put an URL in comments here.

src/blockdev.rs
```diff
@@ -519,6 +520,32 @@ pub struct SavedPartitions {
...
+        // Linux filesystem
+        0x01 => uuid!("0FC63DAF-8483-4772-8E79-3D69D8477DE4"),
+        // Linux Filesystem Data
+        0x83 => uuid!("0FC63DAF-8483-4772-8E79-3D69D8477DE4"),
+        // Linux LVM
+        0x8e => uuid!("E6D6D379-F507-44C2-A23C-238F2A3DF928"),
+        // Linux Swap
+        0x82 => uuid!("0657FD6D-A4AB-43C4-84E5-0933C84B4F4F"),
+        _ => uuid!("00000000-0000-0000-0000-ffff00000000"),
```
lucab, line 534: I'm having some hard time cross-checking where this "generic partition type UUID" is coming from. Did you invent it or sourced it from somewhere?

src/blockdev.rs
prestist: Thats a great point, I will add the wiki links in a comment, thank you :)

src/blockdev.rs
prestist: LOL I thought I copied it from the wiki page, but I am not unsure It should be just an empty GUID as its not known.

docs/release-notes.md
prestist: waaay better, thank you

src/blockdev.rs
```diff
@@ -519,6 +520,36 @@ pub struct SavedPartitions {
...
+        _ => uuid!("00000000-0000-0000-0000-000000000000"),
+    }
+}
+
+fn translate_mbr_to_gpt(mbr_partition: &MBRPartitionEntry) -> GPTPartitionEntry {
+    let mut gpt_partition = GPTPartitionEntry::empty();
+
+    gpt_partition.partition_type_guid = *translate_mbr_types_to_gpt(mbr_partition.sys).as_bytes();
+    gpt_partition.unique_partition_guid = *Uuid::new_v4().as_bytes();
+    gpt_partition.starting_lba = u64::from(mbr_partition.starting_lba);
```
cgwalters, line 547: I think it's a bit more common to just use `as u64` here.  That said there's also the `try_from()` variant which might make sense to actually validate things.

src/blockdev.rs
```diff
@@ -14,6 +14,7 @@
 
 use anyhow::{anyhow, bail, Context, Result};
 use gptman::{GPTPartitionEntry, GPT};
+use mbrman::{MBRPartitionEntry, MBR};
```
cgwalters, line 17: Ah neat, I wasn't aware of these crates.  Looks like a good thing to use in https://github.com/containers/bootc/pull/30

src/blockdev.rs
```diff
@@ -566,39 +597,58 @@ impl SavedPartitions {
...
+                    let mbr = MBR::read_from(disk, u32::try_from(sector_size)?)
+                        .context("reading disk as MBR")?;
+
+                    // cross-check MBR sector size
+                    Self::verify_mbr_sector_size(&mbr, sector_size)?;
+
+                    for (i, p) in mbr.iter() {
+                        if Self::matches_filters_mbr(i.try_into().unwrap(), p, filters) {
+                            // if the partition is using any of the last 33 sectors, we need to bail.
+                            if p.starting_lba + p.sectors < mbr.disk_size - 33 {
```
cgwalters, line 632: May be worth hosting `33` into a `const FINAL_SECTORS: u32 = 33` above?

src/blockdev.rs
```diff
@@ -566,39 +597,58 @@ impl SavedPartitions {
...
+                                partitions.push((
+                                    i.try_into().context("convert usize into u32")?,
+                                    translate_mbr_to_gpt(p),
+                                ));
+                            } else {
+                                bail!("MBR partition {} is using the last 33 sectors of the disk, which is not supported by GPT", i);
```
cgwalters, line 638: Then this bit can reference it using inline format as `{FINAL_SECTORS}` instead of 33.

cgwalters (commented):
This is a superficial review; I'm not very familiar with this code or the problem it's solving.

src/blockdev.rs
bgilbert: We'd also need a comment explaining how we arrived at 33 sectors.  But note also that MBR is supported on 4Kn-sector disks (they just aren't bootable).  The backup GPT requires fewer sectors in that case.

GPTs are allowed to reserve more than the minimum number of sectors at either end of the disk.  We don't really check for that in the non-MBR case either, and don't have a way to, since we haven't seen the install image's partition table.  But I believe in that case we'd fail to restore partitions after installing and would then roll back to a fresh GPT with only the saved partitions, thus not losing data.

Long story short, I think we should probably skip this check.  The dry run below should catch overlarge partitions if there are any.  (But we should still have a test case for this.)

src/blockdev.rs
```diff
@@ -566,39 +597,58 @@ impl SavedPartitions {
...
                 {
-                    bail!("saving partitions from an MBR disk is not yet supported");
+                    let mbr = MBR::read_from(disk, u32::try_from(sector_size)?)
+                        .context("reading disk as MBR")?;
+
+                    // cross-check MBR sector size
+                    Self::verify_mbr_sector_size(&mbr, sector_size)?;
+
+                    for (i, p) in mbr.iter() {
+                        if Self::matches_filters_mbr(i.try_into().unwrap(), p, filters) {
```
bgilbert, line 630: The `unwrap` is inconsistent with the `try_into().context()?` below.  It seems very unlikely that even a crafted partition table would overflow a u32; let's just be consistent in how we handle the conversion.

---

This is starting to feel too deeply nested, and it seems a bit uncomfortable that `matches_filters_mbr` is almost the same as `matches_filters`.  Would it make sense to swap the order of translation and filtering, by translating the entire MBR to GPT (in a helper function) and then using the original flow?  We'd probably need to add an argument to `matches_filters` to disable matching by label (so `*` doesn't match translated MBR partitions) but that's still less code duplication.

Or actually, maybe it's okay for `*` to match MBR partitions.

src/blockdev.rs
```diff
@@ -566,39 +597,58 @@ impl SavedPartitions {
...
+                            // if the partition is using any of the last 33 sectors, we need to bail.
+                            if p.starting_lba + p.sectors < mbr.disk_size - 33 {
+                                partitions.push((
+                                    i.try_into().context("convert usize into u32")?,
```
bgilbert, line 634: If we're going to wrap with an error, let's make it clear what the problem was.

src/blockdev.rs
```diff
@@ -1569,17 +1643,35 @@ mod tests {
...
         // test trying to save partitions from a MBR disk
         let mut disk = make_unformatted_disk();
-        gptman::GPT::write_protective_mbr_into(&mut disk, 512).unwrap();
-        // label only
-        SavedPartitions::new(&mut disk, 512, &vec![label("*i*")]).unwrap();
+        let mut mbr = mbrman::MBR::new_from(&mut disk, 512 as u32, [0xff; 4])
+            .expect("could not create partition table");
+
+        // create a mbr partition to copy over to gpt
+        mbr[1] = mbrman::MBRPartitionEntry {
```
bgilbert, line 1650: Seems useful to wrap this into a function similar to `make_part()`.

src/blockdev.rs
```diff
@@ -1569,17 +1643,35 @@ mod tests {
 
         // test trying to save partitions from a MBR disk
         let mut disk = make_unformatted_disk();
-        gptman::GPT::write_protective_mbr_into(&mut disk, 512).unwrap();
-        // label only
-        SavedPartitions::new(&mut disk, 512, &vec![label("*i*")]).unwrap();
+        let mut mbr = mbrman::MBR::new_from(&mut disk, 512 as u32, [0xff; 4])
+            .expect("could not create partition table");
+
```
bgilbert, line 1648: One logical test case spans multiple code blocks, which is confusing.

src/blockdev.rs
```diff
@@ -1569,17 +1643,35 @@ mod tests {
...
         // index only
-        assert_eq!(
-            SavedPartitions::new(&mut disk, 512, &vec![Index(index(1), index(1))])
-                .unwrap_err()
-                .to_string(),
-            "saving partitions from an MBR disk is not yet supported"
-        );
-        // label and index
+        let saved = SavedPartitions::new(&mut disk, 512, &vec![Index(index(1), index(1))]).unwrap();
+        assert!(saved.is_saved());
```
bgilbert, line 1662: We're only checking that we think we succeeded, but we're not actually verifying the results against expected values.  Let's fix that.  In particular, we should check the resulting type GUIDs.

We should be testing a non-trivial MBR layout, where we aren't asking to save all of the partitions.  That layout should include logical partitions.  And we should also test saving a wildcard label `*` (which we might decide should match either everything or nothing).

We should also test with a 4Kn disk.

src/blockdev.rs
```diff
@@ -566,39 +597,58 @@ impl SavedPartitions {
...
                     && disk_has_mbr(disk).context("checking if disk has an MBR")?
                 {
-                    bail!("saving partitions from an MBR disk is not yet supported");
+                    let mbr = MBR::read_from(disk, u32::try_from(sector_size)?)
+                        .context("reading disk as MBR")?;
+
+                    // cross-check MBR sector size
+                    Self::verify_mbr_sector_size(&mbr, sector_size)?;
+
+                    for (i, p) in mbr.iter() {
```
bgilbert, line 629: `iter()` does return the extended partition, which would allow `--save-partindex` to match on it, which is surely not what we want.  We could skip the extended partition (`MBRPartitionEntry.is_extended()`) or error if someone tries to select it.  The former is easier, and is consistent with our semantics of letting users ask to save partitions that don't actually exist.

bgilbert (commented):
In the commit message, the prefix should be `blockdev:`, not `blockdev.rs:`.  Also I think "best effort" probably undersells the result.  We do translate the important parts (offset/length/partition number); the boot flag isn't relevant and the type code is basically ignored on Linux.  Let's explain which parts of the translation are fuzzy, but not introduce doubt about the validity of the result.

PR #1048, opened 2022-11-17
dockerfile: rebase to f37 from f36

prestist:
https://github.com/coreos/fedora-coreos-tracker/issues/1223

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:36 AS builder
+FROM registry.fedoraproject.org/fedora:37 AS builder
 RUN dnf install -y cargo git-core libzstd-devel openssl-devel xz-devel
 WORKDIR /build
 COPY Cargo.* ./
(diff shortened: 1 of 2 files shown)


dustymabe (approved):
LGTM

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:36 AS builder
+FROM registry.fedoraproject.org/fedora:37 AS builder
```
jlebon, line 1: This needs to update the second `FROM` below too.

jlebon (commented):
We've previously noted these bumps in a "Packaging Changes" section. (And actually circling back to [this comment](https://github.com/coreos/ignition/pull/1505#issuecomment-1319247130), we should probably do that there too?)

Dockerfile
prestist: Awesome catch thank you, and ok sorry I was going off of the butane train, didnt look on all apps; assumed it was a standard not to mention it.

jlebon (approved):
Super minor/optional: for the commit message, `Dockerfile` would be more correct since that's the name of the file, and "rebase from f36 to f37" would read better. :)

PR #1070, opened 2022-12-14
live: support new `igninfo.json` ISO file

jlebon:
This file includes information about where the Ignition embed area is located. This will allow Ignition embedding to work on s390x, where we can't just use the `ignition.img` trick. For backwards compatibility, we still look for `ignition.img` if `igninfo.json` isn't found.


jlebon:
Now with test fixtures added!

nikita-dubrovskii:
LGTM

tested on s390x

jacobemery:
Looks like it's failing on 'Build RPMs' for Fedora:

```
ERROR: Exception(rpms/rust-coreos-installer-0.16.1.102.g8e5eb9c-2.fc37.src.rpm) Config(fedora-37-x86_64) 2 minutes 23 seconds
INFO: Results and/or logs in: /var/lib/mock/fedora-37-x86_64/result
ERROR: Command failed: 
 # bash --login -c /usr/bin/rpmbuild -ba --noprep --noclean --target x86_64 --nodeps /builddir/build/SPECS/rust-coreos-installer.spec
```

bgilbert (commented):
LGTM generally.  See also comments in https://github.com/coreos/coreos-assembler/pull/3283#pullrequestreview-1325270925.

jlebon:
Rebased and review comments addressed!

tests/images/iso-extract-minimal-iso.sh
```diff
@@ -22,6 +22,11 @@ grepq() {
 iso=$1; shift
 iso=$(realpath "${iso}")
 
+if [[ "${iso}" == *.s390x.iso.xz ]]; then
```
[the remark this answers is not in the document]
jlebon: I went with a more liberal wildcard for this.

src/live/embed.rs
```diff
@@ -503,14 +504,41 @@ struct InitrdEmbedArea {
...
         let f = iso
-            .get_path(COREOS_INITRD_EMBED_PATH)
+            .get_path(&igninfo.file.to_uppercase())
             .context("finding initrd embed area")?
             .try_into_file()?;
+        let file_offset = igninfo.offset.unwrap_or(0);
+        let iso_offset = f.address.as_offset() + file_offset;
+        let length = igninfo
+            .length
+            .unwrap_or(f.length as usize - file_offset as usize);
```
jlebon, line 539: The double `as usize` casting isn't great here. One is from `u32` which seems reasonable but the other is from `u64` which seems more questionable. I brushed over this since in practice we know it'll fit with the offset values we're expecting. If we want to be more careful though, we could do this subtraction in u64 mode and then `TryFrom` the result into a `usize`.

jacobemery:
All tests passed! 👀

src/live/embed.rs
bgilbert: A valid offset can't be larger than the 32-bit maximum ISO9660 file size, and an invalid one can underflow the subtraction (in release builds) even with `u64`.  The consequence would be a short read error in `Region::read()`, which doesn't seem too bad.  I'd say we can leave it as is.

bgilbert (approved):

PR #1094, opened 2023-01-19
blockdev: use `-p` when calling `blkid`

jlebon:
By default, `blkid` will return cached data, which we don't want because
it might be stale. We need to use`-p` to make sure it directly probes
the block devices and bypasses the cache.

With `-p`, `blkid` requires passing the devices directly. Call it once
to gather the list of devices (we trust the cache enough for this) and
then again with `-p`.

We originally used `lsblk` here which uses the udev database, but this
was changed in https://github.com/coreos/coreos-installer/pull/813
because it failed to return the filesystem label in some instances. It
might be worth revisiting this at some point and find out if we were
just missing a `udevadm settle` somewhere.

See also: https://github.com/coreos/fedora-coreos-config/pull/2181#issuecomment-1397386896

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -15,6 +15,8 @@ Minor changes:
 
 Internal changes:
 
+- verify-unique-fs-label: Use `-p` when calling `blkid` to avoid stale cache
+
 Packaging changes:
 
 - Require `mbrman` ≥ 0.5.0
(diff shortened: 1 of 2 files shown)


bgilbert:
CI is unhappy:

```
[    3.062880] systemd[1]: Starting coreos-unique-boot.service - Ensure Unique `boot` Filesystem Label...
[    3.065984] rdcore[529]: blkid: The low-level probing mode requires a device
[    3.066512] rdcore[529]: Error: "blkid" "-p" failed with exit status: 4
[    3.072422] systemd[1]: coreos-unique-boot.service: Main process exited, code=exited, status=1/FAILURE
[    3.073203] systemd[1]: coreos-unique-boot.service: Failed with result 'exit-code'.
```

jlebon:
Updated and tested!

bgilbert (approved):

PR #1101, opened 2023-01-30
blockdev: Only compile regex once

cgwalters:
Lazily compile and cache the regexp we use to parse lsblk output just once - doing it on every line of the output is a bit wasteful. (Noticed this when copy-pasting the code into bootc)

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -14,6 +14,7 @@
 
 use anyhow::{anyhow, bail, Context, Result};
 use gptman::{GPTPartitionEntry, GPT};
+use lazy_static::lazy_static;
 use nix::sys::stat::{major, minor};
 use nix::{errno::Errno, mount, sched};
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


jlebon (commented):
Good catch. We use `lazy_static` elsewhere to compile regexes, so might as well use that here too for consistency and avoid a new dep.

prestist:
Hey @cgwalters would you mind adding a small concise comment in the release notes found here => https://github.com/coreos/coreos-installer/blob/main/docs/release-notes.md#upcoming-coreos-installer-0170-unreleased

This is an extra step we added to some of our repos to allow for an easy release cycle.

cgwalters:
bootc already uses once_cell consistently, but I downgraded back to lazy_static for now for here.

prestist (approved):
Wow nice optimization! LGTM

bgilbert (approved):
I concur with the `skip-notes` label; this isn't a user-facing change.

Issue #1109, opened 2023-02-08
Customizing kargs in iso unexpectedly creates ignition config.


willhaines:
# Bug #

Booting a live .iso with no ignition config present (default for RHCOS) drops you to a terminal where you can configure network interfaces, then run the coreos-installer. Some methods of installing OpenShift rely on this behavior.

After customizing the kernel args in an RHCOS .iso, we found that a (nearly empty) ignition file is suddenly present. This causes the live environment to drop you in a login prompt, rather than a usable terminal.

## Host Operating System Version ##

Fedora 36

## Target Operating System Version ##

RHCOS 4.12.2

## coreos-installer Version ##

v0.16.1

## Expected Behavior ##

```
$  coreos-installer iso ignition show rhcos-live.x86_64.iso
Error: No embedded Ignition config.
$  coreos-installer iso ignition show rhcos-live.x86_64.kargs.iso
Error: No embedded Ignition config.
```

## Actual Behavior ##

```
$  coreos-installer iso ignition show rhcos-live.x86_64.iso
Error: No embedded Ignition config.
$  coreos-installer iso ignition show rhcos-live.x86_64.kargs.iso
{"ignition":{"version":"3.3.0"}}
```

## Reproduction Steps ##

```
$ coreos-installer iso customize --live-karg-append foobar -o rhcos-live.x86_64.kargs.iso rhcos-live.x86_64.iso
```

## Other Information ##

This can be worked around by removing the ignition config, after making other customizations.

```
$ coreos-installer iso customize --live-karg-append foobar -o rhcos-live.x86_64.kargs.iso rhcos-live.x86_64.iso
$ coreos-installer iso ignition remove rhcos-live.x86_64.kargs.iso
```

bgilbert:
Good diagnosis, thanks for the report!  Fix in https://github.com/coreos/coreos-installer/pull/1110.

jlebon:
Note a workaround for this is to use `coreos-installer iso kargs modify` instead.

PR #1110, opened 2023-02-08
customize: don't unnecessarily embed a live Ignition config

bgilbert:
We were unconditionally embedding an Ignition config, even when only modifying the live kargs, which don't need one.  This broke autologin in the CoreOS live ISO, which only occurs when no Ignition config is present.

Fixes https://github.com/coreos/coreos-installer/issues/1109.


jlebon (approved):
Nice!

prestist (approved):

PR #1113, opened 2023-02-15
signing-keys: add Fedora 39 key

aaradhak:
We branched F38 from rawhide so rawhide is now F39.


bgilbert (approved):

jlebon:
Minor note: remember to keep a blank line separator between the commit subject and message. Otherwise git treats it all as the subject (check e.g. `git log --oneline`).

PR #1118, opened 2023-02-27
cargo: bump MSRV to 1.66.0

bgilbert:
We no longer need to support older Rust versions.

Use the new `dep:` syntax to avoid creating an implicit `clap_mangen` feature.


prestist (approved):
LGTM

PR #1120, opened 2023-03-01
ci: migrate to new kolaTestIso()

jlebon:
The new `kola testiso` code will test NM keyfiles by default, so there's no need to do a separate run anymore.

While we're here, just accept the new default list of tests from `kola testiso` by only denylisting the `iso-as-disk` and `iso-live-login` tests and otherwise not overriding the tests to run.

This will run more tests than before. We'll likely cut down tests we consider redundant soon, but that should be transparent to this code.


bgilbert (approved):
LGTM, thanks!

PR #1122, opened 2023-03-02
live: warn when `pxe customize` output requires Ignition kargs

bgilbert:
When installing with `coreos.inst` kargs, Ignition kargs are not normally used.  However, when a customized PXE initrd embeds a live Ignition config, Ignition kargs are required.  Warn the user in this case.

For https://github.com/coreos/fedora-coreos-tracker/issues/1426.

### DIFF
--- a/src/live/mod.rs
+++ b/src/live/mod.rs
@@ -432,6 +432,11 @@ pub fn pxe_customize(config: PxeCustomizeConfig) -> Result<()> {
 
     let live = LiveInitrd::from_common(&config.common, features)?;
     let initrd = live.into_initrd()?;
+    if initrd.get(INITRD_IGNITION_PATH).is_some() {
+        eprintln!(
+            "PXE configuration must include kernel arguments:\n\tignition.firstboot ignition.platform.id=metal"
+        );
+    }
 
     // append customizations to output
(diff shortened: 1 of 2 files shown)


src/live/mod.rs
```diff
@@ -432,6 +432,11 @@ pub fn pxe_customize(config: PxeCustomizeConfig) -> Result<()> {
 
     let live = LiveInitrd::from_common(&config.common, features)?;
     let initrd = live.into_initrd()?;
+    if initrd.get(INITRD_IGNITION_PATH).is_some() {
+        eprintln!(
+            "PXE configuration must include kernel arguments:\n\tignition.firstboot ignition.platform.id=metal"
+        );
+    }
```
dustymabe, line 439: So this is just a warning the user will always see? I guess it has to be because it can't know what kargs are going to be set in the PXE config, right?

dustymabe (approved):
LGTM

src/live/mod.rs
bgilbert: [We already have a couple of those](https://github.com/coreos/coreos-installer/blob/db3b5127a7b15209757363194055cb84cf258385/src/live/customize.rs#L378-L384).  And yes; on PXE the user needs to configure their own kargs.

PR #1125, opened 2023-03-05
tests/images: only test installed binary if env var set

bgilbert:
In general, we want to test the binary built in the source tree and raise an error if that binary is missing.  Historically we failed to notice this case and fell back to testing the installed binary, if present. CoreOS CI took advantage of this.

Fail if the coreos-installer binary isn't in the source tree, unless `COREOS_INSTALLER_TEST_INSTALLED_BINARY` is set, in which case print a warning and continue.

Also have CI delete the OS copy of coreos-installer before overwriting it with rsync.  This ensures we don't test the wrong thing if the OS copy is elsewhere in `PATH`, or if coreos-installer is missing from the install tree.

Followup to https://github.com/coreos/coreos-installer/pull/999#discussion_r1004688601.

### DIFF
--- a/tests/images.sh
+++ b/tests/images.sh
@@ -3,8 +3,20 @@
 set -euo pipefail
 
 export PATH
-if [[ -d "$(realpath "$(dirname "$0")/../target")" ]]; then
-    PATH="$(realpath "$(dirname "$0")/../target/${PROFILE:-debug}"):$PATH"
+bindir="$(realpath -m "$(dirname "$0")/../target/${PROFILE:-debug}")"
+if [[ -x "$bindir/coreos-installer" ]]; then
+    PATH="$bindir:$PATH"
+elif command -v coreos-installer >/dev/null; then
+    echo "$bindir/coreos-installer not found"
+    if [[ -n "${COREOS_INSTALLER_TEST_INSTALLED_BINARY:-}" ]]; then
+        echo "COREOS_INSTALLER_TEST_INSTALLED_BINARY set; testing $(command -v coreos-installer)"
+    else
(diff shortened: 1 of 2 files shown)


travier (approved):
This looks like valid Bash but I'm not fully into the details for the CI here so this is a weak LGTM.

tests/images.sh
```diff
@@ -3,8 +3,18 @@
 set -euo pipefail
 
 export PATH
-if [[ -d "$(realpath "$(dirname "$0")/../target")" ]]; then
-    PATH="$(realpath "$(dirname "$0")/../target/${PROFILE:-debug}"):$PATH"
+bindir="$(realpath -m "$(dirname "$0")/../target/${PROFILE:-debug}")"
+if [[ -x "$bindir/coreos-installer" ]]; then
+    PATH="$bindir:$PATH"
+else
```
jlebon, line 9: Should this be e.g.

```
elif command -v coreos-installer >/dev/null; then
```

?

jlebon (approved):
Looks sane overall. One thing that'd be nice to strengthen is making sure we're testing the coreos-installer we think we are. The CoreOS CI code does an `rsync`, and assumes it's overwriting the built-in coreos-installer in the process. Maybe we should do an explicit `rm /usr/bin/coreos-installer` before the `rsync`? That way it makes it obvious that the original coreos-installer is gone.

tests/images.sh
bgilbert: I was relying on `which` to error out if the command doesn't exist, but your approach is clearer.

bgilbert:
Updated!

jlebon (approved):

PR #1130, opened 2023-03-09
cargo: require ignition-config 0.3

bgilbert:
ignition-config 0.2 doesn't support Ignition config spec 3.4.0. coreos-installer itself doesn't care, but it'd be confusing if some coreos-installer 0.17.0 binaries supported 3.4.0 configs and some didn't. Explicitly require ignition-config 0.3 to warn downstreams.


prestist (approved):
LGTM; Good catch!

PR #1131, opened 2023-03-09
Revert "osmet: avoid install performance regression in debug mode on Rust 1.64+"

bgilbert:
The Rust regression was fixed by https://github.com/rust-lang/rust/pull/102760, which is in Rust 1.66+. Our MSRV is 1.66.0, so we can drop this.

This reverts commit 1ccb9b091afddca29fe64135882a48b7295e6ab8.


prestist (approved):
LGTM!

Issue #1158, opened 2023-04-04
Get error about SSL when install FCOS in beaker using pxe


HuijingHei:
See console logs:
```
[  103.080254] systemd-resolved[2869]: System hostname changed to 'ampere-mtsnow-altramax-13.khw4.lab.eng.bos.redhat.com'. 
[  103.140581] systemd[1]: Starting NetworkManager-dispatcher.service - Network Manager Script Dispatcher Service... 
[  103.170913] coreos-installer-service[2952]: coreos-installer install /dev/nvme0n1 --ignition-url http://file.nay.redhat.com/~hhei/beaker-install.ign --insecure-ignition --firstboot-args rd.neednet=1 ip=dhcp  --image-url https://builds.coreos.fedoraproject.org/prod/streams/next/builds/38.20230322.1. 
[  103.220301] coreos-installer-service[2952]: 0/aarch64/fedora-coreos-38.20230322.1.0-metal.aarch64.raw.xz --insecure --fetch-retries infinite 
[  103.260377] NetworkManager[2881]: <info>  [1677024077.4527] device (eno1): state change: ip-check -> secondaries (reason 'none', sys-iface-state: 'managed') 
[  103.300233] NetworkManager[2881]: <info>  [1677024077.4530] device (eno1): state change: secondaries -> activated (reason 'none', sys-iface-state: 'managed') 
[  103.340206] NetworkManager[2881]: <info>  [1677024077.4537] manager: NetworkManager state is now CONNECTED_SITE 
[  103.370441] NetworkManager[2881]: <info>  [1677024077.4542] device (eno1): Activation: successful, device activated. 
[  103.400186] NetworkManager[2881]: <info>  [1677024077.4550] manager: NetworkManager state is now CONNECTED_GLOBAL 
[  103.430176] NetworkManager[2881]: <info>  [1677024077.4557] manager: startup complete 
[  103.460168] systemd[1]: Finished NetworkManager-wait-online.service - Network Manager Wait Online. 
[  103.490154] systemd[1]: Started NetworkManager-dispatcher.service - Network Manager Script Dispatcher Service. 
[  103.520243] systemd[1]: Reached target network-online.target - Network is Online. 
[  103.550191] systemd[1]: Starting coreos-installer.service - CoreOS Installer... 
[  103.580178] systemd[1]: iscsi.service: Unit cannot be reloaded because it is inactive. 
[...]

PR #1159, opened 2023-04-10
service: wait for chronyd to sync the system time for `coreos-installer`

HuijingHei:
See https://github.com/coreos/coreos-installer/issues/1158


HuijingHei:
I have not yet tested it.

jlebon:
Two thoughts on this:
1. Note that `After=chronyd.service` doesn't guarantee that the clock has been updated by the time our unit runs, only that chronyd was started.
2. I'm still confused why we need this since coreos-installer should be retrying. It seems like we should dig into why chronyd doesn't eventually kick in while we retry?

PR #1193, opened 2023-05-23
build-sys: Use new `tier = 2` for cargo-vendor-filterer

cgwalters:
This fixes the problem described in the comment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,12 +20,8 @@ tag-message = "coreos-installer v{{version}}"
 
 # See https://github.com/coreos/cargo-vendor-filterer
 [package.metadata.vendor-filter]
-# cargo-vendor-filterer supports wildcards but requires some degree of LLVM
-# support for every matching platform, which isn't necessarily available
-# out of the box.  Use these as stand-ins.  Note that architecture-specific
-# crate dependencies (apart from wasm32) are uncommon in the ecosystem, but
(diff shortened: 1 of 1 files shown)


cgwalters:
see also e.g.
- https://github.com/coreos/rpm-ostree/pull/4429
- https://github.com/containers/bootc/pull/104

bgilbert:
Since this widens the set of platforms we're including, it could conceivably change the contents of the vendor tarball.  Does it?  If so, we should probably release-note the change to vendored platforms under "packaging changes".

Also, it looks like there are a few other packages that will need updates.  Could you do those as well?
- [ ] afterburn
- [ ] bootupd
- [ ] ssh-key-dir
- [ ] zincati

cgwalters:
> Since this widens the set of platforms we're including, it could conceivably change the contents of the vendor tarball. Does it?

Just verified that there are no changes to the vendored crate set.  It'd be surprising if there were - last I dug at this I didn't find any architecture-dependent crates, just platform dependent.

bgilbert:
Okay, cool.  This LGTM; leaving open so we don't forget to update the other projects in https://github.com/coreos/coreos-installer/pull/1193#issuecomment-1560223004.

bgilbert:
Thanks for doing the other PRs.  Once you get the Zincati one posted, I'll approve this.

PR #1194, opened 2023-05-25
systemd/coreos-installer-generator: Properly read cmdline

travier:
Fix ShellCheck warning:
```
In ./systemd/coreos-installer-generator line 13:
cmdline=( $(</proc/cmdline) )
          ^---------------^ SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).
```

See: https://github.com/coreos/coreos-installer/pull/999
See: https://bugzilla.redhat.com/show_bug.cgi?id=1989551

### DIFF
--- a/systemd/coreos-installer-generator
+++ b/systemd/coreos-installer-generator
@@ -10,7 +10,7 @@ exec 1>/dev/kmsg; exec 2>&1
 
 UNIT_DIR="${1:-/tmp}"
 
-cmdline=( $(</proc/cmdline) )
+IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
 karg() {
     local name="$1" value="$2"
     for arg in "${cmdline[@]}"; do


systemd/coreos-installer-generator
```diff
@@ -10,7 +10,7 @@ exec 1>/dev/kmsg; exec 2>&1
 
 UNIT_DIR="${1:-/tmp}"
 
-cmdline=( $(</proc/cmdline) )
+IFS=" " read -r -a cmdline <<< "$(</proc/cmdline)"
```
bgilbert, line 13: Couldn't this just be:

```suggestion
IFS=" " read -r -a cmdline < /proc/cmdline
```

travier: I don't know. I copy / pasted the fix from https://github.com/coreos/coreos-installer/pull/999/files#diff-a31b15f1dd66afbb6fe505b362441e553a822b80292f285a5ef93cb89cb0b233R21

bgilbert: While I think we could probably clean this up, I agree that there's value to staying consistent.

PR #1203, opened 2023-06-02
rdcore: Juggle physical root versus deployment root

cgwalters:
rdcore: Juggle physical root versus deployment root

This is prep for supporting composefs, where the physical root
is distinct from the deployment root.  Specifically for the LUKS
case, we can find `/etc/crypttab` only in the deployment root.

Otherwise, we suffix the passed path (usually `/sysroot`) that
was mounted in the initramfs with `/sysroot` to find the physical
root.

xref https://github.com/ostreedev/ostree/issues/2867

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -6,6 +6,19 @@ nav_order: 8
 
 ## Upcoming coreos-installer 0.19.0 (unreleased)
 
+Major changes:
+
+
+Minor changes:
+
+
+Internal changes:
+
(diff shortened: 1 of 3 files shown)


cgwalters:
Ping

cgwalters:
I'd really like to turn on composefs by default soon for ostree, and this is going to block that.

jlebon (approved):

cgwalters:
Looks like there hasn't been a release since this got merged?  This is blocking composefs enablement in rawhide which I'd like to do soon.

PR #1230, opened 2023-07-07
s390x: enable 'iso kargs' commands

nikita-dubrovskii:
```
$ coreos-installer iso kargs show  ~/fcos/builds/latest/s390x/*.iso
mitigations=auto,nosmt coreos.liveiso=fedora-coreos-38.20230706.dev.0 ignition.firstboot ignition.platform.id=metal
$ md5sum builds/latest/s390x/*.iso
91d89e3f69a23f750f7498ee2b30d02e  builds/latest/s390x/fedora-coreos-38.20230706.dev.0-live.s390x.iso

$ coreos-installer iso kargs modify -a foo ~/fcos/builds/latest/s390x/*.iso
$ coreos-installer iso kargs show  ~/fcos/builds/latest/s390x/*.iso
mitigations=auto,nosmt coreos.liveiso=fedora-coreos-38.20230706.dev.0 ignition.firstboot ignition.platform.id=metal foo
$ md5sum builds/latest/s390x/*.iso
5c8b1ba5dd7916300101d47c3a72da50  builds/latest/s390x/fedora-coreos-38.20230706.dev.0-live.s390x.iso

$ coreos-installer iso kargs modify -d foo  ~/fcos/builds/latest/s390x/*.iso
$ coreos-installer iso kargs show  ~/fcos/builds/latest/s390x/*.iso
mitigations=auto,nosmt coreos.liveiso=fedora-coreos-38.20230706.dev.0 ignition.firstboot ignition.platform.id=metal
$ md5sum builds/latest/s390x/*.iso
91d89e3f69a23f750f7498ee2b30d02e  builds/latest/s390x/fedora-coreos-38.20230706.dev.0-live.s390x.iso
```


src/live/embed.rs
```diff
@@ -448,13 +455,15 @@ impl KargEmbedAreas {
...
             args,
+            arch,
         })
     }
 
     fn parse(region: &Region) -> Result<String> {
         Ok(String::from_utf8(region.contents.clone())
             .context("invalid UTF-8 in karg area")?
             .trim_end_matches('#')
+            .trim_end_matches('\0')
```
dustymabe, line 466: dumb question.. why do we need to use `\0` on s390x? I feel like we could add some comments on this somewhere (maybe not here, but when `placeholder` is set below.

src/live/embed.rs
```diff
@@ -267,13 +267,15 @@ struct KargEmbedAreas {
...
     args: String,
+    arch: Option<String>,
 }
 
 #[derive(Deserialize, Serialize)]
 struct KargEmbedInfo {
     default: String,
     files: Vec<KargEmbedLocation>,
     size: usize,
+    arch: Option<String>,
```
dustymabe, line 278: I'm wondering if changing these structs will introduce a backwards compat issue.. For example if I use coreos-installer that's released today to embed kargs in an ISO and then later try to read those using new coreos-installer (after this merges and new release is out) will it work because no arch was specified?

dustymabe (approved):
A few questions.. though I'll say I'm not familiar with this code and not really good at rust.

src/live/embed.rs
nikita-dubrovskii: I don't know and will ask around. using `#` breaks the boot

src/live/embed.rs
nikita-dubrovskii: I've checked modified installer & old ISO, old installer & new ISO - it works on x86. But s390x requires new installer and new image.

src/live/embed.rs
dustymabe: Thanks!

src/live/embed.rs
bgilbert: `iso extract minimal-iso` serializes `KargEmbedInfo`, and I believe would write `"arch": null` to an old ISO.  We should probably mark this `#[serde(skip_serializing_if = "Option::is_none")]`.

But also, I think `arch` is an overly indirect way of encoding the desired suffix; see https://github.com/coreos/coreos-assembler/pull/3527#discussion_r1263118197.

bgilbert:
Please remember to add a new test fixture in `fixtures/iso`.

src/live/embed.rs
nikita-dubrovskii: Right, it's `null`. Marked with `skip_serializing_if`.

nikita-dubrovskii:
> Please remember to add a new test fixture in `fixtures/iso`.

Thx, wasn't aware about this. Added and enabled `s390x minimal-iso & kargs` tests as well.

src/live/embed.rs
nikita-dubrovskii: Found only some s390x assembler and C code, where `0` is used as filler.

src/live/embed.rs
```diff
@@ -420,19 +433,25 @@ impl KargEmbedAreas {
...
 
-        Some(Self::build(length, default, regions)).transpose()
+        Some(Self::build(length, default, None, None, regions)).transpose()
     }
 
-    fn build(length: usize, default: String, regions: Vec<Region>) -> Result<Self> {
+    fn build(
+        length: usize,
+        default: String,
+        arch: Option<String>,
```
dustymabe, line 442: We don't need this anymore do we?

```suggestion
```

src/live/embed.rs
```diff
@@ -477,8 +498,14 @@ impl KargEmbedAreas {
                 self.length
             );
         }
-        let mut contents = vec![b'#'; self.length];
-        contents[..formatted.len()].copy_from_slice(formatted.as_bytes());
+
+        let filler = self.filler.unwrap_or('#');
+        let size = match self.arch.as_ref() {
+            Some(arch) if arch == "s390x" => formatted.len() - 1, // s390x ISO: kargs line stored within cdboot.img file ends with zeros and doesn't contain '\n'
```
dustymabe, line 504: hmm. for some reason I thought the reason we encoded the `filler` was so we could drop encoding the architecture: https://github.com/coreos/coreos-assembler/pull/3527#discussion_r1263118197

but we still need to?

src/live/embed.rs
nikita-dubrovskii: we do

src/live/embed.rs
nikita-dubrovskii: as i got from @bgilbert comments it'd be good to have the `filler`:
> non-s390x always wants # and s390x always wants \0, which isn't guaranteed to always be true

but we still need to skip `\n` for s390x, so `arch` field is also required

dustymabe: could we just encode this as `endbyte` or something? it feels like basically the same thing as the argument for why to encode `filler` versus relying on the architecture like you were doing before.

bgilbert: I agree with Dusty that we should avoid having an `arch` field.  See https://github.com/coreos/coreos-assembler/pull/3527#discussion_r1267632070.

fixtures/iso/INDEX
```diff
@@ -41,9 +41,17 @@ embed-areas-2023-03.x86_64.iso.xz
         live-initrd-network
     installer-config-directives from 0.16.0
 
-embed-areas-2023-03.s390x.iso.xz
```
bgilbert, line 44: Why is this one being removed?  We should always retain old fixtures; that's the purpose of having them.

fixtures/iso/INDEX
```diff
@@ -41,9 +41,17 @@ embed-areas-2023-03.x86_64.iso.xz
...
+embed-areas-2023-07.x86_64.iso.xz
+    igninfo.json pointing to ignition.img (no offset and no length)
+    kargs.json pointing to kargs areas and embedding defaults
```
bgilbert, line 46: `kargs.json` is different from the earlier fixtures.  We should specify how.

tests/images/iso-kargs.sh
```diff
@@ -120,9 +124,23 @@ offset=$(coreos-installer dev show iso --kargs "${iso}" | jq -r .kargs[0].offset
...
-filler="$(printf '%*s' $((${length} - ${expected_args_len} - 1)) | tr ' ' '#')"
-if ! echo -en "${expected_args}\n${filler}" | cmp -s <(dd if=${iso} skip=${offset} count=${length} bs=1 status=none) -; then
-    fatal "Failed to manually round-trip kargs"
+if [[ -n "${s390x}" ]]; then
+    # There is no filler or newline character on s390x's kargs line, but null bytes:
+    #    $ hexdump -C -s 138457 -n 896 builds/latest/s390x/*.iso
+    #    00021cd9  69 67 6e 69 74 69 6f 6e  2e 70 6c 61 74 66 6f 72  |ignition.platfor|
+    #    00021ce9  6d 2e 69 64 3d 6d 65 74  61 6c 00 00 00 00 00 00  |m.id=metal......|
+    # Create temp file with kargs and nulls and compare it to bytes of the iso image:
+    kargs_tmp=$(mktemp)
```
bgilbert, line 133: We're already in a temporary directory that will be automatically cleaned up.  Let's just use a hardcoded name here.

tests/images/iso-kargs.sh
```diff
@@ -120,9 +124,23 @@ offset=$(coreos-installer dev show iso --kargs "${iso}" | jq -r .kargs[0].offset
...
+    echo -n "${expected_args}" > "${kargs_tmp}"
+    truncate -s "${length}" "${kargs_tmp}"
+    if ! cat "${kargs_tmp}" | cmp -s <(dd if=${iso} skip=${offset} count=${length} bs=1 status=none) -; then
```
bgilbert, line 136: Nit:
```suggestion
    if ! cmp -s <(dd if=${iso} skip=${offset} count=${length} bs=1 status=none) "${kargs_tmp}"; then
```

src/live/embed.rs
```diff
@@ -162,10 +162,20 @@ struct Region {
     pub contents: Vec<u8>,
     #[serde(skip_serializing)]
     pub modified: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    end: Option<char>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pad: Option<char>,
```
bgilbert, line 168: Nit: let's reverse the order of these to match the function argument order.

src/live/embed.rs
```diff
@@ -280,6 +292,10 @@ struct KargEmbedInfo {
 struct KargEmbedLocation {
     path: String,
     offset: u64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    end: Option<char>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pad: Option<char>,
```
bgilbert, line 298: Likewise.

src/live/embed.rs
```diff
@@ -454,7 +478,7 @@ impl KargEmbedAreas {
     fn parse(region: &Region) -> Result<String> {
         Ok(String::from_utf8(region.contents.clone())
             .context("invalid UTF-8 in karg area")?
-            .trim_end_matches('#')
+            .trim_end_matches(region.pad.unwrap_or('#'))
             .trim()
```
bgilbert, line 482: If we're going to support arbitrary end markers, this `trim()` needs to be updated to handle those.

tests/images/iso-kargs.sh
```diff
@@ -130,10 +130,10 @@ if [[ -n "${s390x}" ]]; then
     #    00021cd9  69 67 6e 69 74 69 6f 6e  2e 70 6c 61 74 66 6f 72  |ignition.platfor|
     #    00021ce9  6d 2e 69 64 3d 6d 65 74  61 6c 00 00 00 00 00 00  |m.id=metal......|
     # Create temp file with kargs and nulls and compare it to bytes of the iso image:
-    kargs_tmp=$(mktemp)
+    kargs_tmp=cmdline.s390x
```
bgilbert, line 133: Optional: the variable could be dropped now.

bgilbert (approved):
I haven't reviewed this carefully, but LGTM broadly.

The second commit should be squashed.

bgilbert (approved):

dustymabe (approved):

Issue #1231, opened 2023-07-08
Validate ignition before installation


vilgotf:
# Feature Request #

Coreos should do some basic validation of provided ignition files before writing them to disk in the installation process.

## Desired Feature ##

When doing an install just now, I accedentally provided a non raw URL to the `-I` flag, meaning it downloaded HTML instead of by ignition file. It would be great if coreos-installer could catch this rather than the user needing to track it down. Feeding coreos-installer an invalid ignition file currently doesn't even bring up link-local networking on the host which makes troubleshooting awkward for headless deployments.

The validation does not need to be thorough in any way shape or form, just trying to parse the data as JSON should be enough.

bgilbert:
Thanks for the report!  Fix in https://github.com/coreos/coreos-installer/pull/1232.

PR #1232, opened 2023-07-08
install: verify that Ignition config is valid JSON

bgilbert:
If the user accidentally specifies a URL to an HTML page instead of an Ignition config, detect that before we install.

Fixes https://github.com/coreos/coreos-installer/issues/1231.


dustymabe (approved):
I'm not a rust expert.

Trivial LGTM

PR #1274, opened 2023-08-22
WIP: s390x: run chreipl only on zVM

nikita-dubrovskii:
There is no need to re-IPL disk under QEMU/KVM

### DIFF
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -14,9 +14,9 @@
 
 use crate::blockdev::Mount;
 use crate::io::{visit_bls_entry, visit_bls_entry_options, Initrd, KargsEditor};
-use crate::runcmd;
 use crate::s390x::ZiplSecexMode;
 use crate::util::cmd_output;
+use crate::{runcmd, runcmd_output};
 use anyhow::{anyhow, Context, Result};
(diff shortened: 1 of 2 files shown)


src/s390x/zipl.rs
```diff
@@ -28,8 +28,11 @@ use tempfile::{Builder, NamedTempFile};
 
 /// Sets the boot device to `dev` using `chreipl`.
```
dustymabe, line 29: Could we add a comment here to explain why we only want to do this on zvm? As someone who doesn't know anything about s390x it would probably help me understand. What are the other platfomrs and what will `systemd-detect-virt` return on s390x that isn't zVM?

nikita-dubrovskii:
I'm inclined to close this PR. But some details worth to mention here:
When we install Linux on zVM we always use so-called `reader` devices (with kernel,ramfs,cmdline) to boot, once system installation is done and `zipl` updated the bootloader, we can mark `dasd` / `scsi` as next boot target.
Qemu doesn't require but still supports `chreipl`, as it has `bootindex` option.
And here comes a problem -  `kola` uses `iso` image to install VM and doesn't change boot order (`bootindex=1`) after installation,  so CDROM is always the 1st boot device. So let's just keep this functionality

PR #1304, opened 2023-09-19
Revert "ci: run cosa unprivileged temporarily for virtiofs bug"

jlebon:
This reverts commit 2247ffe1f8f64317da570a8d230bfc1ff782164b.

This shouldn't be necessary now with the workaround built into cosa:

https://github.com/coreos/coreos-assembler/pull/3625


prestist (approved):
LGTM

PR #1338, opened 2023-11-16
s390x: add zVM Secure IPL support

nikita-dubrovskii:
Implements proposal from: https://github.com/coreos/enhancements/pull/10

### DIFF
--- a/docs/cmd/install.md
+++ b/docs/cmd/install.md
@@ -163,4 +163,7 @@ Advanced Options:
           indefinitely.
 
           [default: 0]
+
+      --secure-ipl
+          Enable IBM Secure IPL
 ```
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -8,6 +8,7 @@ nav_order: 8
 
 Major changes:
 
+- install: Add z/VM Secure IPL support
 
 Minor changes:
 
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -14,10 +14,11 @@
 
 use crate::blockdev::Mount;
 use crate::io::{visit_bls_entry, visit_bls_entry_options, Initrd, KargsEditor};
-use crate::runcmd;
 use crate::s390x::ZiplSecexMode;
 use crate::util::cmd_output;
-use anyhow::{anyhow, Context, Result};
+use crate::{runcmd, runcmd_output};
+use anyhow::{anyhow, bail, Context, Result};
+use lazy_static::lazy_static;
 use nix::mount::MsFlags;
 use regex::Regex;
 use std::fs::{copy, create_dir_all, read_dir, DirEntry, File};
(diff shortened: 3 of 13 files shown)


src/s390x/zipl.rs
```diff
@@ -33,6 +34,69 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+                    .arg(&cap[1])
+                    .arg("portname")
+                    .arg(&cap[2][0..8])
+                    .arg(&cap[2][8..])
+                    .arg("lun")
+                    .arg(&cap[3][0..8])
+                    .arg(&cap[3][8..]);
+            }
+        }
+        _ => bail!("unsupported device: {} id: {}", devtype, id),
```
jlebon, line 93: Ouf, that's unfortunate. Any interface in e.g. `/sys` we could use instead?

systemd/coreos-installer-reboot-loaddev.service
```diff
@@ -0,0 +1,16 @@
...
+
+Requires=coreos-installer.target
+After=coreos-installer.target
+After=coreos-installer-reboot.service
+OnFailure=emergency.target
+OnFailureJobMode=replace-irreversibly
+
+[Service]
+Type=simple
+ExecStart=/usr/sbin/vmcp ipl loaddev
```
jlebon, line 14: Does this tell the hypervisor to do a hard reset or does the guest still have an orderly shutdown?

nikita-dubrovskii: Hardreset. I guess it's not a problem on zVM

src/s390x/zipl.rs
nikita-dubrovskii: Do you mean using `/sys` to get dev/lun/port? Yep, it's possible, but requires a lot of code, which would be similar to what `lszdev` has/does. I was thinking about it, but we already use `lsblk` , so decided to use existing tool to get exact ids of `install_target`

docs/cmd/install.md
```diff
@@ -163,4 +163,7 @@ Advanced Options:
           indefinitely.
 
           [default: 0]
+
+      --secure-ipl
+          Enable zVM Secure IPL
```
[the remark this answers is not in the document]
nikita-dubrovskii: There are plans to install RHCOS on LPAR directly, i don't mind to change the name, but than in the future we'd have to add smth like `--lpar-secure-ipl` . I'd prefer just `--secure-ipl` for both zVM & LPAR.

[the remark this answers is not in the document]
jlebon: Ahh OK. Is the interface and implementation planned to be the same?

src/s390x/zipl.rs
jlebon: OK.

Can you (1) factor out the parsing into separate functions, e.g. `parse_lszdev_dasd` and `parse_lszdev_zfcp`, and (2) in those functions, provide sample output in a comment of what we're trying to parse?

systemd/coreos-installer-reboot-loaddev.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Reboot from zVM LOADDEV after CoreOS Installer
+ConditionVirtualization=zvm
+ConditionPathExists=/run/coreos-installer-loaddev
+
+Requires=coreos-installer.target
+After=coreos-installer.target
+After=coreos-installer-reboot.service
```
jlebon, line 8: Hmm, wouldn't we have already issued a `systemctl reboot` if we run after `coreos-installer-reboot.service`?

I think it'd be clearer if we didn't have both services run.

This service also currently ignores `skip_reboot`.

Let's instead have both of them fire off of `/run/coreos-installer-reboot` and add `ConditionKernelCommandLine=!zvm` in `coreos-installer-reboot.service`?

Then we can get rid of `/run/coreos-installer-loaddev`.

systemd/coreos-installer-reboot-loaddev.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Reboot from zVM LOADDEV after CoreOS Installer
+ConditionVirtualization=zvm
```
jlebon, line 3: I think it'd be cleaner without this one and relied only on `coreos.inst.secure_ipl`. That way, we don't even have to think through the case where `coreos.inst.secure_ipl` is given but `ConditionVirtualization=zvm` is somehow false.

src/s390x/zipl.rs
```diff
@@ -33,6 +34,101 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+    }
+}
+
+/// Sets zVM Secure Boot (Secure IPL) boot device to `dev`.
+pub fn set_loaddev<P: AsRef<Path>>(dev: P) -> Result<()> {
+    // check if system is zVM guest
+    if !Path::new("/dev/vmcp").exists() {
+        return Ok(());
+    }
```
jlebon, line 105: Should we silently accept this? ISTM like we should fail pretty loudly if secure IPL was requested in an invalid context.

systemd/coreos-installer-reboot-loaddev.service
nikita-dubrovskii: Hmm, actually now i see an issue here:
- zVM require `vmcp ipl loaddev` to restart with Secure IPL
- LPAR has no `vmcp` and regular `reboot` is enough

Looks like it's better to make a shell script and call appropriate `reboot` depending in virtualization. Wdyt?

docs/release-notes.md
```diff
@@ -8,6 +8,7 @@ nav_order: 8
 
 Major changes:
 
+- install: Add zVM Secure IPL support
```
jlebon, line 11:
```suggestion
- install: Add z/VM Secure IPL support
```

src/s390x/zipl.rs
```diff
@@ -33,11 +34,116 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+            cap[3].to_string(),
+        ));
+    }
+    bail!("bad zFCP id: {}", line);
+}
+
+fn parse_lszdev<P: AsRef<Path>>(dev: P) -> Result<Loaddev> {
+    // We don't want to traverse sysfs and do same stuff lszdev does,
+    // so just call it to get required info. Here is sample output:
+    // $ lszdev -c TYPE,ID -n
```
jlebon, line 76: I assume you checked with `--list-columns` that there aren't more targeted columns we could use to avoid regexing things?

src/s390x/zipl.rs
```diff
@@ -33,11 +34,116 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+    match devtype {
+        "dasd-eckd" => parse_lszdev_eckd(id),
+        "zfcp-lun" => parse_lszdev_zfcp(id),
+        _ => bail!("unsupported device: {} id: {}", devtype, id),
+    }
+}
+
+/// Sets zVM Secure Boot (Secure IPL) boot device to `dev`.
+pub fn set_loaddev<P: AsRef<Path>>(dev: P) -> Result<()> {
+    if !secure_boot_is_enabled()? {
```
jlebon, line 102: Can we rename this function `secure_ipl_is_supported()`?

systemd/coreos-installer-reboot-loaddev.service
jlebon, line 1: Since this handles both the z/VM and LPAR case, it'd be more accurate to call it e.g. `coreos-installer-secure-ipl-reboot.service` since `loaddev` is only done in one of those cases.

systemd/coreos-installer-reboot-loaddev.service
```diff
@@ -0,0 +1,14 @@
...
+ConditionPathExists=/run/coreos-installer-reboot
+ConditionKernelCommandLine=coreos.inst.secure_ipl
+Requires=coreos-installer.target
+After=coreos-installer.target
+OnFailure=emergency.target
+OnFailureJobMode=replace-irreversibly
+
+[Service]
+Type=simple
+ExecStart=/bin/bash -euo pipefail -c '[[ -e /dev/vmcp ]] && vmcp ipl loaddev || systemctl --no-block reboot'
```
jlebon, line 12: The strict mode switches are not useful here.

```suggestion
ExecStart=/bin/sh -c '[ -e /dev/vmcp ] && vmcp ipl loaddev || systemctl --no-block reboot'
```

systemd/coreos-installer-secure-ipl-reboot.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Reboot from zVM LOADDEV after CoreOS Installer
```
jlebon, line 2: The unit is designed to handle the LPAR case, so we might as well reflect that in its description. E.g.

```suggestion
Description=Reboot in Secure IPL mode after CoreOS Installer
```

systemd/coreos-installer-secure-ipl-reboot.service
```diff
@@ -0,0 +1,14 @@
...
+Requires=coreos-installer.target
+After=coreos-installer.target
+OnFailure=emergency.target
+OnFailureJobMode=replace-irreversibly
+ConditionPathExists=/run/coreos-installer-reboot
+ConditionKernelCommandLine=coreos.inst.secure_ipl
+
+[Service]
+Type=simple
+ExecStart=/bin/sh -c '[ -e /dev/vmcp ] && vmcp ipl loaddev || systemctl --no-block reboot'
```
jlebon, line 12: Minor: a comment here about z/VM and LPAR would be useful. E.g.

> On z/VM, we need to use `vmcp ipl loaddev`. Otherwise, for the LPAR case, we can just reboot as usual.

?

jlebon (approved):
Some minor comments but LGTM as is too.

systemd/coreos-installer-secure-ipl-reboot.service
nikita-dubrovskii: Done

src/s390x/zipl.rs
```diff
@@ -33,11 +34,116 @@ pub fn chreipl<P: AsRef<Path>>(dev: P) -> Result<()> {
...
+fn parse_lszdev_eckd(line: &str) -> Result<Loaddev> {
+    // ECKD ID looks like: 0.0.5223, we need only last part of it (5223)
+    lazy_static! {
+        static ref REGEX: Regex = Regex::new(r#"[[:digit:]].[[:digit:]].([[:xdigit:]]+)"#).unwrap();
+    }
+    if let Some(cap) = REGEX.captures_iter(line).next() {
+        return Ok(Loaddev::Eckd(cap[1].to_string()));
+    }
+    bail!("bad ECKD id: {}", line);
+}
```
cgwalters, line 52: This is obviously totally fine, but just noting I think it's a one-liner to parse this without regexps:

```
line.split('.').nth(2).ok_or_else(|| anyhow!("Bad EKCD")).map(|s| Loaddev::Eckd(s.to_string()))
```
or so. (Though validation is looser)

nikita-dubrovskii: Yep, the reason to use regex here was to validate format of ID, which probably not that necessary.

PR #1360, opened 2023-12-15
install: support grub2/30_console.cfg for console replacement

dustymabe:
Upstream we are starting to use grub configs that are partially baked into bootupd [1] and partially baked into the image via the OS vendor (in our case a grub2/30_console.cfg file [2]). Let's look first to see if grub2/30_console.cfg exists when trying to write new console settings before falling back to grub2/grub.cfg.

[1] https://github.com/coreos/bootupd/pull/543
[2] https://github.com/coreos/fedora-coreos-config/pull/2769

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -636,11 +636,17 @@ fn write_console(mountpoint: &Path, platform: Option<&str>, consoles: &[Console]
 
     // set grub commands
     if grub_commands != metal_spec.grub_commands {
-        let path = mountpoint.join("grub2/grub.cfg");
-        let grub_cfg = fs::read_to_string(&path).context("reading grub.cfg")?;
+        // prefer the new grub2/30_console.cfg, but fallback to grub2/grub.cfg
+        let mut name = "grub2/30_console.cfg";
+        let mut path = mountpoint.join(name);
+        if !path.exists() {
+            name = "grub2/grub.cfg";
+            path = mountpoint.join(name);
+        }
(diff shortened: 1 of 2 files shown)


src/install.rs
```diff
@@ -636,11 +636,17 @@ fn write_console(mountpoint: &Path, platform: Option<&str>, consoles: &[Console]
...
-        let grub_cfg = fs::read_to_string(&path).context("reading grub.cfg")?;
+        // prefer the new grub2/50_coreos.cfg, but fallback to grub2/grub.cfg
+        let mut name = "grub2/50_coreos.cfg";
+        let mut path = mountpoint.join(name);
+        if !path.exists() {
+            name = "grub2/grub.cfg";
+            path = mountpoint.join(name);
+        }
+        let grub_cfg = fs::read_to_string(&path).context(format!("reading {}", name))?;
         let new_grub_cfg = update_grub_cfg_console_settings(&grub_cfg, &grub_commands)
```
cgwalters, line 647: Very very minor but style is usually to use `.with_context(|| format!(...))` to avoid string allocation even on non-error paths.

cgwalters (approved):
Looks sane to me

src/install.rs
dustymabe: ok I updated it.. shamefully I know very little about rust, but have been meaning to change that for a long time :(

dustymabe:
local tests with this are looking good.

prestist (approved):
LGTM

PR #1416, opened 2024-02-12
Dockerfile: rebase from f38 to f39

prestist:
https://github.com/coreos/fedora-coreos-tracker/issues/1490#issuecomment-1938656032


c4rt0 (approved):
Thanks @prestist 
lgtm

PR #1422, opened 2024-02-19
s390x: call zipl using exact kernel, ramdisk and cmdline

nikita-dubrovskii:
There is no need to copy `boot/loader/entries` folder and to create temporary `zipl.conf` , instead we could parse bls config and tell `zipl` which image and disk to use. This also fixes an issue when images comes with `bootprefix`.

Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1667

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -14,6 +14,7 @@ Minor changes:
 
 Internal changes:
 
+- zipl: Directly pass kernel, initrd and kargs instead of BLS entry
 
 Packaging changes:
 
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::blockdev::Mount;
-use crate::io::{visit_bls_entry, visit_bls_entry_options, Initrd, KargsEditor};
+use crate::io::{visit_bls_entry, Initrd};
 use crate::s390x::ZiplSecexMode;
 use crate::util::cmd_output;
 use crate::{runcmd, runcmd_output};
@@ -21,7 +21,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use lazy_static::lazy_static;
 use nix::mount::MsFlags;
 use regex::Regex;
-use std::fs::{copy, create_dir_all, read_dir, DirEntry, File};
+use std::fs::{read_dir, DirEntry, File};
 use std::io::{Read, Write};
 use std::path::{Path, PathBuf};
 use std::process::Command;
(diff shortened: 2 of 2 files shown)


jmarrero (approved):
this looks good to me, thanks for digging on this @nikita-dubrovskii 

@prestist can you give this another look too?

src/s390x/zipl.rs
```diff
@@ -330,73 +330,37 @@ pub fn zipl<P: AsRef<Path>>(
...
-                        &firstboot_kargs
-                            .split_whitespace()
-                            .map(|s| s.to_string())
-                            .collect::<Vec<String>>(),
-                    );
-                }
+        let (kernel, initrd, mut options) = get_info_from_bls(boot)?;
+        // we need a full path to kernel and initrd
+        let kernel = boot.join(&kernel[1..]);
+        let initrd = boot.join(&initrd[1..]);
```
dustymabe, line 338: and because we have the `boot -> .` [symlink](https://github.com/osbuild/osbuild/issues/1566#issuecomment-1921908500) it doesan't matter if the ultimate path ends up being `/boot/boot/ostree/foo/vmlinuz...` ?

nikita-dubrovskii: yes, exactly, symlink does the trick

docs/release-notes.md
```diff
@@ -14,6 +14,7 @@ Minor changes:
 
 Internal changes:
 
+- zipl: specify exact kernel, ramdisk and cmdline instead of copying & modifying of bls config
```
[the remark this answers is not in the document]
dustymabe: No yes. We do need a release notes entry.

My question is whether or not we should include a link in the release notes entry to https://github.com/coreos/fedora-coreos-tracker/issues/1667

(sorry the link I pasted in https://github.com/coreos/coreos-installer/pull/1422#discussion_r1494677156 was wrong. I should have linked https://github.com/coreos/fedora-coreos-tracker/issues/1667)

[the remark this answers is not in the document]
nikita-dubrovskii: I don't think so. Commit message contains the link, but let's wait what @jlebon says about that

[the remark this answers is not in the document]
dustymabe: ehh. doesn't need to wait. we can just ship it if everyone else is happy with it

dustymabe (approved):

docs/release-notes.md
jlebon, line 17: Minor:

```suggestion
- zipl: Directly pass kernel, initrd and kargs instead of BLS entry
```

src/s390x/zipl.rs
```diff
@@ -330,73 +330,37 @@ pub fn zipl<P: AsRef<Path>>(
...
-                if let Some(firstboot_kargs) = extract_firstboot_kargs(&firstboot_contents)? {
-                    extra.extend_from_slice(
-                        &firstboot_kargs
-                            .split_whitespace()
-                            .map(|s| s.to_string())
-                            .collect::<Vec<String>>(),
-                    );
-                }
+        let (kernel, initrd, mut options) = get_info_from_bls(boot)?;
+        // we need a full path to kernel and initrd
```
jlebon, line 336: Shouldn't this be e.g.

```suggestion
        // we need a relative path to kernel and initrd
```

?

jlebon: Does this then work because `--image` and `--ramdisk` are looked for relative to `--target`?

src/s390x/zipl.rs
```diff
@@ -330,73 +330,37 @@ pub fn zipl<P: AsRef<Path>>(
...
+        if let Some(kargs) = kargs {
+            options = format!("{options} {kargs}");
+        }
 
-        runcmd!("zipl", "--blsdir", blsdir, "--config", conffile.path())
+        runcmd!(
+            "zipl",
+            "-V",
+            "--target",
+            boot.to_str().unwrap(),
```
jlebon, line 356:
```suggestion
            boot.as_ref(),
```

jlebon (approved):
Some minor details, but looks sane overall!

src/s390x/zipl.rs
dustymabe: I'm confused.

On line 337 and 338 below he is making the paths full absolute paths, which matches his comment of `we need a full path to kernel and initrd`.

jlebon: Right yes, I misread those lines.

dustymabe: Does that resolve this line of questioning then?

prestist (approved):
Looks good to me, nothing major that I see needs to change. Agree with most of the nits.

jlebon (approved):

PR #1423, opened 2024-02-19
install: use console.cfg rather than 30_console.cfg

dustymabe:
In https://github.com/coreos/fedora-coreos-tracker/issues/1671 we decided to use a non-static grub file for console configuration named console.cfg.


dustymabe:
Creating as draft because I still need to test this in concert with other changes

dustymabe:
ok this is looking good in testing I think.

dustymabe:
resolved conflicts

jlebon (approved):

PR #1424, opened 2024-02-20
signing-keys: add Fedora 41 key, drop Fedora 38 key

marmijo:
We branched F40 from rawhide so rawhide is now F41.


dustymabe (approved):
LGTM

PR #1428, opened 2024-02-22
packit: update dist_git_branches to target fedora-all

prestist:
After the latest release we noticed that f40 was not included, this is because its not stable yet. which means that it would be ignored by packit.


dustymabe (approved):
LGTM

jlebon (approved):

PR #1473, opened 2024-05-21
rootmap: fix multipath + LUKS case

jlebon:
In the multipath + LUKS case, `get_luks_uuid()` would incorrectly
skip over the multipath partition containing the LUKS header because
`is_dm_device()` returned true. The code eventually errors out when it
gets to the disks backing the multipath device.

The `is_dm_device()` check was added as part of 69b706d ("rootmap:
handle filesystems with LUKS integrity") to correctly handle the LUKS
integrity case in the Secure Execution path. There, the device right
under the LUKS device is another crypt device mapper device used for
integrity that we need to skip over.

Instead of generically checking for a device mapper target, check
specifically that it's a LUKS integrity target before deciding to skip.

Part of: https://github.com/coreos/fedora-coreos-tracker/issues/1728

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -14,6 +14,7 @@ Minor changes:
 
 Internal changes:
 
+- rootmap: Fix support for LUKS on multipath
 
 Packaging changes:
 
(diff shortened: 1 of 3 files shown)


jlebon:
CI is currently broken (see also https://github.com/coreos/coreos-installer/pull/1474).

jlebon:
This is ready for review now!

jmarrero (approved):
lgtm

yasminvalim (approved):
LGTM ✅

Issue #1469, opened 2024-05-14
coreos-installer iso customize【--dest-ignition】be accompanied by the 【--dest-device】


duyiwei7w:
# Bug #
coreos-installer iso customize【--dest-ignition】be accompanied by the 【--dest-device】

## Host Operating System Version ##
NestOS 【a variant of CoreOS devrloping in openEuler】

## coreos-installer Version ##
0.16.0

## Expected Behavior ##
use " coreos-installer iso customize my-iso --dest-ignition 【ign file path】-o test.iso "
I can successfully boot test,iso

## Actual Behavior ##
can't boot test.iso
coreos-installer service failed.  need <DEVICE> information ,such as /dev/sda

## Solve Steps ##

use “coreos-installer iso customize my-iso --dest-ignition 【ign file path】--dest-device /dev/sda -o test.iso "
Boot test.iso successfully!

## Other Information ##

I prefer to categorize this issue under the direction of user convenience. In fact, I don't consider it a bug, much like when we directly boot, we need to execute coreos-installer install /dev/sda --ignition-file [ign file path].
However, I still feel that providing some hints to users or forcing users to use both commands simultaneously when using customize may be beneficial. Please have the maintainer review this issue, and if it needs to be modified, I'm willing to give it a try.

duyiwei7w:
In addition, I would like to inquire about how to use the ISO image generated by coreos-install iso ignition embed -i test.ign -o test.iso. When I directly boot it, I find it is still in live mode, which is somewhat similar to coreos-installer iso customize --live-ignition.

jlebon:
> coreos-installer iso customize【--dest-ignition】must be accompanied by the 【--dest-device】

The reason this is not enforced is that you can also provide the destination device through other means, e.g. via kargs or via installer configs.

That said, I think we could provide a message when doing this that we expect the destination device to be provided through other means.

> In addition, I would like to inquire about how to use the ISO image generated by `coreos-install iso ignition embed -i test.ign -o test.iso`. When I directly boot it, I find it is still in live mode, which is somewhat similar to `coreos-installer iso customize --live-ignition.`

Correct. `iso ignition embed` predates the more friendly `iso customize` and its `--live-ignition` switch.

duyiwei7w:
> The reason this is not enforced is that you can also provide the destination device through other means, e.g. via kargs or via installer configs.
> That said, I think we could provide a message when doing this that we expect the destination device to be provided through other means.

[...]

PR #1480, opened 2024-06-11
cmdline: Provide dest_device when using iso/pxe customize

duyiwei7w:
During the execution of coreos-installer install, users are required to provide dest-device through parameters or config file. However, when executing iso/pxe customize, it is not necessary to provide dest-device to successfully execute customize. This will obviously result in errors when using a customized image for installation. Some previous discussions can be found at: 

https://github.com/coreos/coreos-installer/issues/1469

This modification unifies the behavior of the customize command and the install command during the image install phase, requiring users to provide dest-device through parameters or config files.

### DIFF
--- a/src/cmdline/mod.rs
+++ b/src/cmdline/mod.rs
@@ -260,9 +260,9 @@ pub struct CommonCustomizeConfig {
     /// Install destination device
     ///
     /// Automatically run installer, installing to the specified destination
-    /// device.  The resulting boot media will overwrite the destination
-    /// device without confirmation.
-    #[arg(long, value_name = "path")]
+    /// device that the user must provide.  The resulting boot media will
+    /// overwrite the destination device without confirmation.
+    #[arg(long, value_name = "path", required_unless_present = "installer_config")]
     pub dest_device: Option<String>,
     /// Kernel and bootloader console for dest
     ///


jlebon:
Kernel arguments can also be used to specify the destination device (`coreos.inst.install_dev`).

How about instead as soon as any `--dest` argument is passed but not `--dest-device` we print a note that we expect the destination device to be passed either through a later `customize` call or via `coreos.inst.install_dev`?

duyiwei7w:
Thank you for your suggestion @jlebon . As far as I know, if you use `coreos-installer install` parameters, in addition to directly specifying `dest_device` or using `configfile`. We can also use kernel parameters to pass the destination device .

So I want to make as little modification as possible in customize, and keep the behavior basically consistent with the coreos-installer install concept.

That is, the user is required to enter `--dest-device` by default, and a prompt is printed if there is no dest_device. If configfile is used, dest_device does not need to be specified. For advanced kernel parameters, there is no unified requirement, and it will not detect whether coreos.inst.install_dev is explicitly specified.

duyiwei7w:
Also, as you said, if the user enters any` --dest` parameter but no `--dest-device`, we print a prompt. But if the user passes the target disk through` installerconfig` or `kernel parameters`, then this prompt is unnecessary.

In other words, we may need to parse all the customize parameters entered by the user again to confirm that they have not been passed to the destination device  before printing the prompt. In fact, it is better to directly ask the user to use -`-dest-device` as mentioned above.

jlebon:
We need to be flexible in how the ISO can be used. For example, you can imagine someone providing a pre-customized ISO/PXE initramfs to a user, with the understanding that the user will customize it further with the actual destination device (or e.g. add the needed karg in their PXE config).

But note also that the `customize` command can also be used to customize a live ISO that will be used solely for its live environment, with no intent of installing a CoreOS system.

PR #1494, opened 2024-07-24
packit: downstream_package_name is not inherited by packages

prestist:
Declare downstream_package_name for each package; i.e fedora and centos.


yasminvalim (approved):
LGTM ✅

PR #1511, opened 2024-08-12
Use  blkid instead of lsblk to get disk info when udev is not present

ravanelli:
- OSBuild does not allow udev rules in loop devices, for this reason can not rely on labels on the partitions to create the osmet image, we
need to skip this part and pass only the fstype
for the it to be mounted.

### DIFF
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -109,19 +109,34 @@ impl Disk {
                 if devinfo.get("TYPE").map(|s| s.as_str()) != Some("part") {
                     continue;
                 }
-                let (mountpoint, swap) = match devinfo.get("MOUNTPOINT") {
-                    Some(mp) if mp == "[SWAP]" => (None, true),
-                    Some(mp) => (Some(mp.to_string()), false),
-                    None => (None, false),
-                };
-                result.push(Partition {
-                    path: name.to_owned(),
-                    label: devinfo.get("LABEL").map(<_>::to_string),
-                    fstype: devinfo.get("FSTYPE").map(<_>::to_string),
-                    parent: self.path.to_owned(),
-                    mountpoint,
-                    swap,
-                });
+                // only trust lsblk output for the following fields if we have udev
+                if have_udev() {
+                    let (mountpoint, swap) = match devinfo.get("MOUNTPOINT") {
+                        Some(mp) if mp == "[SWAP]" => (None, true),
+                        Some(mp) => (Some(mp.to_string()), false),
+                        None => (None, false),
+                    };
+                    result.push(Partition {
+                        path: name.to_owned(),
+                        label: devinfo.get("LABEL").map(<_>::to_string),
+                        fstype: devinfo.get("FSTYPE").map(<_>::to_string),
+                        parent: self.path.to_owned(),
+                        mountpoint,
+                        swap,
+                    });
+                } else {
+                    let devinfo = blkid_single(Path::new(name))?;
+                    // note TYPE here: blkid uses TYPE instead of FSTYPE
+                    let fstype = devinfo.get("TYPE").map(<_>::to_string);
+                    result.push(Partition {
+                        path: name.to_owned(),
+                        label: devinfo.get("LABEL").map(<_>::to_string),
+                        fstype: fstype.to_owned(),
+                        parent: self.path.to_owned(),
+                        mountpoint: None,
+                        swap: fstype.is_some_and(|s| s == "swap"),
+                    });
+                }
             }
         }
         Ok(result)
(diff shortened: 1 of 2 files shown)


src/osmet/mod.rs
```diff
@@ -93,10 +93,17 @@ pub fn pack_osmet(config: PackOsmetConfig) -> Result<()> {
...
     let disk = Disk::new(&config.device)?;
-    let boot = disk.mount_partition_by_label("boot", mount::MsFlags::MS_RDONLY)?;
-    let root = disk.mount_partition_by_label("root", mount::MsFlags::MS_RDONLY)?;
-
     // now, we do a first scan of the boot partition and pick up files over a certain size
+    let boot;
+    let root;
+    if !config.root.trim().is_empty() && !config.boot.trim().is_empty() {
+        boot = disk.mount_partition_by_label(&config.boot, false, mount::MsFlags::empty())?;
+        root = disk.mount_partition_by_label(&config.root, false, mount::MsFlags::empty())?;
```
jlebon, line 101: Hmm, I think it's confusing to have `mount_partition_by_label` take either a label _or_ a full device path.

jlebon (commented):
Hmm, how about instead, we keep the interface the same (no new command-line arguments), but in `get_partitions`, we check if udev is present (e.g. check for `/run/udev/control`; we have code that does something similar already), and if so then use `lsblk()`, otherwise use `blkid()`.

Both return a `Result<Vec<HashMap<String, String>>>` we can iterate over. But of course in the blkid case, some of the key names to pull from will be different.

jlebon:
> Both return a `Result<Vec<HashMap<String, String>>>` we can iterate over. But of course in the blkid case, some of the key names to pull from will be different.

Though I think this is glossing over an important detail. `blkid()` will return information about all devices, whereas `lsblk()` only returns information about the passed device. Since we can still use `lsblk()` at least to get the list of partitions on the disk, we can then call a `blkid_single()` that only does blkid on one device (or just change `blkid()` to take an `dev: Option<String>`) to get the info we need. So basically something like:

```rust
fn get_partitions(&self) -> Result<Vec<Partition>> {
    // walk each device in the output
    let mut result: Vec<Partition> = Vec::new();
    for devinfo in lsblk(Path::new(&self.path), true)? {
        if let Some(name) = devinfo.get("NAME") {
            // Only return partitions.  Skip the whole-disk device, as well
            // as holders like LVM or RAID devices using one of the partitions.
            if devinfo.get("TYPE").map(|s| s.as_str()) != Some("part") {
                continue;
            }
            // only trust lsblk output for the following fields if we have udev
            if have_udev() {
                let (mountpoint, swap) = match devinfo.get("MOUNTPOINT") {
                    Some(mp) if mp == "[SWAP]" => (None, true),
                    Some(mp) => (Some(mp.to_string()), false),
                    None => (None, false),
                };
                result.push(Partition {
                    path: name.to_owned(),
                    label: devinfo.get("LABEL").map(<_>::to_string),
                    fstype: devinfo.get("FSTYPE").map(<_>::to_string),
                    parent: self.path.to_owned(),
                    mountpoint,
                    swap,
                });
            } else {
                let devinfo = blkid_single(name)?;
                // note TYPE here: blkid uses TYPE instead of FSTYPE
                let fstype = devinfo.get("TYPE").map(<_>::to_string);
                result.push(Partition {
                    path: name.to_owned(),
                    label: devinfo.get("LABEL").map(<_>::to_string),
                    fstype,
                    parent: self.path.to_owned(),
                    None,
                    fstype.is_some_and(|s| s == "swap"),
                });
            }
        }
    }
    Ok(result)
}
```

ravanelli:
> Though I think this is glossing over an important detail. `blkid()`

Great, in the current code I used bklid for a single device as well adding the dev path:
```
     Command::new("blkid");
      cmd.arg(path);
 ```
 Fixing the rest ;)

File: src/blockdev.rs [hunk 2]
```diff
@@ -893,19 +915,31 @@ fn blkid() -> Result<Vec<HashMap<String, String>>> {
...
+            cmd.arg("--cache-file");
+            cmd.arg("/dev/null");
+            cmd.arg("-o");
+            cmd.arg("device");
+            cmd_output(&mut cmd)?
+        };
 
+        cmd_dev.extend(devices
+            .lines()
+            .filter_map(|line|Some(Path::new(line.trim())))
```
jlebon, line 932: Can just use `map` here if we don't need filtering functionality.

src/blockdev.rs
```diff
@@ -1141,10 +1175,13 @@ fn disk_has_mbr(file: &mut (impl Read + Seek)) -> Result<bool> {
     Ok(sig == [0x55, 0xaa])
 }
 
+pub fn have_udev() -> bool {
+    Path::new("/run/udev/control").exists()
+}
```
jlebon, line 1180: Minor: spacing

src/blockdev.rs
```diff
@@ -893,19 +915,31 @@ fn blkid() -> Result<Vec<HashMap<String, String>>> {
...
 
+        cmd_dev.extend(devices
+            .lines()
+            .filter_map(|line|Some(Path::new(line.trim())))
+        );
+    } else {
+        if let Some(path) = dev {
+            cmd_dev.extend(path);
+        }
+    }
```
jlebon, line 938: A more idiomatic way to restructure all this is:

```rust
let cmd_dev = if let Some(paths) = dev {
    paths
} else {
    let devices = {
        ...
    };
    devices.lines().map(|line| Path::new(line.trim())).collect()
};
```

ravanelli: Thanks for the free rust class @jlebon ;)

jlebon (approved):
Just added spacing and tweaked some of the variable names.

LGTM!

PR #1516, opened 2024-08-20
packit: drop CentOS packaging

prestist:
While releasing coreos-installer, we found that the current CentOS packaging did not introduce enough benefit to outweigh the process change required to keep other repos consistent. We will revisit once additional support is added for running builds/tests/ticketing.

Ref: https://github.com/packit/packit-service/issues/2106


yasminvalim (approved):
/lgtm

travier (approved):

PR #1592, opened 2025-01-28
DNM: github/workflows: Try with --security-opt=seccomp=unconfined

travier:


travier:
```
Insufficient rights.
```

🤔

jlebon:
Hmm, it seems like it got past the problem spot though at least?

PR #1597, opened 2025-02-03
s390x: prepare for the upcoming 'pvimg' tool replacing 'genprotimg'

nikita-dubrovskii:
Since s390-tools commit [f4cf4ae6ebb1](https://github.com/ibm-s390-linux/s390-tools/commit/195579cf0bcff320ed9ebe2cab6b59c9cc817f6e)  (Remove genprotimg-C and switch to genprotimg-Rust implementation) 
FCCOS/RHCOS may come with symlink to `pvimg create`. 
New `pvimg` made some cmdline options obsolete (`-V`) and by default doesn't overwrite the output image (`sdboot`), which leads to an issue https://github.com/openshift/os/issues/1731.

For now keep the call to `genprotomg`, even if it's symlink, but let's:
 - remove the `sdboot` before generating new one
 - use long cmdline options supported by both tool.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -13,7 +13,7 @@ Minor changes:
 
 
 Internal changes:
-
+- s390x: Use options and logic compatible with both C-based `genprotimg` and Rust-based `pvimg`
 
 Packaging changes:
 
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -281,6 +281,13 @@ fn generate_sdboot(
 
     // finally, Secure Execution sd-boot image
     let sdboot = mountpoint.join("sdboot");
+
+    // C 'genprotimg' tool no longer exists and was replaced by symlink to Rust 'pvimg create',
+    // which by default doesn't overwrite the output image.
+    // For backward compatibility let's silently remove the 'sdboot'.
+    let _ = std::fs::remove_file(&sdboot);
+
+    // FIXME: in F42/el10 switch to 'pvimg create' with '--overwrite' flag.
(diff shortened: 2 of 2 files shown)


src/s390x/zipl.rs
```diff
@@ -281,16 +281,23 @@ fn generate_sdboot(
 
     // finally, Secure Execution sd-boot image
     let sdboot = mountpoint.join("sdboot");
+
+    // C 'genprotimg' tool no longer exists and was replaced by symlink to Rust 'pvimg create',
+    // which by default doens't overwrite the output image.
```
travier, line 286:
```suggestion
    // which by default doesn't overwrite the output image.
```

docs/release-notes.md
```diff
@@ -13,7 +13,7 @@ Minor changes:
 
 
 Internal changes:
-
+- s390x: remove 'sdboot' before generating new one
```
travier, line 16:
```suggestion
- s390x: Use options and logic compatible with both `genprotimg` implementations (C and Rust)
```

travier (commented):
Some nits but mostly LGTM

src/s390x/zipl.rs
nikita-dubrovskii: Good catch! Thx!

hbrueckner (approved):
/lgtm
Thanks!

travier (approved):
I only have limited context and did not test this but LGTM

nikita-dubrovskii:
> I only have limited context and did not test this but LGTM

No worries, i've tested it locally

mhartmay:
LGTM, thanks for taking care!

PR #1598, opened 2025-02-06
signing-keys: add Fedora 43 key, drop Fedora 40 key

aaradhak:
Fedora 42 branched from rawhide, so rawhide is now F43.
Updated the release-notes doc on signing keys to add the F43 key and drop the F40 key.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1851


dustymabe (approved):
LGTM

PR #1609, opened 2025-02-20
cargo: add feature flag which disables nmstate by default

prestist:
With the latest release for f42 dependencies which are required for nmstate will no longer be available.
There is a PR which has landed https://github.com/nmstate/nmstate/pull/2820 but a build of nmstate needs to land with those changes in f42.

To prevent issues for coreos-installer add a featureflag to make nmstate dependency an optional one. If the feature 'use-nmstate' is not available coreos-installer will lose functionality around customizing network settings. By adding the feature 'use-nmstate' functionality would be restored.


dustymabe:
Should we close this? Or still think it's worth merging?

PR #1618, opened 2025-02-26
Cargo.toml: bump MSRV to 1.84.1

HuijingHei:


prestist (changes requested):
Thank you for taking this, this looks good; I just have a small nit,  and we should also bump it to 1.84.1

Can we combine both the first commit and second into one commit with the commit message being  something like this "cargo: bump MSRV to 1.84.1"

HuijingHei:
Updated, welcome to review, thanks!

prestist (approved):
LGTM, thank you!

PR #1648, opened 2025-04-17
.cci.jenkinsfile: switch to `cosa osbuild`

dustymabe:
These can now be built in a single invocation so let's do that.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -240,7 +240,7 @@ jobs:
       - name: Cache build artifacts
         uses: Swatinem/rust-cache@v2
       - name: Install dependencies
-        run: dnf install -y gcc git-core libzstd-devel openssl-devel util-linux awk
+        run: dnf install -y gcc git-core libzstd-devel openssl-devel util-linux util-linux-script awk
       - name: cargo build
         run: cargo build
       - name: Help text line length
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -52,7 +52,7 @@ jobs:
           # Specify additional fields not implied by the job name.
           key: ${{ matrix.arch }}
       - name: Install dependencies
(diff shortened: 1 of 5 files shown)


yasminvalim (approved):
/lgtm

marmijo (approved):

dustymabe:
Pushed up a second commit which i think will fix a test that was failing.

For the Rust CI failures; this PR obviously has nothing to do with Rust so I'll ignore them.

jlebon:
I think GitHub Actions CI just needs awk added to https://github.com/coreos/coreos-installer/blob/6e6f3f23359d344abefdbb1d6237e7f1c146f6b3/.github/workflows/rust.yml#L55.

.github/workflows/rust.yml
```diff
@@ -240,7 +240,7 @@ jobs:
       - name: Cache build artifacts
         uses: Swatinem/rust-cache@v2
       - name: Install dependencies
-        run: dnf install -y gcc git-core libzstd-devel openssl-devel util-linux awk
+        run: dnf install -y gcc git-core libzstd-devel openssl-devel util-linux util-linux-script awk
```
dustymabe, line 243: is util-linux-script an RPM? my system seems to not think so.

prestist: Odd, it is on mine. I tired to run with util-linux rpm only and got that the "script" command was not found. I then Locally tried to run a script and got the suggestion to install util-linux-script and then locally that worked, so I added it to the ci, (which it seems to have fixed the docs)

dustymabe: ok. it was broken out to a subpackage in F42, which explains why I didn't see it (i'm on F41).

👍

dustymabe:
ok @prestist - this one should be ready to go now that I've fixed all the little papercut issues from the filenames of the ISO/kernel changing.

PR #1677, opened 2025-07-09
rootmap: use full path for 'root=' karg when rootfs is directly on multipath

nikita-dubrovskii:
Issue: https://github.com/coreos/fedora-coreos-tracker/issues/1980

This was first observed on s390x builders under high system load, where `ext.config.multipath.resilient` would intermittently fail during subsequent boot:
```
[    2.781559] multipathd[321]: sdd [8:48]: path added to devmap 0xcadf6fadb3ee446d
[    2.853163] multipathd[321]: sdb [8:16]: path added to devmap 0x000000000000000b
[    3.012431] systemd[1]: Reached target coreos-multipath-wait.target - CoreOS Wait For Multipathed Boot.
[    3.139605] systemd[1]: Mounting sysroot.mount - /sysroot...
[    3.450666] mount[806]: mount: /sysroot: fsconfig system call failed: /dev/sdd4: Can't open blockdev.
```

It looks like a race condition between multipathd taking ownership of the root device and systemd trying to mount /sysroot. This might happen because the udev database isn't ready in time.


src/bin/rdcore/rootmap.rs
```diff
@@ -30,6 +30,20 @@ use crate::cmdline::*;
 /// composefs, the mount of / and /sysroot will be distinct.
 const PHYSICAL_ROOT_MOUNT: &str = "sysroot";
 
+fn is_on_multipath(device: &str) -> Result<bool> {
```
jlebon, line 33: Minor: Let's move this lower down near the other utility functions?

src/bin/rdcore/rootmap.rs
```diff
@@ -30,6 +30,20 @@ use crate::cmdline::*;
...
+    let on_mpath = find_parent_devices(device)?
```
jlebon, line 34: Hmm, I don't think the logic is quite right here. We shouldn't check if any of the parent devices is multipath, but rather whether the immediate parent is.

We already have `backing_devices` which is a list of all the parent devices in order from lowest to highest (ending with the final device containing the rootfs itself). So... I think basically:
- get the _before last_ device in that Vec
- call `lsblk_single`, check if `TYPE` is `mpath`

?

jlebon (approved):
Thanks for working on this! This is actually part of a larger story around being more resilient to multiple `root` labels (https://github.com/openshift/os/issues/1787).

PR #1682, opened 2025-07-19
.cci.Jenkinsfile: don't use dev profile for CI testing

dustymabe:
The debug binaries are much larger and we were running out of
space in some tests such as the `ext.config.rpm-ostree.kernel-replace`
test.

Use the release profile here which creates much smaller binaries.


jlebon (approved):
Thanks!

PR #1692, opened 2025-09-02
Uses blkid -p instead of lsblk to bypass the cache

yasminvalim:
**BUG REPORTED**:

  OCP 4.17 Single node subsequent reboot after initial install fails. 

**SOLUTION:**
Switch from using `lsblk` to `blkid -p` in the `get_filesystem_uuid()` function to bypass cached device information and retrieve fresh filesystem UUIDs.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -21,6 +21,7 @@ Internal changes:
 - Support cryptsetup-2.8.0's UUIDs naming of dm-integrity devices
 - rootmap: Inject `root=/dev/disk/by-uuid/dm-mpath-$UUID` when on multipath
 - rootmap: Inject `mpath.wwid=$WWID` when on multipath
+- rootmap: Bypass cache when querying filesystem UUID for `root karg` 
 
 Packaging changes:
 
--- a/src/blockdev.rs
+++ b/src/blockdev.rs
@@ -496,7 +496,9 @@ impl Mount {
     }
 
     pub fn get_filesystem_uuid(&self) -> Result<String> {
-        let devinfo = lsblk_single(Path::new(&self.device))?;
+        // We used to use lsblk_single, but its cache may be stale after mkfs.
+        // blkid_single doesn't use cache.
+        let devinfo = blkid_single(Path::new(&self.device))?;
         devinfo
             .get("UUID")
             .map(String::from)


src/bin/rdcore/rootmap.rs
```diff
@@ -39,6 +39,12 @@ pub fn rootmap(config: RootmapConfig) -> Result<()> {
     let physical_mount = Mount::from_existing(physical_root_path)?;
     let device = PathBuf::from(physical_mount.device());
 
+    // Force reread to get updated UUIDs and ensure correct UUID after partition resizing
+    if let Ok(mut fd) = std::fs::File::open(&device) {
+        let _ = reread_partition_table(&mut fd, false);
+    }
```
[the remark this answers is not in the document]
yasminvalim: I guess it makes sense. I think I could apply. 🤔

jlebon (commented):
Hmm, I don't think rereading the partition table is the fix for https://issues.redhat.com/browse/OCPBUGS-57573. The issue there has nothing to do with a stale partition table. The issue is that the rootfs filesystem UUID is stale.

To quote my comment from that bug:

> Note to CoreOS devs: I think the easiest fix for this is to switch the rdcore rootmap code to use the blkid helper for this instead of lsblk: https://github.com/coreos/coreos-installer/blob/d8c6a212b1870ef3d809ff38e7e252aa4a047b7c/src/blockdev.rs#L818

I.e. `get_filesystem_uuid()` is what we call below when generating the `root=` karg; that function currently calls `lsblk_single`, but let's have it instead call `blkid_single` which uses `blkid -p` to bypass the cache.

src/blockdev.rs
```diff
@@ -496,7 +496,7 @@ impl Mount {
     }
 
     pub fn get_filesystem_uuid(&self) -> Result<String> {
-        let devinfo = lsblk_single(Path::new(&self.device))?;
+        let devinfo = blkid_single(Path::new(&self.device))?;
```
jlebon, line 499: Maybe let's add a comment here like:

> Note we used to use `lsblk_single` here, but the filesystem UUID in the cache may be stale if we just `mkfs`'ed it. `blkid_single` doesn't use any cache.

?

docs/release-notes.md
```diff
@@ -13,6 +13,7 @@ Major changes:
 Minor changes:
 
 - Restore formatting of progress reporting to pre 0.24.0 behavior.
+- Fix UUID mismatch bug using `blkid -p` instead of `lsblk` to bypass the cache
```
jlebon, line 16: `rdcore` changes usually go in "Internal changes" instead.

Also, how about:

> rootmap: Bypass cache when querying filesystem UUID for `root` karg

?

PR #1696, opened 2025-09-29
Pre release 0.25.0

yasminvalim:
https://github.com/coreos/coreos-installer/issues/1695


prestist:
I would make it a seprate commit but I think the update to dep's has resulted in the output to change from doc gen. 

You should be able to re gen them by running "make docs"

prestist (commented):
LGTM, lets update the final commit to say 

" man/*: re-generate docs after deps update"
wdyt?

Rather then
" docs: rerun make docs to update docs to the right format after updati…

…ng dependencies"

Issue #1171, opened 2023-04-20
`coreos-installer install` tries to modify `grub.cfg` to apply console settings even on s390x


jlebon:
# Bug #

## coreos-installer Version ##

v0.17.0

## Expected Behavior ##

On s390x, since there's no GRUB config, using `coreos-installer install --console` or the `console` key in installer dropins only sets the appropriate karg and doesn't try to look for a GRUB config.

## Actual Behavior ##

On s390x, `coreos-installer install --console` or the `console` key in installer dropins tries to look for a GRUB config that doesn't exist and fails the install.

## Reproduction Steps ##

  1. Run `coreos-installer install --console` on s390x.

## Other Information ##

Output:

```
coreos-installer install --config-file /etc/coreos/installer.d/mantle.yaml --fetch-retries infinite
Running with arguments: --image-url http://10.0.2.2:34353/fedora-coreos-38.20230419.dev.0-metal.s390x.raw.xz --ignition-file /var/opt/pointer.ign --console ttysclp0 --insecure /dev/vda --fetch -retries infinite
Downloading image from http://10.0.2.2:34353/fedora-coreos-38.20230419.dev.0-metal.s390x.raw.xz
Downloading signature from http://10.0.2.2:34353/fedora-coreos-38.20230419.dev.0-metal.s390x.raw.xz.sig
Failed to fetch signature: fetching signature URL
Signature not found; skipping verification as requested
...
Read disk 479.3 MiB/479.3 MiB (100%)
Writing Ignition config
Error: configuring console
Caused by:
    0: reading grub.cfg
    1: No such file or directory (os error 2)
Resetting partition table
Error: install failed
```

nikita-dubrovskii:
Fixed in upcoming coreos-installer release

PR #1732, opened 2026-03-12
install: Fix --console flag on s390x

nikita-dubrovskii:
On s390x, installations with `--console` would fail because the installer
tried to modify GRUB config files that don't exist (s390x uses `zipl`).
Skip `write_console()` on s390x and apply console kernel arguments
directly to BLS entries instead. Also refactor karg modification logic
into a dedicated function for better readability.

Issue: https://github.com/coreos/coreos-installer/issues/1171

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -14,9 +14,6 @@
 
 use anyhow::{bail, Context, Result};
 use nix::mount;
-use regex::{Captures, Regex};
-use serde::Deserialize;
-use std::collections::HashMap;
 use std::fs::{self, File, OpenOptions, Permissions};
 use std::io::{self, BufReader, Seek, SeekFrom, Write};
 use std::num::NonZeroU32;
@@ -31,8 +28,16 @@ use crate::io::*;
 use crate::s390x;
 use crate::source::*;
 
+#[cfg(not(target_arch = "s390x"))]
+use {
+    regex::{Captures, Regex},
+    serde::Deserialize,
+    std::collections::HashMap,
+};
+
 // Match the grub.cfg console settings commands in
 // https://github.com/coreos/coreos-assembler/blob/main/src/grub.cfg
+#[cfg(not(target_arch = "s390x"))]
 const GRUB_CFG_CONSOLE_SETTINGS_RE: &str = r"(?P<prefix>\n# CONSOLE-SETTINGS-START\n)(?P<commands>([^\n]*\n)*)(?P<suffix># CONSOLE-SETTINGS-END\n)";
 
 pub fn install(config: InstallConfig) -> Result<()> {
@@ -417,6 +422,10 @@ fn write_disk(
         if let Some(platform) = config.platform.as_ref() {
             write_platform(mount.mountpoint(), platform).context("writing platform ID")?;
         }
+
+        // On s390x, "coreos-installer install --console" tries to look for a GRUB config that doesn't exist and fails the install.
+        // Also s390x doesn't have any configuration in platforms.yaml, so platforms.json is not included in the image
+        #[cfg(not(target_arch = "s390x"))]
         if config.platform.is_some() || !config.console.is_empty() {
             write_console(
                 mount.mountpoint(),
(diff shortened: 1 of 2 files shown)


nikita-dubrovskii:
Checked on s390x, `ttyS123` was successfully added:
```
$coreos-installer install --console ttyS123 /dev/vdb
Downloading Fedora CoreOS stable s390x metal image (raw.xz) and signature
> Read disk 582.4 MiB/582.4 MiB (100%)   
gpg: Signature made Sat Feb 28 07:39:40 2026 UTC
gpg:                using RSA key C6E7F081CF80E13146676E88829B606631645531
gpg: Good signature from "Fedora (43) <fedora-43-primary@fedoraproject.org>" [ultimate]
Adding console kernel arguments
Looking for components in '/lib/s390-tools'
Secure boot support: no
Target device information
  Device..........................: fd:10
  Partition.......................: fd:13
  Device name.....................: vdb
  Device driver name..............: virtblk
  Type............................: disk partition
  Disk layout.....................: SCSI disk layout
  Geometry - start................: 2048
  File system block size..........: 1024
  Physical block size.............: 512
  Device size in physical blocks..: 786432
Building bootmap in '/tmp/coreos-installer-JRNSyV'
Adding IPL section
  initial ramdisk...: /tmp/coreos-installer-JRNSyV/boot/ostree/fedora-coreos-52a6950d57dd4628986758f0492289b6f56082f9c95318e473506f9ec7f4c051/initramfs-6.18.10-200.fc43.s390x.img
  kernel image......: /tmp/coreos-installer-JRNSyV/boot/ostree/fedora-coreos-52a6950d57dd4628986758f0492289b6f56082f9c95318e473506f9ec7f4c051/vmlinuz-6.18.10-200.fc43.s390x
  kernel parmline...: 'rw mitigations=auto,nosmt ostree=/ostree/boot.1/fedora-coreos/52a6950d57dd4628986758f0492289b6f56082f9c95318e473506f9ec7f4c051/0 ignition.platform.id=metal console=ttyS123,9600n8 ignition.firstboot'
  component address:
    heap area.......: 0x00002000-0x00005fff
    stack area......: 0x0000f000-0x0000ffff
    internal loader.: 0x0000a000-0x0000dfff
    parameters......: 0x00009000-0x000091ff
    kernel image....: 0x00010000-0x00ea0fff
    parmline........: 0x00ea1000-0x00ea11ff
    initial ramdisk.: 0x00ec0000-0x04d33fff
    environment blk.: 0x00eb1000-0x00eb13ff
zIPL environment block content:
Preparing boot device for LD-IPL: vdb (0000).
Installing on base disk: vdb.
Detected SCSI PCBIOS disk layout.
Writing SCSI master boot record.
Syncing disks...
Done.
Updating re-IPL device
Re-IPL type: ccw
Device:      0.0.0001
Loadparm:    ""
clear:       0

Note: detected other devices with a filesystem labeled `boot`:
  - /dev/vdc3
The installed OS may not work correctly if there are multiple boot filesystems.
Before rebooting, investigate whether these filesystems are needed and consider
wiping them with `wipefs -a`.

Install complete.

```

dustymabe:
Took me some time to understand all of the pieces here.

Even though we don't ship s390x information in https://github.com/coreos/fedora-coreos-config/blob/testing-devel/platforms.yaml maybe it would be nice if we didn't stop that from working in the future if we were to add entries there? 

Looking at https://github.com/coreos/coreos-installer/issues/1171 would this patch also solve the problem?

```patch
diff --git a/src/install.rs b/src/install.rs
index 12704eb..7f18fc9 100644
--- a/src/install.rs
+++ b/src/install.rs
@@ -634,7 +634,8 @@ fn write_console(mountpoint: &Path, platform: Option<&str>, consoles: &[Console]
             .context("setting platform kernel arguments")
     })?;
 
-    // set grub commands
+    // set grub commands (skip on s390x because no GRUB there)
+    #[cfg(not(target_arch = "s390x"))]
     if grub_commands != metal_spec.grub_commands {
         // prefer the new grub2/console.cfg, but fallback to grub2/grub.cfg
         let mut name = "grub2/console.cfg";
```

In this case we still build up the grub_commands list in the earlier code block, but we just don't do anything with it.

src/install.rs
```diff
@@ -466,6 +471,56 @@ fn write_disk(
...
+            let console_kargs: Vec<_> = consoles.iter().map(|c| c.karg()).collect();
+            kargs_editor.append_if_missing(&console_kargs);
+        }
+    }
+    if kargs_editor != KargsEditor::default() {
+        eprintln!("Modifying kernel arguments");
+
+        // On s390x there is no GRUB config, so skip this check to avoid a user warning.
+        #[cfg(not(target_arch = "s390x"))]
+        Console::maybe_warn_on_kargs(append_kargs, "--append-karg", "--console");
```
dustymabe, line 513: If I understand this warning correctly it has nothing to do with GRUB or not GRUB.

The warning is a helper encouraging users to use the "nicer" `--console` argument/option versions embedding a `console=foobar` entry in a karg argument/option. It's purely a `coreos-installer` CLI interface thing. 

https://github.com/coreos/coreos-installer/commit/c8a24ea80d75b5447de386a85e279dd3c9ab70ad

nikita-dubrovskii: Here’s the thing: the preferred way is to use `--console`, which is later translated into `grub` configs. However, on `s390x` we now convert it into kargs, so even though we advised users not to set it via `--append-karg`, that was the only workaround before. I guess this warning is not that necessary.

dustymabe: > Here’s the thing: the preferred way is to use `--console`, which is later translated into `grub` configs. 

You do mention something I hadn't thought about before, which is that the GRUB "sugar" here is only done if `--console` is used and not `--append-karg`.

> However, on `s390x` we now convert it into kargs, so even though we advised users not to set it via `--append-karg`, that was the only workaround before. 

Ok. So the workaround for #1171 that we've been telling people is to use `--append-karg` instead? Got it.

> I guess this warning is not that necessary.

Right. I really didn't understand earlier that `--console foo` was doing more things than if you just `--apend-kargs console=foo`. I thought it was just a preferred way of passing the console information for a machine. I think the warning could still be OK on s390x, but if we wanted to neuter it just on s390x that's probably OK too with a nice comment.

nikita-dubrovskii: > Ok. So the workaround for https://github.com/coreos/coreos-installer/issues/1171 that we've been telling people is to use --append-karg instead? Got it.

No, sorry for the confusion, people had to use `--append-karg console=ttyS123456`,  which triggered a warning advising to use `--console`, but that in turn caused an error.

nikita-dubrovskii:
#1736  was merged, closing this

PR #1735, opened 2026-04-08
opencode: add signing-key-rotation and release-prep skills

prestist:
Add two OpenCode automation skills:

- signing-key-rotation: automates adding/dropping Fedora signing keys in src/signing-keys.asc and docs/release-notes.md
- release-prep: automates pre-release PR preparation including dependency updates and release notes drafting


yasminvalim (approved):
LGTM

PR #1736, opened 2026-04-10
install: refactor console handling and disable GRUB on s390x

nikita-dubrovskii:
On s390x, installations with `--console` would fail because the installer attempted to modify GRUB config files that don’t exist (`s390x` uses `zipl`).  

This PR refactors the code to better separate `--platform` and `--console` logic, and disables all GRUB-related code paths for `s390x`.

Issue: https://github.com/coreos/coreos-installer/issues/1171

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -414,33 +414,17 @@ fn write_disk(
             write_ignition(mount.mountpoint(), &config.ignition_hash, ignition)
                 .context("writing Ignition configuration")?;
         }
-        if let Some(platform) = config.platform.as_ref() {
-            write_platform(mount.mountpoint(), platform).context("writing platform ID")?;
-        }
-        if config.platform.is_some() || !config.console.is_empty() {
-            write_console(
-                mount.mountpoint(),
-                config.platform.as_deref(),
-                &config.console,
-            )
-            .context("configuring console")?;
-        }
-        if let Some(firstboot_args) = config.firstboot_args.as_ref() {
-            write_firstboot_kargs(mount.mountpoint(), firstboot_args)
-                .context("writing firstboot kargs")?;
-        }
-        if !config.append_karg.is_empty() || !config.delete_karg.is_empty() {
-            eprintln!("Modifying kernel arguments");
 
-            Console::maybe_warn_on_kargs(&config.append_karg, "--append-karg", "--console");
-            visit_bls_entry_options(mount.mountpoint(), |orig_options: &str| {
-                KargsEditor::new()
-                    .append(config.append_karg.as_slice())
-                    .delete(config.delete_karg.as_slice())
-                    .maybe_apply_to(orig_options)
-            })
-            .context("deleting and appending kargs")?;
-        }
+        configure_kernel_arguments(
+            mount.mountpoint(),
+            config.firstboot_args.as_deref(),
+            config.platform.as_deref(),
+            &config.console,
+            &config.append_karg,
+            &config.delete_karg,
+        )
+        .context("configuring kernel arguments")?;
+
         if let Some(network_config) = network_config.as_ref() {
             copy_network_config(mount.mountpoint(), network_config)?;
         }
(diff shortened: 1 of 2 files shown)


src/install.rs
```diff
@@ -414,33 +416,25 @@ fn write_disk(
...
+        )
+        .context("configuring kernel arguments")?;
+
+        #[cfg(not(target_arch = "s390x"))]
+        configure_grub(
+            mount.mountpoint(),
+            config.platform.as_deref(),
+            &config.console,
+        )
+        .context("configuring GRUB")?;
```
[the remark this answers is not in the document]
nikita-dubrovskii: > Redundant Reads

Indeed, adding `thread_local + RefCell` to cache the file would make the code less readable, so I intentionally skipped that change. In the worst case, `platforms.json` may be read three times, which should not introduce significant overhead. 

> Multiple Writes

Nothing new here - this was already the case before. I considered using a single instance of `KargsEditor` and passing it to functions as `&mut`, but in my opinion that makes the code uglier.

nikita-dubrovskii:
Conflicts with: https://github.com/coreos/coreos-installer/pull/1732.

This PR adds further refactoring to clean up the code and better separate the logic. Also, while `platforms.json` is currently never present on s390x, it may appear in the future - this change prepares for that by only skipping GRUB-related steps.

src/install.rs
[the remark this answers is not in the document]
nikita-dubrovskii:
```
diff --git a/src/install.rs b/src/install.rs
index 57f9393..2f77753 100644
--- a/src/install.rs
+++ b/src/install.rs
@@ -17,6 +17,7 @@ use nix::mount;
 #[cfg(not(target_arch = "s390x"))]
 use regex::{Captures, Regex};
 use serde::Deserialize;
+use std::cell::RefCell;
 use std::collections::HashMap;
 use std::fs::{self, File, OpenOptions, Permissions};
 use std::io::{self, BufReader, Seek, SeekFrom, Write};
@@ -612,15 +613,32 @@ struct PlatformSpec {
     kernel_arguments: Vec<String>,
 }
 
+thread_local! {
+    static PLATFORM_SPECS_CACHE: RefCell<Option<HashMap<String, PlatformSpec>>> = const { RefCell::new(None) };
+}
+
 /// Read the platform specifications from the platforms.json file.
+/// Uses a singleton cache to avoid re-reading the file multiple times.
 fn read_platform_specs(mountpoint: &Path) -> Result<HashMap<String, PlatformSpec>> {
+/// Uses a singleton cache to avoid re-reading the file multiple times.
 fn read_platform_specs(mountpoint: &Path) -> Result<HashMap<String, PlatformSpec>> {
-    match fs::read_to_string(mountpoint.join("coreos/platforms.json")) {
-        Ok(json) => serde_json::from_str::<HashMap<String, PlatformSpec>>(&json)
-            .context("parsing platform table"),
-        // no table for this image?
-        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(Default::default()),
-        Err(e) => Err(e).context("reading platform table"),
-    }
+    PLATFORM_SPECS_CACHE.with(|cache| {
+        let mut cache_ref = cache.borrow_mut();
+
+        if let Some(specs) = cache_ref.as_ref() {
+            return Ok(specs.clone());
+        }
+
+        // Load platform specs from file
+        let specs = match fs::read_to_string(mountpoint.join("coreos/platforms.json")) {
+            Ok(json) => serde_json::from_str::<HashMap<String, PlatformSpec>>(&json)
+                .context("parsing platform table")?,
+            // no table for this image?
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => Default::default(),
+            Err(e) => return Err(e).context("reading platform table"),
+        };
+
+        *cache_ref = Some(specs.clone());
+        Ok(specs)
+    })
 }
 
 /// Configure platform-specific kernel arguments
```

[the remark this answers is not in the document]
nikita-dubrovskii: Here is a patch, i don't mind to bring it back. cc @dustymabe

src/install.rs
dustymabe, line 1: I'm not the biggest fan of including ~10 `#[cfg(not(target_arch = "s390x"))]` statements in here mainly for code readability. Can we minimize it to just 2?

- line 430
- line 932


I realize this probably makes the binary size smaller, but how much does it really save? It would have to be a lot for it to be worth it I think.

src/install.rs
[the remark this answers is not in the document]
dustymabe: not worth it IMO

src/install.rs
nikita-dubrovskii: In such case there will be a lot of warnings like: 
```
warning: function `build_grub_console_commands` is never used
```

dustymabe: so not a hard failure?

@travier WDYT?

travier (approved):
I did not test this code but the style and implementation LGTM.

src/install.rs
travier: I think we should include them as it makes it clear that those functions are not used on a specific arch. It matches what I've seen in other Rust projects AFAIK.

dustymabe (approved):

src/install.rs
travier: If we want to minimize it further then we could split the code into another file and set the conditionnal there.

dustymabe: ok. man this really makes the code ugly though.

PR #1753, opened 2026-06-23
s390x: use pvimg tool instead of obsolete genprotimg

nikita-dubrovskii:
el9,el10 and rhel-9.6 now ship s390utils-base >= 2.36.0, so switch from the obsolete `genprotimg` tool to `pvimg`.

We can also use the `--overwrite` option instead of manually removing the existing `sdboot` image.

Btw, `genprotimg` is now just a symlink to `pvimg`.
```
$ file /usr/bin/genprotimg
/usr/bin/genprotimg: symbolic link to pvimg
```

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -19,6 +19,7 @@ Internal changes:
 
 - install: Fix '--console' flag handling on s390x
 - install: Disable GRUB configuration on s390x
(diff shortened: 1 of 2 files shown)


Rolv-Apneseth (approved):
LGTM

jbtrystram (approved):

PR #1754, opened 2026-06-25
s390x: Support multiple hostkeys in rdcore zipl

nikita-dubrovskii:
Allow specifying --hostkey/-k multiple times for flexible Secure Execution configurations.

### DIFF
--- a/src/bin/rdcore/cmdline.rs
+++ b/src/bin/rdcore/cmdline.rs
@@ -136,8 +136,8 @@ pub struct ZiplConfig {
     pub secex_mode: s390x::ZiplSecexMode,
 
     /// Path to hostkey
-    #[arg(long, value_name = "HOSTKEY")]
-    pub hostkey: Option<String>,
+    #[arg(short = 'k', long = "hostkey", value_name = "HOSTKEY")]
+    pub hostkeys: Option<Vec<String>>,
 
     /// Append kernel argument
     #[arg(long, value_name = "KARG")]
--- a/src/s390x/zipl.rs
+++ b/src/s390x/zipl.rs
@@ -226,7 +226,7 @@ fn get_info_from_bls(boot: &Path) -> Result<(String, String, String)> {
 fn generate_sdboot(
     mountpoint: &Path,
     boot: &Path,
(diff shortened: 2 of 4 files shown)


src/s390x/zipl.rs
```diff
@@ -268,15 +268,14 @@ fn generate_sdboot(
...
+    let hostkeys = match hostkeys {
+        Some(keys) => keys.iter().map(PathBuf::from).collect(),
+        None => find_files("/etc/se-hostkeys", |e: &DirEntry| {
             Ok(e.file_name()
                 .to_str()
                 .map(|p| p.starts_with("ibm-z-hostkey-"))
                 .unwrap_or_default())
-        })?
+        })?,
     };
```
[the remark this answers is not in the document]
nikita-dubrovskii: Nonrelevant

src/bin/rdcore/cmdline.rs
```diff
@@ -136,8 +136,8 @@ pub struct ZiplConfig {
     pub secex_mode: s390x::ZiplSecexMode,
 
     /// Path to hostkey
-    #[arg(long, value_name = "HOSTKEY")]
-    pub hostkey: Option<String>,
+    #[arg(short = 'k', long = "hostkey", value_name = "HOSTKEY")]
```
dustymabe, line 139: I like long options. I won't require us to not add `-k` support here, but I'd recommend against it. Usually for short options (unless you use a CLI tool often by hand) you just end up looking up what it means anyway. I'd rather just keep the long option.

dustymabe (approved):
LGTM - one strong-ish recommendation

src/bin/rdcore/cmdline.rs
nikita-dubrovskii: Yep, long options are better. I was testing the case where many keys are required, and after typing `--hostkey` for the third time I ended up using the short version.

Also, `pvimg-create` also uses `-k` for a similar option: 
```
-k, --host-key-document <FILE>
Use FILE as a host-key document. Can be specified multiple times and must be specified at least once.
```

PR #1762, opened 2026-08-24
blockdev: restore source images first/last usable LBAs

### LINKED ISSUE #1761: re-provisioning using `--save-partindex` may lead to ignition failures
Re-provisioning will cause ignition to fail when the `--save-partindex` flag is used and the partition has `size_mib: 0`. The cause of this issue is that the the installer will generate GPT tables with different fist and last usable LBAs depending on whether or not `--save-partindex` is used. When it is used, the first usable LBA will be 34, when it is not used, it will be padded to 2048 (the last usable LBA will also have this difference, but exact number depends on disk size). Because of this, if you initially use the installer without the flag to setup the partitions, and you later re-provision and keep a partition, `size_mib: 0` will have different meanings, and the re-provision will fail to boot.

## Expected Behavior ##

I can use the installer to setup my system, and later re-provision using the `save-partindex` flag and have a bootable system.

## Actual Behavior ##

After re-provisioning, first boot will drop to the emergency shell with the following error on `ignition-disk.service`:

    partition 5 didn't match: size did not match

## Reproduction Steps ##

1.  Create a butane config with a partition that fills the disk, the ignition this produces when booting the images.
    ```
    variant: fcos
    version: 1.5.0
    storage:
      disks:
        - device: /dev/disk/by-id/virtio-targetdisk
          partitions:
            - number: 4
              label: root
              size_mib: 8192
              resize: true
            - number: 5
              label: var
              size_mib: 0
      filesystems:
        - device: /dev/disk/by-partlabel/var
          format: ext4
          label: var
          path: /var/mnt/var-data
          with_mount_unit: true
    ```
2.  Use the installer normally
    ```
    acervera tests → truncate -s 25G target.raw
    
    acervera tests → sudo losetup -Pf --show target.raw
    /dev/loop0
    
    acervera tests → sudo coreos-installer install --insecure --platform qemu --ignition conf.ign --image-file fc45std.raw /dev/loop0
    Copying image from fc45std.raw
    Reading signature from fc45std.raw.sig
    Couldn't read signature file: No such file or directory (os error 2)
    Signature not found; skipping verification as requested
    > Read disk 3.0 GiB/3.0 GiB (100%)    
    Writing Ignition config
    Setting platform to qemu
    Install complete.
    
    acervera tests → sudo losetup -d  /dev/loop0
    ```

    Here booting will succeed, and after first boot the GPT table looks like this:

    ```
    acervera tests → sgdisk -p target.raw 
    Disk target.raw: 52428800 sectors, 25.0 GiB
    Sector size (logical): 512 bytes
    Disk identifier (GUID): D6E5A7AA-C1D3-4752-B2F5-F8687718B3EC
    Partition table holds up to 128 entries
    Main partition table begins at sector 2 and ends at sector 33
    First usable sector is 2048, last usable sector is 52426752
    Partitions will be aligned on 2048-sector boundaries
    Total free space is 0 sectors (0 bytes)
    
    Number  Start (sector)    End (sector)  Size       Code  Name
       1            2048            4095   1024.0 KiB  EF02  BIOS-BOOT
       2            4096          264191   127.0 MiB   EF00  EFI-SYSTEM
       3          264192         1050623   384.0 MiB   EA00  boot
       4         1050624        17827839   8.0 GiB     8304  root
       5        17827840        52426752   16.5 GiB    8300  var
    ```

3.  Run the installer again, this time lets try to preserve the 5th partition
    ```
    acervera tests → sudo losetup -Pf --show target.raw
    /dev/loop0
    
    acervera tests → sudo coreos-installer install --insecure --platform qemu --ignition conf.ign --image-file fc45std.raw --save-partindex 5 /dev/loop0
    Copying image from fc45std.raw
    Reading signature from fc45std.raw.sig
    Couldn't read signature file: No such file or directory (os error 2)
    Signature not found; skipping verification as requested
    Saving partition 5 ("var") to new partition 5
    
    Writing Ignition config
    Setting platform to qemu
    Install complete.
    
    acervera tests → sudo losetup -d  /dev/loop0
    ```
    Booting the re-provisioned disk, we will see the following error:
    ```
    Displaying logs from failed units: sysroot.mount
    ignition-disks.service
    Aug 24 09:11:06 systemd[1]: sysroot.mount: Failed to load environment files: No such file or directory
    Aug 24 09:11:06 systemd[1]: sysroot.mount: Failed to spawn 'mount' task: No such file or directory
    Aug 24 09:11:06 systemd[1]: sysroot.mount: Failed with result 'resources'.
    Aug 24 09:11:06 systemd[1]: Failed to mount sysroot.mount - /sysroot.
    Aug 24 09:11:05 ignition[687]: }
    Aug 24 09:11:05 ignition[687]: disks: createPartitions: op(2): op(3): [finished] reading partition table of "/run/ignition/dev_aliases/dev/disk/by-id/virtio-targetdisk"
    Aug 24 09:11:05 systemd[1]: ignition-disks.service: Main process exited, code=exited, status=1/FAILURE
    Aug 24 09:11:05 ignition[687]: disks: createPartitions: op(2): running sgdisk with options: [--pretend --delete=4 --delete=5 --new=4:1050624:+16777216 --new=5:17827840:+0 --info=5 /run/ignition/dev_aliases/dev/disk/by-id/virtio-targetdisk]
    Aug 24 09:11:05 systemd[1]: ignition-disks.service: Failed with result 'exit-code'.
    Aug 24 09:11:05 ignition[687]: disks: createPartitions: op(2): resizing partition 4
    Aug 24 09:11:05 systemd[1]: Failed to start ignition-disks.service - Ignition (disks).
    Aug 24 09:11:05 ignition[687]: disks: createPartitions: op(2): [failed]   partitioning "/run/ignition/dev_aliases/dev/disk/by-id/virtio-targetdisk": partition 5 didn't match: size did not match (specified 34600927, got 34598913)
    Aug 24 09:11:05 ignition[687]: Ignition failed: create partitions failed: partition 5 didn't match: size did not match (specified 34600927, got 34598913)
    Aug 24 09:11:05 systemd[1]: ignition-disks.service: Triggering OnFailure= dependencies.
    ```
    Lookig at the GPT table of this image, we can see that the last usable LBA and the first usable LBA are different than before, leading to a different last usable sector.
    ```
    09:13:22 acervera@acervera-thinkpadx1carbongen11 tests → sgdisk -p target.raw 
    Disk target.raw: 52428800 sectors, 25.0 GiB
    Sector size (logical): 512 bytes
    Disk identifier (GUID): F70EAFE1-3842-4177-B03A-745BAA5A45C1
    Partition table holds up to 128 entries
    Main partition table begins at sector 2 and ends at sector 33
    First usable sector is 34, last usable sector is 52428766
    Partitions will be aligned on 2048-sector boundaries
    Total free space is 11569084 sectors (5.5 GiB)
    
    Number  Start (sector)    End (sector)  Size       Code  Name
       1            2048            4095   1024.0 KiB  EF02  BIOS-BOOT
       2            4096          264191   127.0 MiB   EF00  EFI-SYSTEM
       3          264192         1050623   384.0 MiB   EA00  boot
       4         1050624         6262783   2.5 GiB     8304  root
       5        17827840        52426752   16.5 GiB    8300  var
    ```
## Other Information ##
One workaround for this issue is to simply have `resize: true` in the config, however, `resize: false` is the default, and I think it makes sense for someone to want resizing to be disabled to avoid any potential issues / data loss.

I have opened a PR trying to address this issue by making sure that the first and last usable LBAs always match the base image, but I am not too familiar with how GPT tables, so a review would be very welcome.

angelcerveraroldan:
`update_from` is called when the installer is executed with a flag to preserve a partition, this re-computes the first usable lba as well as the last usable lba using GPTs minimums, instead, we want to persevere the source images alignment. Otherwise, we will have differently formatted GPT tables depending on whether or not we preserved a partition, and this will lead to issues with ignition.

closes: #1761


src/blockdev.rs
```diff
@@ -794,6 +809,11 @@ impl SavedPartitions {
     }
 }
 
+fn adjust_gpt_alignment(gpt: &mut GPT, source_first_usable_lba: u64) {
+    gpt.header.first_usable_lba = source_first_usable_lba;
+    gpt.header.last_usable_lba = gpt.header.backup_lba + 1 - source_first_usable_lba;
```
jbtrystram, line 814: IIUC this assumes the last LBA would have move as much as the first one moved ? Can't we just do the same as you did for the first LBA and simply restore the value ? 
I am having trouble following the reasonning so a comment would help :)

jbtrystram (commented):
I am not well-versed into this so i won't approve but it was an opportunity to open wikipedia and learn a bit, thanks. 
Left a question

src/blockdev.rs
angelcerveraroldan: Good question! I should certainly add a comment to clarify.

The `first_usable_lba` in the source image is constant, and it does not depend on the disk size (since its always just at the start), so its ok to just copy it over. 

However, the value of `last_usable_lba` depends on the size of the disk, so we cannot just restore the one that was in the base image.

Before we adjust the alignment, `update_from` will have correctly moved the GPT headers `backup_lba` to the end of the disk, but it also sets `last_usable_lba` to the GPT minimum.

PR #1765, opened 2026-09-02
install: inject network config into initramfs instead of copying to disk

### LINKED ISSUE #1758: try to have `--copy-network` support ISCSI based installs via initramfs file injection
Traditionally our support for ISCSI in CoreOS/OpenShift has necessitated using DHCP for networking and disallowed static networking configuration.

It is possible, however, to configure SAN/ISCSI information in the firmware of some hardware platforms (CISCO UCS being an example) and thus not requiring that the source of networking information come from DHCP/iPXE. In this scenario DHCP isn't in the picture and the desire to configure networking statically is a legitimate one.

Based on the earlier DHCP assumption `coreos-installer install --copy-network` hasn't worked for ISCSI based deployments because the [implementation](https://github.com/coreos/coreos-installer/blob/431a12076c5a2ef3eb1e3ab0eec4d5a1145c8f99/src/install.rs#L414-L440) that copies files into the /boot/ partition of the installed disk, which (you guessed it) isn't available in the initramfs until after networking has been brought up and the SAN disk becomes available.

Now, the service that actually picks up the copy-network files and puts them to use has supported for some time [copy-network files "added" to the initramfs](https://github.com/coreos/coreos-installer/issues/545) (see https://github.com/coreos/fedora-coreos-config/pull/1371), but that was only implemented for live boots (boots from the ISO or PXE) and not "diskful" boots from the hard drive.

What we should investigate/consider here is to:

1. Support initramfs injection for copy-network for diskful boots
2. remove using the /boot filesystem for copy-network functionality at all (simplifies code paths)

Using intramfs file injection would mean the files are available prior to the disks (and network) needing to be accessible, solving the chicken/egg problem we've had in the past.

This should also remove the need for something like https://github.com/coreos/coreos-installer/issues/1389

ravanelli:
- Rewrite copy_network_config() to build an xz-compressed newc CPIO
    archive of the network configuration files and append it to the existing
    initramfs, rather than copying files to a directory under /boot. This
    makes network config available in RAM before target disks are
    initialized, removing the dependency on a mounted /boot partition.

- Fix: https://github.com/coreos/coreos-installer/issues/1758

Tests done using the great doc Dusty did: https://dustymabe.com/2024/05/10/configuring-and-using-iscsi/

```
coreos-installer install /dev/sda \
  --copy-network \
  --image-file fedora-coreos.raw.xz \
  --insecure \
  --console ttyS0,115200n8 \
  -i config.ign \
  --append-karg "rd.neednet=1" \
  --append-karg "rd.iscsi.firmware=1" 
  ```
Fake  file created and copied to the installed system with the existing enp1s0.nmconnection 
```

Copying image from fedora-coreos-44.20260901.20.dev0-metal.x86_64.raw.xz
Reading signature from fedora-coreos-44.20260901.20.dev0-metal.x86_64.raw.xz.sig
Couldn't read signature file: No such file or directory (os error 2)
Signature not found; skipping verification as requested
[528266.477347]  sda: sda1 sda2 sda3 sda4
> Read disk 1004.9 MiB/1004.9 MiB (100%)
[528386.728102] GPT:Primary header thinks Alt. header is not at the end of the disk.
[528386.733673] GPT:6166527 != 62914559
[528386.736144] GPT:Alternate GPT header not at the end of the disk.
[528386.740832] GPT:6166527 != 62914559
[528386.744110] GPT: Use GNU Parted to correct GPT errors.
[528386.748988]  sda: sda1 sda2 sda3 sda4
[528387.828482] EXT4-fs (sda3): mounted filesystem 96d15588-3596-4b3c-adca-a2ff7279ea63 r/w with ordered data mode. Quota mode: none.
Writing Ignition config
Configuring console settings
Modifying kernel arguments
Copying networking configuration from /etc/NetworkManager/system-connections/
Copying /etc/NetworkManager/system-connections/test-copy.nmconnection to initramfs
Copying /etc/NetworkManager/system-connections/enp1s0.nmconnection to initramfs
Appended network configuration to initramfs /boot/ostree/fedora-coreos-001490ce387856204060b4e7e9cd9f065739bac1668108dcf120d17bbfe88d52/initramfs-7.1.12-200.fc44.x86_64.img



[core@localhost ~]$ ls /etc/NetworkManager/system-connections
/etc/NetworkManager/system-connections/test-copy.nmconnection
/etc/NetworkManager/system-connections/enp1s0.nmconnection
```
For the feature check:

```
> Read disk 969.2 MiB/969.2 MiB (100%)
gpg: Signature made Thu Sep  3 00:02:44 2026 EDT
gpg:                using RSA key 36F612DCF27F7D1A48A835E4DBFCF71C6D9F90A6
gpg: Good signature from "Fedora (44) <fedora-44-primary@fedoraproject.org>" [ultimate]
[528618.434675] GPT:Primary header thinks Alt. header is not at the end of the disk.
[528618.444402] GPT:6045695 != 62914559
[528618.449984] GPT:Alternate GPT header not at the end of the disk.
[528618.456898] GPT:6045695 != 62914559
[528618.462775] GPT: Use GNU Parted to correct GPT errors.
[528618.469539]  sda: sda1 sda2 sda3 sda4
[528619.040243] EXT4-fs (sda3): mounted filesystem 96d15588-3596-4b3c-adca-a2ff7279ea63 r/w with ordered data mode. Quota mode: none.
Copying networking configuration from /etc/NetworkManager/system-connections/
[528619.074244] EXT4-fs (sda3): unmounting filesystem 96d15588-3596-4b3c-adca-a2ff7279ea63.

Error: This OS image does not support --copy-network via initramfs. Please use a newer OS image.

Resetting partition table
[528619.162749]  sda:
Error: install failed
```

### CODE REVIEW

**Comment by ravanelli** [comment]
Do not merge yet, it still need the feature check part

File: src/install.rs [hunk 1]
```diff
@@ -781,34 +781,59 @@ fn apply_grub_console_commands(grub_cfg: &str, commands: &[String]) -> Result<St
...
+                entry.file_name().to_string_lossy()
+            );
+            initrd.add(&archive_path, contents);
         }
     }
 
+    if initrd.is_empty() {
+        eprintln!("No networking configuration files found in {net_config_src}");
+        return Ok(());
+    }
```
dustymabe, line 815: seems heavyweight to do this check later (here) instead of up front just checking if the `net_config_src` dir is empty an skipping this part?

src/install.rs
```diff
@@ -781,34 +781,59 @@ fn apply_grub_console_commands(grub_cfg: &str, commands: &[String]) -> Result<St
...
+
+    if !initrd_file.exists() {
+        bail!("initramfs file not found at {:?}", initrd_file);
+    }
+
+    let mut file = OpenOptions::new()
+        .write(true)
+        .append(true)
+        .open(&initrd_file)
+        .with_context(|| format!("opening initramfs {} for appending", initrd_file.display()))?;
```
dustymabe, line 830: maybe rather than the full path here (includes tmpdir components) we just log the basename of `initrd_file.display()`

src/install.rs
```diff
@@ -781,34 +781,59 @@ fn apply_grub_console_commands(grub_cfg: &str, commands: &[String]) -> Result<St
         .into_owned())
 }
 
-/// Copy networking config if asked to do so
+/// Embed networking config into the in-memory initramfs at /etc/coreos-firstboot-network
+/// so it is available in RAM before target disks are initialized.
 fn copy_network_config(mountpoint: &Path, net_config_src: &str) -> Result<()> {
     eprintln!("Copying networking configuration from {net_config_src}");
```
dustymabe, line 787: I'm wondering how much different the implementation here is with what is in the underlying functions of: 

https://github.com/coreos/coreos-installer/blob/1f8ea303e5bb431caa15001c07595b215d198ce7/src/live/mod.rs#L210-L219

ravanelli: aa indeed!

dustymabe:
@ravanelli I made OSfeatures/compatibility check recommendation prior to knowing that we would essentially also still be supporting the old path (`coreos-rescue-misplaced-network.service` from https://github.com/coreos/fedora-coreos-config/pull/4336 )

Do you think it's still worth doing that compatibility work?

src/install.rs
```diff
@@ -781,34 +782,69 @@ fn apply_grub_console_commands(grub_cfg: &str, commands: &[String]) -> Result<St
...
+    }
+
+    let mut file = OpenOptions::new()
+        .write(true)
+        .append(true)
+        .open(&initrd_file)
+        .with_context(|| format!("opening initramfs {initrd_rel} for appending"))?;
+
+    file.write_all(&initrd_bytes)
+        .with_context(|| format!("appending network config to initramfs {initrd_rel}"))?;
```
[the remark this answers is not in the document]
dustymabe: Could be good to do this.

ravanelli:
> @ravanelli I made OSfeatures/compatibility check recommendation prior to knowing that we would essentially also still be supporting the old path (`coreos-rescue-misplaced-network.service` from [coreos/fedora-coreos-config#4336](https://github.com/coreos/fedora-coreos-config/pull/4336) )
> 
> Do you think it's still worth doing that compatibility work?

I think it depends on the path we take for the backport and how far back we want to go.

This feature check will warn about using a newer coreos-installer with an older image, while the rescue in f-c-c covers the opposite scenario: an older coreos-installer with newer images. Adding the check would help avoid the issue in the first place, but either way, we’re covered if it happens.

If we’re only backporting this to one or two versions, maybe it’s worth dropping the check?

dustymabe:
> I think it depends on the path we take for the backport and how far back we want to go.
> 

ideally we'd only ever take this back maybe to 4.22 (rhel-9.8) and not older.

> This feature check will warn about using a newer coreos-installer with an older image,

I guess this could happen

> while the rescue in f-c-c covers the opposite scenario: an older coreos-installer with newer images.

This more likely to happen.

> Adding the check would help avoid the issue in the first place, but either way, we’re covered if it happens.
> 
> If we’re only backporting this to one or two versions, maybe it’s worth dropping the check?

yeah. The OS Features stuff is a lot of machinery. I think we needed it when we weren't doing `coreos-rescue-misplaced-network.service` but now that we have that and a plan to deprecate it in el11 it's probably fine to not have the OS Features pieces.

ravanelli:
Dropping the Feature check 

>> This feature check will warn about using a newer coreos-installer with an older image, while the rescue in f-c-c covers the opposite scenario: an older coreos-installer with newer images. Adding the check would help avoid the issue in the first place, ~~but either way, we’re covered if it happens.~~

Just rephrasing it: with a newer `coreos-installer` and an older image, we won't be able to recover using the rescue service, since the an old image won't have it,  but for now, we'll drop it.

ravanelli:
coderabbit review --agent

src/install.rs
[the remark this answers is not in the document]
ravanelli: done, moved to the tmp file instead

dustymabe:
One thing I think we are missing is preserving ownership and permissions from the original file. Something like this (written by `<openai/gpt-5.6-luna>`):

```diff
diff --git a/src/io/initrd.rs b/src/io/initrd.rs
index cd4dbc4..c723c7f 100644
--- a/src/io/initrd.rs
+++ b/src/io/initrd.rs
@@ -15,9 +15,11 @@
 use anyhow::{anyhow, bail, Context, Result};
 use cpio::{write_cpio, NewcBuilder, NewcReader};
 use lazy_static::lazy_static;
+use nix::unistd::{fchown, Gid, Uid};
 use std::collections::BTreeMap;
 use std::fs::{self, File};
 use std::io::{self, BufRead, Cursor, Read, Write};
+use std::os::unix::fs::MetadataExt;
 use std::path::Path;
 use tempfile::NamedTempFile;
 use xz2::stream::{Check, Stream};
@@ -187,6 +189,8 @@ impl Initrd {
         let dir = source
             .parent()
             .expect("initrd file must have a parent directory");
+        let metadata = fs::metadata(source)
+            .with_context(|| format!("statting initramfs {}", source.display()))?;
         let mut tmp = tempfile::Builder::new()
             .prefix(".coreos-installer-initrd.")
             .append(true)
@@ -202,6 +206,15 @@ impl Initrd {
         tmp.as_file_mut()
             .write_all(&initrd_bytes)
             .with_context(|| format!("appending data to initramfs: {}", source.display()))?;
+        fchown(
+            tmp.as_file(),
+            Some(Uid::from_raw(metadata.uid())),
+            Some(Gid::from_raw(metadata.gid())),
+        )
+        .with_context(|| format!("preserving ownership of {}", source.display()))?;
+        tmp.as_file()
+            .set_permissions(metadata.permissions())
+            .with_context(|| format!("preserving permissions of {}", source.display()))?;
         tmp.as_file()
             .sync_all()
             .context("syncing temp initramfs to disk")?;
```

src/install.rs
```diff
@@ -781,34 +782,67 @@ fn apply_grub_console_commands(grub_cfg: &str, commands: &[String]) -> Result<St
...
         }
     }
 
+    let mut initrd = Initrd::default();
+    initrd.embed_network_files(&keyfiles)?;
+
+    let initrd_bytes = initrd.to_bytes().context("building network config initrd")?;
+
+    let (_, initrd_rel, _) = get_bls_info(mountpoint)?;
+    let initrd_file = mountpoint.join(&initrd_rel[1..]);
```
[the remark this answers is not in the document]
dustymabe: I think we need to consider doing this..

> This assumes the BLS path always starts with /. A relative path such as initramfs.img becomes nitramfs.img, and an empty path can panic. Paths containing .. are also not constrained to the mounted filesystem.

src/io/initrd.rs
```diff
@@ -165,16 +176,82 @@ impl Initrd {
...
+        Ok(tmp)
+    }
+
+    pub fn embed_network_files(&mut self, keyfiles: &[String]) -> Result<()> {
+        for path in keyfiles {
+            let data = fs::read(path).with_context(|| format!("reading {path}"))?;
+            let mode = fs::metadata(path)
+                .with_context(|| format!("statting {path}"))?
+                .permissions()
+                .mode();
```
dustymabe: Preserving the mode of the network keyfiles wasn't a problem I was trying to address in https://github.com/coreos/coreos-installer/pull/1765#issuecomment-5806196399 (for the record you did address my concern there, just also did this extra thing too)

This may cause more problems than it's worth since we haven't been doing this in the past (i.e. the existing code that copies network prior to this PR doesn't do this, right?).

dustymabe (approved):
This looks good to me. Thank you for working on it.

Under normal situations I would love to compile and test this out myself, but I just don't have time for that right now.

Since this is an important piece of work maybe we could loop in someone like JB to help validate this with you.

Either way don't let me block this. LGTM.

src/install.rs
```diff
@@ -793,34 +781,48 @@ fn apply_grub_console_commands(grub_cfg: &str, commands: &[String]) -> Result<St
...
-        let destpath = net_config_dest.join(entry.file_name());
         if srcpath.is_file() {
-            eprintln!("Copying {} to installed system", srcpath.display());
-            fs::copy(&srcpath, destpath).context("Copying networking config")?;
+            eprintln!("Copying {} to initramfs", srcpath.display());
+            keyfiles.push(srcpath.to_string_lossy().into_owned());
         }
     }
 
+    let mut initrd = Initrd::default();
```
prestist: With this change is there any risk of unsupported target images silently losing copied networking?

src/install.rs
```diff
@@ -793,34 +781,48 @@ fn apply_grub_console_commands(grub_cfg: &str, commands: &[String]) -> Result<St
...
 
-/// Copy networking config if asked to do so
+/// Embed networking config into the in-memory initramfs at /etc/coreos-firstboot-network
+/// so it is available in RAM before target disks are initialized.
 fn copy_network_config(mountpoint: &Path, net_config_src: &str) -> Result<()> {
-    eprintln!("Copying networking configuration from {net_config_src}");
-
-    // get the path to the destination directory
-    let net_config_dest = mountpoint.join("coreos-firstboot-network");
+    if fs::read_dir(net_config_src)
```
prestist: Wait so if this is empty the it fails, in the past function would have accepted an empty source, I think that would be valid in a DHCP deployed env?

prestist (commented):
Just a quick question or two

PR #1769, opened 2026-09-16
install: accept Butane YAML configs via --ignition-file/--ignition-url

### LINKED ISSUE #1770: install: accept Butane YAML configs directly via --ignition-file/--ignition-url
## Background

With [coreos/ignition#2235](https://github.com/coreos/ignition/pull/2235) merged, Butane has been absorbed into Ignition and `ignition`'s `config.Parse()` now transparently handles both Ignition JSON and Butane YAML at boot time. This is the groundwork needed to close [coreos/fedora-coreos-tracker#2006](https://github.com/coreos/fedora-coreos-tracker/issues/2006).

Today, users must manually transpile their Butane config before passing it to coreos-installer:

```
butane < config.bu > config.ign
coreos-installer install --ignition-file config.ign /dev/sda
```

Since the installed system's Ignition can now handle Butane YAML natively, coreos-installer can accept Butane configs directly and pass them through as-is — no transpilation step needed:

```
coreos-installer install --ignition-file config.bu /dev/sda
```

## What needs to change

- `src/install.rs`: The current validation rejects anything that isn't valid JSON. It should also accept Butane YAML (identified by a top-level `variant` key).
- Help text for `--ignition-file` and `--ignition-url` should document both formats.
- Generated docs and man pages updated accordingly.

Note: the `iso customize` and `pxe customize` paths manipulate the config directly and should remain Ignition JSON only.

## References

- [coreos/ignition#2235](https://github.com/coreos/ignition/pull/2235) — Butane merged into Ignition
- [coreos/fedora-coreos-tracker#2006](https://github.com/coreos/fedora-coreos-tracker/issues/2006) — Allow passing Butane configs directly to instances

prestist:
Closes #1770

## Why

With [coreos/ignition#2235](https://github.com/coreos/ignition/pull/2235) now merged, Butane has been absorbed into Ignition and `ignition`'s `config.Parse()` transparently handles both Ignition JSON and Butane YAML (identified by a top-level `variant` key) at boot time. This means a CoreOS system can already accept a Butane YAML config embedded on disk — coreos-installer was the only thing standing in the way.

This is the coreos-installer side of [coreos/fedora-coreos-tracker#2006](https://github.com/coreos/fedora-coreos-tracker/issues/2006): allowing users to skip the manual `butane` transpilation step and pass their Butane config directly to `--ignition-file` or `--ignition-url`.

## What changed

- **`src/install.rs`**: Replace the JSON-only validation with a check that accepts either valid Ignition JSON or Butane YAML (any YAML with a top-level `variant` key). The config is passed through as-is to the installed system; installed Ignition handles transpilation at first boot. No new dependencies — `serde_yaml` is already used by the project.
- **`src/cmdline/install.rs`**: Update `--ignition-file` and `--ignition-url` help text to document both formats.
- **Generated docs/man pages**: Regenerated via `make docs` to reflect the updated help text.

## What didn't change

The `iso customize` and `pxe customize` paths are unaffected — they need to manipulate the config directly and remain Ignition JSON only.


src/install.rs
```diff
@@ -71,15 +71,27 @@ pub fn install(config: InstallConfig) -> Result<()> {
...
+        BufReader::with_capacity(BUFFER_SIZE, &mut file)
+            .read_to_end(&mut buf)
+            .context("reading config file")?;
+        let is_ignition = serde_json::from_slice::<serde_json::Value>(&buf).is_ok();
+        let is_butane = !is_ignition
+            && serde_yaml::from_slice::<serde_yaml::Value>(&buf)
+                .ok()
+                .and_then(|v| v.get("variant").cloned())
+                .is_some();
+        if !is_ignition && !is_butane {
```
yasminvalim, line 91: Could we add a quick unit test for this, or is it already covered somewhere else? What do you think?

yasminvalim:
I think it will need to be rebased since we are having the TMT error.

prestist:
@yasminvalim hmm I think its actually not upstream

yasminvalim (approved):
/LGTM

PR #1772, opened 2026-09-18
cargo: coreos-installer release 0.27.0

yasminvalim:
Release for coreos-installer v0.27.0


Cargo.lock
```diff
@@ -274,7 +274,7 @@ dependencies = [
  "nmstate",
  "openssl",
  "pipe",
- "rand 0.9.5",
+ "rand 0.8.8",
```
prestist, line 277: why did we downgrade?

